PR #12, opened 2017-06-17
Upgradle to 4.0

wolfs:
Upgradle to 4.0.

Note that the location of compiled classes changed in Gradle 4.0 from `build/classes/<sourceSet>` to `build/classes/<language>/<sourceSet>`.
I fixed a test with respect to this but don't know if anything else is broken.

I also added a missing dependency which made the build fail when running with `--parallel` in `archunit-junit`.


wolfs:
> I think the 4.0 folder structure should just be added, to be ignored as well as '/build/classes/test/'.
Question is though, since ArchUnit can in theory test any bytecode, not just Java, if it would make sense, to ignore '/build/classes/*/test' (I suppose that e.g. Groovy would be 'classes/groovy/test' and Scala 'classes/scala/test' ?)

I think that makes perfect sense. I just didn't look deeply enough at the code to make that change.

codecholeric:
I'll merge this and make the changes on master, thanks again :smiley:

codecholeric:
Done -> https://github.com/TNG/ArchUnit/commit/13a962df57e1789f87352a8e475a8918bc17f197

PR #13, opened 2017-06-18
Simplify build

wolfs:
I am going to try to simplify the build. The first step is adding the asm license directly to `archunit:shadowJar` instead of repackaging the jar again.
If at some point in the process you see a problem please advice what your intention was and I will try to adapt.
Feel free to merge this PR in its state. I will create a new one when I am ready with the next step.

### DIFF
--- a/build-utils/license/license.gradle
+++ b/build-utils/license/license.gradle
@@ -85,35 +85,13 @@ task updateLicenses {
     }
 }
 
-def configure3rdPartyLicenses = {
-    File jarPath = shrinkDependencies.archivePath
-    File jarTempPath = tempJar(jarPath)
-
-    task add3rdPartyLicenses(type: Zip, dependsOn: shrinkDependencies) {
-        onlyIf {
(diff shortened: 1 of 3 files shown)


codecholeric:
Oh yes, please do simplify this crazy build :wink: 
The only thing I care about, is that the resulting artifact looks the same (on install and deploy... Any way I don't have to apply the "fix the POM" closure to installer AND deployer separately? Always scared that I break one and don't realize it, cause I only check the POM for the other one)
Anyway, I'll look into the changes tonight :smiley:

codecholeric:
Looks good, I just did a comparison of archunit-0.5.0-SNAPSHOT.jar and archunit-junit-0.5.0-SNAPSHOT.jar and they look the same after installing to my local repo :smiley: 
The solution before would have automatically grown, if multiple modules would have repackaged ASM, while now you have to set the property, but I guess, that would have been YAGNI anyway :wink: (since the core module repackages the whole library, it would be highly dubious, if another module would do so as well...)
Gonna merge this, thanks a lot!!!

PR #24, opened 2017-09-30
failing rules also show how often a rule was violated

hankem:
Especially when working with _legacy_ code, rules may have a large number of violations.
It is often helpful to have this number explicitly displayed, for example to easily judge whether a refactoring reduces the number or not.

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/lang/ConfiguredMessageFormat.java
+++ b/archunit/src/main/java/com/tngtech/archunit/lang/ConfiguredMessageFormat.java
@@ -29,9 +29,8 @@ class ConfiguredMessageFormat {
 
     String formatFailure(HasDescription rule, Collection<String> failureMessages, Priority priority) {
         String violationTexts = Joiner.on(System.lineSeparator()).join(failureMessages);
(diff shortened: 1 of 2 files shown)


codecholeric:
Thanks for the PR :smiley: 
I'll gladly merge this, do you think, you could add a test for this improved feature though? E.g. in `ArchRuleTest`

```
@Test
public void reports_number_of_violations() {
    EvaluationResult result = all(classes()).should(addFixedNumberOfViolations(3)).evaluate(javaClassesViaReflection(Object.class, String.class));

    assertThat(result.getFailureReport().toString()).contains("(6 times)");
}
```
where
```Java
private static ArchCondition<JavaClass> addFixedNumberOfViolations(final int number) {
    return new ArchCondition<JavaClass>(String.format("be violated exactly %d times", number)) {
        @Override
        public void check(JavaClass item, ConditionEvents events) {
            for (int i = 0; i < number; i++) {
                events.add(SimpleConditionEvent.violated(item, item.getSimpleName() + " violation " + i));
            }
        }
    };
}
```
I'd hate to destroy the number of violations in half a year :wink:

hankem:
Of course – especially given that you pointed out both the appropriate location and the implementaion! ;)

PR #25, opened 2017-09-30
added shell script to run various docker-compose options:

gernotstarke:
(b) build a custom docker container for archunit
(d) local development, serve on 0.0.0.0:4000
(p) generate html source once, then exit (using production _config.yml)

In addition, I modified the Dockerfile to start with alpine linux and install required deps.


codecholeric:
Thanks for the PR!!
I know the Jekyll + Github Pages thing is pretty self-explanatory :wink: , but maybe you can help me to understand a couple of the changes:

* What exactly does GP trigger to build the page? I mean, I just figured out, if I remove the github-pages Gem, the page doesn't get deployed anymore. So why does it work with the adjusted config, even though github-pages is removed?
* Probably the same direction, but is the destination `zz-site` something magically important, or just randomly chosen?
* How is the workflow now? So far, I would start the local server, make my changes to the markdown, push them, and gh-pages gets magically rebuilt and deployed

Sorry for my ignorance, but I fully want to understand the flow, even if it would just magically work.

Other than that, maybe there would be some way to remove the duplication between `_config.yml` and `_config-dev.yml`, because as far as I see, they're pretty similar except for the URL?

codecholeric:
Just to summarize our phone call for later reference :wink: 
- GP should trigger the normal Jekyll build, there should be no additional magic. the `jekyll` Gem works for Arc42, so it should work here, too. However, if it doesn't, I'll switch back to `github-pages` and accept the warnings
- `zz-site` is just a directory you use to create your production release, because you deploy outside of GP as well, I won't really need it though, if I just deploy through GP (same as the production docker-compose for now)
- The workflow will be to start docker compose with a `_dev`-yml file. the normal `_config.yml` will be used for automatic GP deployment. Thus I'll merge the PR and for now remove the production variants, since I only deploy to GP for now.
And yes, `_config.yml` and `_config-dev,yml` are very redundant, but to script this might not make things easier, either.
Anyway, thanks a lot for all the support :smiley: I'll try to get this all working now.

Issue #30, opened 2017-10-09
archunit-example should have (completely) self-contained build


gernotstarke:
When running 'gradle test' on the archunit-example directory, I get:

53 tests, 0 failures, 53 ignored.

As an ArchUnit-user, I don't want to test the whole library, but only a (small) example...

(this issue belongs to the more general #22 )

codecholeric:
But those 53 tests are tests only for `archunit-example`, these are 53 tests, testing the main code of `archunit-example` and @Ignored, since they all fail to demonstrate, how errors would look like...
They are also reused within `archunit-integration-test`, where the expected errors of those "usecases" are asserted.
You're certainly right, that there should be a README.md explaining this...

codecholeric:
By self-contained you mean, it's a completely separate project, where you add the Maven coordinates of `archunit` as a dependency?
Because in my view that would have some drawbacks, too.
* At the moment, all example rules are at the same time used for integration tests, since integration tests should run against the current code base, I'd have to duplicate the rules then, which I don't like too much
* It would be easier to forget to adjust `archunit-example`, if it would refer to a possibly different version (it would only be up to date, if the `maven install` step was performed, or the current version is published as SNAPSHOT
* Of course I could make `archunit-example:compile` running after `archunit:mavenInstall`, thus enforcing `mavenInstall`, when `archunit-integration-test:test` is run, but this doesn't feel like good practice as well

On the other hand, the only advantage would be to not have to compile the base `archunit` as a user, who just wants to run `archunit-example` (which on my current machine takes 3s altogether).

gernotstarke:
as a potential new user, I don't care about these drawbacks - I need examples to get started.

And: The example should be KISS (meaning: shortest possible setup time, just a single gradle build task ("run")).

codecholeric:
Okay, but at the moment, the process is
```
git clone $ARCHUNIT_URL checkout
cd checkout/archunit-example
../gradlew build -P example
```
I wouldn't exactly call that a complicated setup :wink: (I would imagine most people would import the project into their IDE anyway, and run selected tests?)

It would also be possible, to only run reasonable tasks, if `-P example` is set, so that `./gradlew` can be run from the root without the whole project being built and tested.

I'm just trying to figure out, if there is no solution, that is convenient for an user of `archunit-example` and a developer / CI running the full build.

gernotstarke:
Peter, you're right, it's not really complicated. But as a user, I don't want the source + tests, just the example...

[...]

PR #36, opened 2017-10-19
Interfaces syntax

### LINKED ISSUE #28: Should-Syntax should contain 'interface' handling
While there is a predicate syntax
```
classes().that().implement(foo)
```
the only way to filter for interfaces is using
```
classes().that(are(JavaClass.Predicates.INTERFACES))
```
The fluent API should allow filtering for interfaces
```
classes().that().areInterfaces()
classes().that().areNoInterfaces()
```
as well as asserting interfaces
```
classes().should().beInterfaces()
classes().should().notBeInterfaces()
```

hajotka:
This adds are(Not)Interfaces and (not)beInterfaces. 

The should().onlyBeAccessed().byClassesThat().are*Interfaces() doesn't make sense in Java 7, only with Java 8 with default methods. I added the relevant tests and tried them with Java 8, however currently they are ignored. See TODOs in ShouldOnlyBeAccessedByClassesThatTest.

Accessing a constant from an interface does not count as an access - is this as intended? Also see ShouldOnlyBeAccessedByClassesThatTest.java.

Resolves #28

### DIFF
--- a/archunit-example/build.gradle
+++ b/archunit-example/build.gradle
@@ -5,6 +5,7 @@ dependencies {
     compile 'org.apache.geronimo.specs:geronimo-jpa_2.0_spec:1.0'
 
     testCompile project(path: ':archunit-junit')
+    testCompile project(path: ':archunit', configuration: 'tests')
 }
 
 addTestJarTo this
--- a/archunit-example/src/main/java/com/tngtech/archunit/example/service/impl/ServiceImplementation.java
+++ b/archunit-example/src/main/java/com/tngtech/archunit/example/service/impl/ServiceImplementation.java
@@ -1,0 +1,6 @@
+package com.tngtech.archunit.example.service.impl;
+
+import com.tngtech.archunit.example.service.ServiceInterface;
+
+public class ServiceImplementation implements ServiceInterface {
+}
--- a/archunit-example/src/main/java/com/tngtech/archunit/example/service/impl/SomeInterfacePlacedInTheWrongPackage.java
+++ b/archunit-example/src/main/java/com/tngtech/archunit/example/service/impl/SomeInterfacePlacedInTheWrongPackage.java
@@ -1,0 +1,4 @@
+package com.tngtech.archunit.example.service.impl;
+
+public interface SomeInterfacePlacedInTheWrongPackage {
+}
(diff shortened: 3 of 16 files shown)


codecholeric:
Cool, thanks a lot :smiley: 
I would not add the `@Ignore`d test though, because the rest of the test (limiting the test to classes), is valid and useful right now. I would rather add the test without Java 8 specifics, but add a guard, that if we compile ArchUnit with Java 8 language level, the test will fail with a reminder, that the interface case should be added. Unfortunately something like
```
if (hasMethod(Collection.class, "stream")) {
    Assert.fail("Please extend the test case for accesses from interfaces " +
        "(which was not possible with Java 7)");
}
```
doesn't work, because even if we have source 1.7, we sometimes want to compile it with JDK 8, and then the test fails, even though it's not possible to use the language features. So maybe adding a TODO plus a guard in `build.gradle`
```
if (sourceCompatibility != JavaVersion.VERSION_1_7) {
    throw new IllegalStateException('Please fix TODOs marked with "TODO Java 8"')
}
```
would work (or do you have a better idea?)
Concerning your question, why the test case with the constant String doesn't work, the compiler inlines primitives and Strings, you can check this by
```
javap -v someInterface.class
```
and you'll see, that your String value is actually contained within the class file, thus on bytecode level, there is no access. You can fix this though, by making it an `Object`, or similar, because those won't be inlined (maybe we don't need the Java 8 guard then, either, because the test can just be written that way?)
Do you think, it would make sense, to add an `archunit-example` for this added syntax, together with an `archunit-integration-test`?

hajotka:
Indeed, when changing the example to access an Object field, the tests work. So no need for a Java 8 guard here.

I'll have a look at adding an integration test.

codecholeric:
Glad to hear, that it works without any shady workaround :smiley: 
I'll wait for the integration test and then merge the PR (in case you notice some problem while writing the test)

hajotka:
I've added examples + integration tests, hope they aren't too artificial ;)

There was a problem with the build in archunit-examples, a dependency was missing. How about running the examples as a smoke test in the CI?

codecholeric:
Ah, that dependency I've just removed a little while ago :frowning_face:
Yes, I haven't run `-P example`, yes it should be tested to compile (even though the tests fail by intention). Still, I don't think I want the dependency on `configuration: 'tests'` in the long run, since I'm pondering about creating a standalone example project (compare #30)

Anyway, the PR looks good, thank you so much :smiley: I'll merge that now...

codecholeric:
Note: I've removed the dependency in d4a36bd8a367d477aec3116d64e3b0588d1a0764, it was only needed to configure JUnit from Gradle (`@Category(Slow.class)`), instead I've just skipped the `excludeCategories`, if `Slow.class` is not on the classpath. Seems to work for the complete build as well as the example build...

codecholeric:
I've added the excludeCategories only where necessary, so I could remove the dependency again...

PR #43, opened 2017-11-12
Issue #41

### LINKED ISSUE #41: Extend syntax for simple class name handling
While one can do pretty much any filtering / assertion on class names using `haveNameMatching(regex)`, common use cases are to check the simple name for `endsWith`, `startsWith` or `contains`.
These cases should be directly represented within the syntax, without thinking in regex. Thus we should add `classes().`
* `that().haveSimpleNameEndingWith(..)`
* `that().haveSimpleNameStartingWith(..)`
* `that().haveSimpleNameContaining(..)`
* `should().haveSimpleNameEndingWith(..)`
* `should().haveSimpleNameStartingWith(..)`
* `should().haveSimpleNameContaining(..)`

and the respective negations (`notHave...`)

SteluHH:
Hi,

this PR implements everything that should be required in order to resolve #41.
Please review the changes. I was unsure whether to call the negation methods `haveXyzNot` or `notHave` as you suggested, since I had already started naming the methods `haveSimpleNamingNotEndingWith` and it was merged like that. I have no problems with renaming all methods respectively if wanted.

Stef

### DIFF
--- a/archunit-example/src/main/java/com/tngtech/archunit/example/AbstractController.java
+++ b/archunit-example/src/main/java/com/tngtech/archunit/example/AbstractController.java
@@ -1,7 +1,9 @@
 package com.tngtech.archunit.example;
 
+import com.tngtech.archunit.example.web.InheritedControllerImpl;
+
 /**
- * For demo purpose only, see
+ * For demo purpose only, see {@link InheritedControllerImpl}
  */
 public abstract class AbstractController {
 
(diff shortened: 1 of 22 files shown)


codecholeric:
Ah, thank you so much :smiley: 
The naming seems fine to me, I would have strove for consistency with `haveNameMatching`, and that is how you named them, so I fully agree :wink:
I'll look into that in detail during the next days (when I find the time) and get back to you / merge it...

codecholeric:
Looking good, thanks again :smiley:

PR #51, opened 2017-12-09
README improvements

hankem:
#### improved README.md
* streamlined 'TL;DR' section and added dependency line for gradle (for copy & paste convenience)
* simplified `Getting started' example (@Before-method is not needed and only distracts from ArchUnit example)
* concluded 'Getting started' section with link to archunit-example (which helps getting started, but not every reader might arrive at the last 'Where to look next' section)
* simplified 'Extending ArchUnit' section (also rephrasing 'one has to do' -> 'it can be done')
* streamlined 'License' section (while the reason for redistributing ASM & Guava might not be that relevant within this section, their actual licence might however be)
* added link to 'Where to look next' section (to guide the readers)
    
archunit-example/README.md:
* added links to the directories containing example rules and 'production' code (to keep the reader attracted)

#### examples import classes directly in field initialization instead of @Before method

----
> - [x] I hereby agree to the terms of the ArchUnit Contributor License Agreement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,14 @@ framework.
 
 ## TL;DR
 
-If you want to dive right into the first ArchUnit test using JUnit, follow these steps
+If you want to dive right into the first ArchUnit test using JUnit:
 
-1) ArchUnit can be obtained from Maven Central
+1) Get ArchUnit from Maven Central
+* with gradle:
+```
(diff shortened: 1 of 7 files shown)


codecholeric:
Thanks for the adjustments, you're right, importing the classes within `@Before` doesn't add much benefit :wink:

PR #56, opened 2017-12-28
@ArchIgnore annotation allows to specify a reason why the test is ignored

hankem:
Since `@ArchIgnore` corresponds to `@org.junit.Ignore`, it might also be useful to be able to add a reason when an `@ArchTest` is ignored.

---
> I hereby agree to the terms of the ArchUnit Contributor License Agreement.


codecholeric:
Makes sense, thanks a lot :smiley:

PR #68, opened 2018-05-11
Fixed missing `apple.security` to SecurityTest. Resolves #67

### LINKED ISSUE #67: SecurityTest fails with mac JVM
The SecurityTest tests if `java.security.cert` is accessed only by certain packages. On Mac jvms `apple.security` is also accessing the cert package.

DennisRippinger:
Also fixes a problem related to running JUnit Tests from IntelliJ on Mac systems. IntelliJ resides in a folder `IntelliJ App` (mind the space)  which executes the JUnit tests and adds therefore its own Jars to the classpath. When converted from an URL into an URI we recieve an illegal character exception. The fix was to UrlEscape the path variable within the `newUrl` method.


codecholeric:
Thank you so much :smiley: 
I actually had noticed that problem with URLs with spaces a while ago and forgot to fix it, so I'm happy you brought it up again!! (I'm gonna add a small test for the bug, to prevent regression...)

PR #85, opened 2018-07-01
various dependencies updates

based2:


codecholeric:
Thanks for the PR, however the project cannot be built anymore with those changes. I would be great if you could

* either upgrade only the dependencies that work out of the box
* or fix the problems that arise from upgrading those dependencies that cannot just be upgraded by a simple version update

Also note, that this project has to be compatible to Java 7, so you have to make sure that your upgrades are still compatible.
If you get your PR working, it would also be great if you could sign your commits (see https://github.com/TNG/ArchUnit/blob/master/CONTRIBUTING.md). If you do `git commit -s --amend` on your fix, it should solve that.
You can always check if your changes work, by running `./gradlew build` on the command line :wink:

codecholeric:
Ping :wink:

codecholeric:
I've updated the dependencies in a compatible way myself.

PR #88, opened 2018-07-10
Support comments in archunit_ignore_patterns.txt

bedla:
Hi,
it would be nice to comment lines with description of ignored patterns. Pls take a look and say what do you think. Looks like that "# " is good enough :]
Thx 
Ivos

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/lang/ArchRule.java
+++ b/archunit/src/main/java/com/tngtech/archunit/lang/ArchRule.java
@@ -73,6 +73,7 @@ public interface ArchRule extends CanBeEvaluated, CanOverrideDescription<ArchRul
         }
 
         static final String ARCHUNIT_IGNORE_PATTERNS_FILE_NAME = "archunit_ignore_patterns.txt";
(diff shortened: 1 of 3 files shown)


codecholeric:
Wow, thanks for your next PR :smiley: 
Looks good to me, the only thing I'm wondering is if it's necessary to enforce that whitespace after # ?
I guess most people will add it anyway, but that the line suddenly doesn't count as comment anymore, if I write `#ignore that line` seems kind of surprising?

hankem:
I'd also assume that `#` should be sufficient for a comment.

bedla:
i have looked how regexp buddy solves comments and just copied it. no problem to change it ;]

codecholeric:
Thanks for the adjustment! I've made a small change in the test, because I like consistency (even though, especially with Java 8, the AssertJ `assertThatThrownBy` approach is definitely nicer :wink:; with Java 7, cause of all the boilerplate, I have no strong opinion)
Also I've added this to the user guide.
Thanks a lot for contributing!! :smile:

bedla:
thx for merge. i am happy to help :)

PR #91, opened 2018-07-19
fixed class hierarchy resolution

svenackermann:
Hi,
I encountered a problem with the resolution of type hierarchies, where a base class in a quite deep hierarchy from a framework was not resolved correctly. 
I debugged this a lot and think this change should fix it. It seems quite obvious to me that this is wrong, so I decided to just provide a minimal pull request. It's obvious because resolving only the interfaces for a superclass does not add its own superclass-hierarchy to the context. So, it can happen that not the whole hierarchy is resolved.
I would also provide a test case for that, but unfortunately I'm not able to git push at work. I did this change with github's online editor. I can also not build/test this here, so please excuse that I did not completely follow the contribution guidelines. Maybe you can just see this as a bug report with a solution suggestion.

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/importer/ClassGraphCreator.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/importer/ClassGraphCreator.java
@@ -128,7 +128,7 @@ class ClassGraphCreator implements ImportContext {
 
     private void resolveInheritance(String currentTypeName, Function<JavaClass, Set<String>> inheritanceStrategy) {
         for (String parent : inheritanceStrategy.apply(classes.getOrResolve(currentTypeName))) {
-            resolveInheritance(parent, interfaceStrategy);
+            resolveInheritance(parent, inheritanceStrategy);
         }
     }
 


codecholeric:
Thank you so much for debugging and reporting this!! This is indeed an error, I guess it went unnoticed for so long, because it only shows if you have deeply nested super classes missing from your import. I've taken your fix over in https://github.com/TNG/ArchUnit/commit/76dab49d1ec0daa7aa33642be0d8c50b17e90263, added a test for this and could reproduce it (which sufficiently many levels of hierarchy and the classes missing from the original import :wink:)
I've fixed this here: https://github.com/TNG/ArchUnit/releases/tag/v0.8.3 
As soon as the sync with Maven Central is through (prolly 1 or 2 hours), you can pull in the new version `0.8.3`.
So thank you again for providing such a precise bug report :smiley:

codecholeric:
BTW: A quick "it's fixed with 0.8.3" would be appreciated, I'll keep this PR open until then :wink:

codecholeric:
It's through :wink: 
https://search.maven.org/#search%7Cga%7C1%7Carchunit

PR #93, opened 2018-08-03
Supports running on Java Early Access versions

jqno:
I'm trying to run my unit tests on an Early Access version of Java 11 (gotta be prepared 😉), and my ArchUnit test fails while attempting to parse the version string:
```
java.lang.ExceptionInInitializerError
Caused by: java.lang.NumberFormatException: For input string: "11-ea"
````
This PR should fix that.

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/PluginLoader.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/PluginLoader.java
@@ -150,7 +150,7 @@ public class PluginLoader<T> {
         };
 
         private static final Ordering<JavaVersion> FROM_NEWEST_TO_OLDEST_ORDERING = Ordering.explicit(JAVA_9);
-        private static final Pattern VERSION_PATTERN = Pattern.compile("([^.]+).*");
+        private static final Pattern VERSION_PATTERN = Pattern.compile("([^.-]+).*");
 
         public abstract boolean isLessOrEqualThan(String version);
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jqno** [comment]
(I hope it's ok that I used a secondary e-mail address in the sign-off, I don't like leaving my main one in public places on the internet. I do answer this one.)

**Comment by codecholeric** [comment]
Thanks a lot for your contribution :smiley: 
I've made a little adjustment, by just matching the first group of digits (doing `[^.]` was obviously a little short-sighted :wink:). It's kind of what the method name (`parseFirstGroupOfJavaVersion`) suggests anyway.
Now at least either the `matches` and the `parseInt` should work, or the `matches` should already fail.
How fast do you need this released? I'm gonna release 0.9.0 within the next 2-3 weeks I think, if this patch is vitally important for you, I would do a minor release this weekend.

**Comment by jqno** [comment]
Thanks for the quick response! The adjustment seems like a good idea. A release in the next 2-3 weeks is totally fine for me.

**Comment by codecholeric** [comment]
PS: It's hacky :wink:, but as a workaround you could probably wrap the class import into sth. like

```
try {
    System.setProperty("java.version", "11");
    classFileImporter.import... //
} finally {
    System.setProperty("java.version", oldVersion);
}
```

jqno:
I just noticed that the release has taken place. Just wanted to confirm that it does indeed work. Thanks!

(Also: looking forward to seeing your talk at JavaZone ;) )

PR #98, opened 2018-08-15
Filter-out synthetic and bridge members (ctors, methods, fields)

bedla:
Hi, 
I am again playing with ArchUnit and found that it would be helpful for my rules to have synthetic and/or bridge methods filtered out. Basically my usecase was similar to SyntheticBridgeMethodsClass class. I have filtered out all members in PR, but now few tests are failing. From my point of view synthetic/bridge methods are "useless" because they are not in source code and If i understand ArchUnit philosophy correctly it is more about what we see than what could be generated, right? So I think that those failing test-cases should be rewritten... what is your opinion? Am I on right track or am I missing something :)
Thx a lot
Ivos


codecholeric:
I'll look at it as soon as possible. Principally yes, I would say ArchUnit cares about the part of the code that's relevant to humans to understand and compose maintainable systems, so synthetic members should be kind of irrelevant :wink: 
I don't understand why tests would fail though, I don't think there are any tests based on synthetic members?
What I can tell from a quick glance is, that I would not extend `JavaModifier` for `BRIDGE` and `SYNTHETIC`, since `JavaModifier` is part of ArchUnit's public API and we both agree, that `BRIDGE` and `SYNTHETIC` do not belong there :wink: So it would be modifiers that are irrelevant for any user of ArchUnit, I'd try to filter out those members using only `ACC_...` from ASM.

bedla:
all fixed :]

it seems that we need to preserve synthetic ctors to correctly match ctor calls between visibilities.

eg. i have class witch private ctor
```
package com.tngtech.archunit.lang.syntax.elements;

import com.tngtech.archunit.lang.syntax.elements.ShouldOnlyBeAccessedByClassesThatTest;

private static class ShouldOnlyBeAccessedByClassesThatTest.ClassAccessedByPublicClass {
    private ShouldOnlyBeAccessedByClassesThatTest.ClassAccessedByPublicClass() {
    }

    /* synthetic */ ShouldOnlyBeAccessedByClassesThatTest.ClassAccessedByPublicClass(ShouldOnlyBeAccessedByClassesThatTest x0) {
        this();
    }
}
```
which is called by
```
package com.tngtech.archunit.lang.syntax.elements;

import com.tngtech.archunit.lang.syntax.elements.ShouldOnlyBeAccessedByClassesThatTest;

public static class ShouldOnlyBeAccessedByClassesThatTest.PublicClass {
    void call() {
        new ShouldOnlyBeAccessedByClassesThatTest.ClassAccessedByPublicClass(null);
    }
}
```

compare them with original source code...

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaModifier.java
```diff
@@ -46,7 +46,11 @@
     @PublicAPI(usage = ACCESS)
     SYNCHRONIZED(EnumSet.of(ApplicableType.METHOD), Opcodes.ACC_SYNCHRONIZED),
     @PublicAPI(usage = ACCESS)
-    NATIVE(EnumSet.of(ApplicableType.METHOD), Opcodes.ACC_NATIVE);
+    NATIVE(EnumSet.of(ApplicableType.METHOD), Opcodes.ACC_NATIVE),
+    @PublicAPI(usage = ACCESS)
```
bedla: not sure about bridge (they are filteredout), but synthetic ctors are visible for archunit

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterTest.java
```diff
@@ -1857,6 +1871,115 @@ public void is_resilient_against_broken_ClassFileSources() throws MalformedURLEx
...
+    public void synthetic_class() throws Exception {
+        thrown.expect(NullPointerException.class);
+        thrown.expectMessage(containsString("No object with name"));
+
+        final ImportedClasses javaClasses = classesIn("testexamples/syntheticbridge");
+        javaClasses.get("com.tngtech.archunit.core.importer.testexamples.syntheticbridge.SyntheticConstructor$1");
+    }
+
+    @Test
+    public void synthetic_class_switch_map() throws Exception {
```
bedla: Found testcase when synthetic switchmap class is created. It is ok with current changes, class is loaded into com.tngtech.archunit.core.importer.ImportedClasses#additionalClasses and correctly skipped.

codecholeric:
Sorry that I haven't gotten back to this PR for such a long time!! I still want to include this into the next release, so thanks for writing those test cases for bridge methods, etc.!! 
I just need some time to really look into that, see if I have to create some more examples where synthetic methods/constructors are created so I can be sure. But I'll make sure to force take that time ASAP :wink:

bedla:
np, take your time :] it is imo "big" change, for questions just ping me...

codecholeric:
I've finally managed to check this out :wink: Sorry for the long delay. I've somewhat refactored the tests, because the biggest fear I had was to lose dependencies. But anyway, it looks good, I can't think of any other case and all cases I could think of (which was pretty close to the ones you already added) behaved correctly.
I've removed the switch map case though, because I really don't see any need to specify this. If anything, people would want to skip that call to `ordinal()` and `$SwitchMap$`, but I don't see any value in asserting this behavior. Or in other words, if that behavior would change, or those calls would vanish, I wouldn't care and neither do I care about the current behavior, so I don't really see the need for the test :wink: 

So to conclude, I'm gonna merge this for the next version, thanks a lot for the initiative :smile: I think this will make the behavior a little less surprising in certain cases!

bedla:
cool thx for merge :) it is pleasure to help

codecholeric:
Unfortunately now I encountered some problems, guess my original hunch that this is not completely trivial was right :frowning_face: I had to revert the change for now.
Because it's not good enough to just filter those classes and members from the import, we also need to filter out accesses to such classes or the behavior will turn out strange.
I've tried to create a fix, where encountered synthetic members would also be recorded for later and then accesses to those members will be thrown out. But that turned out to cause other problems, because in some cases the compiler creates synthetic methods in derived classes and then those accesses were missing (and I rather have some annoying synthetic info than stuff really missing).
I think we really have look into that closer if we want to fix that issue. In particular how we can filter synthetic targets and still not lose information.
Also methods like `enum.values()` are not synthetic, and neither is `Array.clone()`, so the thing that caused the most confusion wouldn't even be covered for this.
Anyway, I'm not gonna include it in this release because I don't want to break things.

codecholeric:
I've created a new branch with our changes in place, including my latest attempt to fix the problem, if you want to take a look: https://github.com/TNG/ArchUnit/tree/filter-synthetic-code

PR #102, opened 2018-08-19
Use Maven Surefire 2.22.0 in example installation

sormuras:
2.21.0 has some issue when using fork mode and other bugs. They are fixed in 2.22.0.


codecholeric:
Thanks a lot, you're right, in my `archunit-integration-test` project I'm already using 2.22.0, forgot to update this in the user guide.
I'll update the website immediately, too.

PR #105, opened 2018-08-28
Junit4-support for @ArchTest instance fields or static methods in abstract base classes

### LINKED ISSUE #104: @ArchTest instance fields or static methods in abstract base classes
#97 has removed the necessity to declare `@ArchTest`s as static fields or methods.

`archunit-junit4:0.9.1` however fails with 
```
com.tngtech.archunit.junit.ArchTestInitializationException: java.lang.InstantiationException
```
on [`@ArchTest` (instance fields or non-static methods) in (a base class that cannot be instantiated itself, e.g. when the base class is abstract or does not have a default constructor)].

Example:
```
    @AnalyzeClasses(packages = "some.pkg")
    public class ArchTestWithAbstractBaseClass extends AbstractBaseClass {
    }

    abstract class AbstractBaseClass {
        @ArchTest
        ArchRule abstractBaseClassInstanceField = /*...*/

        @ArchTest
        void abstractBaseClassNonStaticMethod(JavaClasses classes) {
            /*...*/
        }
    }
```

The same holds true for `ArchRules.in` a class with such a base class.

Even though such constellations can probably be avoided (we all favor composition over inheritance, don't we...?), ArchUnit should try to instantiate the correct class (`ArchTestWithAbstractBaseClass` instead of `AbstractBaseClass` in the example above).

(Since I'm not experienced with Junit 5, I cannot judge yet whether `archunit-junit5` has similar issues.)

hankem:
Resolves #104

### DIFF
--- a/archunit-junit/junit4/src/main/java/com/tngtech/archunit/junit/ArchRuleDeclaration.java
+++ b/archunit-junit/junit4/src/main/java/com/tngtech/archunit/junit/ArchRuleDeclaration.java
@@ -34,22 +34,24 @@ import static com.tngtech.archunit.junit.ReflectionUtils.withAnnotation;
 abstract class ArchRuleDeclaration<T extends AnnotatedElement> {
     private final Class<?> testClass;
     final T declaration;
+    final Class<?> owner;
     private final boolean forceIgnore;
 
-    ArchRuleDeclaration(Class<?> testClass, T declaration, boolean forceIgnore) {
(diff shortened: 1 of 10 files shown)


codecholeric:
I split the tests a little, since there are three different test classes that deal with the runner running fields, methods and rule sets respectively. Can argue if that split is so great, but as long as it's there, I think each aspect should be sorted in the respective test class (e.g. no test in `ArchUnitRunnerRunsFieldsTest` dealt with `ArchRules.in(..)` so far)
Besides that, thanks a lot!! :smile: Gonna merge it, as soon as Travis is through.

PR #122, opened 2018-11-19
Fixed sample code for Kotlin

dais39:
Hi. 
I was reading ArchUnit user guide and found that sample code for kotlin had wrong syntax.
```
ArchRule rule = ArchRuleDefinition.noClasses().should()...
```
This PR fixed this.


codecholeric:
Thanks a lot for your contribution, I'll happily merge this :smiley:

PR #133, opened 2018-12-14
Update CONTRIBUTING.md

alanktwong:
Update CONTRIBUTING.md with clear example of how to build the project locally for contributor onboarding

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,21 +4,18 @@ Contributions are very welcome. The following will provide some helpful guidelin
 
 ## How to build the project
 
-Assumptions
-
-* Java 9
-
-The following is just an example input/output from a Unix command line. Windows users should use `gradlew.bat` instead
+ArchUnit requires at least JDK 9 to build. The following is just an example input/output from a 
+Unix command line. Windows users should use `gradlew.bat` instead.
 
 ```
-$ java --version
-openjdk 9.0.4
-OpenJDK Runtime Environment (build 9.0.4+11)
-OpenJDK 64-Bit Server VM (build 9.0.4+11, mixed mode)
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -2,6 +2,36 @@
...
 Contributions are very welcome. The following will provide some helpful guidelines.
 
+## How to build the project
+
+### Building locally on an OSX machine
+
+Assumptions
+
+* Java 9
+* Gradle
```
hankem: Using the [Gradle Wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html), which is [part of the repository](https://github.com/TNG/ArchUnit/tree/master/gradle/wrapper), should make the build independent of an external Gradle installation.

But I like the idea to state that JDK 9 is a prerequisite! :+1:

As the build should not depend on the platform (OSX vs. Linux vs. Windows), I would however remove the different OS-specific sections (and the `TBD`s as well 😉).

codecholeric:
Thanks a lot for raising this! I agree with @hankem, that Gradle should not be a requirement. Also there should be no difference in the platform except having the right JDK. But yes, stating that you need Java 9 and how to build the project via Gradle is certainly a good idea. 
I'll add some further thoughts and get back to you...

alanktwong:
Thanks. I'll change the PR

codecholeric:
I've adjusted the section a little, because `java --version` does not necessarily tell, which JDK Gradle is using (could run Gradle with JDK 10 but configure JDK 9 to compile, might depend on `gradle.properties` or `JAVA_HOME`, the usual complexity...)
Do you think this is clear enough to understand?

codecholeric:
Gonna merge this now, thanks again :smiley:

PR #134, opened 2018-12-16
Consider classes referenced in throws-clauses

### LINKED ISSUE #126: DependenciesFromSelf does not consider classes referenced in throws-clause
With f761154228c4c2c4b209134073c6a4019b306d67 JavaClass's method [getDirectDependenciesFromSelf()](https://github.com/TNG/ArchUnit/commit/f761154228c4c2c4b209134073c6a4019b306d67?diff=unified#diff-b5c0368e6ab86b282bfd74b185de72a4) was already extended to take many dependencies into account. However, exception classes referenced in the throws-clauses of the class' methods are not yet considered.

We currently have an ArchUnit-test implemented like this
```    
@ArchTest
public static final ArchRule daosShouldNotExposeSql = noClasses().that()
                                                                 .haveSimpleNameContaining("DAO")
                                                                 .and()
                                                                 .areInterfaces()
                                                                 .should()
                                                                 .dependOnClassesThat()
                                                                 .resideInAPackage("java.sql..");
```

Looking at an example DAO implementation, the first two methods fail the test, while the third doesn't.

```
public interface ExampleDAO
{
    java.sql.ResultSet returns();


    void consumes(java.sql.PreparedStatement statement);


    void throwsException() throws java.sql.SQLException;
}
```

tngwoerleij:
Resolves #126 

Feedback is greatly appreciated. Just started with ArchUnit from a user perspective so my change might miss important parts.


codecholeric:
Thanks a lot for this extensive PR :smile: Looks very good to me!!
I'll add some review comments in a second...
If you feel like it, you could add some `archunit-example` showing some use case for the new API :smiley:

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaCodeUnit.java
```diff
@@ -43,10 +44,11 @@
  * in particular every place, where Java code with behavior, like calling other methods or accessing fields, can
  * be defined.
  */
-public abstract class JavaCodeUnit extends JavaMember implements HasParameterTypes, HasReturnType {
+public abstract class JavaCodeUnit extends JavaMember implements HasParameterTypes, HasReturnType, HasThrowsTypes {
     private final JavaClass returnType;
     private final List<JavaClass> parameters;
     private final String fullName;
+    private final List<JavaClass> throwsTypes;
```
codecholeric: I would call this `throwsDeclarations` and I would not make this of type `List<JavaClass>`, but `List<ThrowsDeclaration>`. A ThrowsDeclaration can offer `JavaClass ThrowsDeclaration.getType()`, but I don't think it should be modeled as a `JavaClass` (which it semantically is not).
I made the same mistake with `parameters` being of type `List<JavaClass>`, now I want to introduce parameter annotations and it's hardly possible without breaking the API :frowning_face: 
The reason in the end is always the same though, type mismatch. A parameter is not really a `JavaClass` and neither is a throws declaration.

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaCodeUnit.java
```diff
@@ -69,6 +72,11 @@ public JavaClassList getParameters() {
         return new JavaClassList(parameters);
     }
 
+    @Override
+    public JavaClassList getThrows() {
```
codecholeric: I would call this `getThrowsDeclarations()` and return `ThrowsDeclarations implements Iterable<ThrowsDeclaration>` (compare statement above)

archunit/src/main/java/com/tngtech/archunit/core/domain/properties/HasThrowsTypes.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.List;
+
+import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
+import static com.tngtech.archunit.base.DescribedPredicate.equalTo;
+import static com.tngtech.archunit.core.domain.Formatters.formatMethodParameterAndThrowsTypeNames;
+import static com.tngtech.archunit.core.domain.JavaClass.namesOf;
+import static com.tngtech.archunit.core.domain.JavaClassList.GET_NAMES;
+
+public interface HasThrowsTypes {
```
codecholeric: I would only introduce this interface if there is more than one reasonable class implementing it. At the moment I only see `JavaCodeUnit`. I think it would make sense to make `CodeUnitCallTarget` implement this interface, too?
The reason for the existence of `CodeUnitCallTarget` (as target of a method / constructor call) as opposed to a `JavaCodeUnit` is the same as above, a call target is not the same as a method / constructor. Because within the bytecode there will be declared a call to something like `SubFoo.bar()`, but maybe `SubFoo` doesn't declare method `bar()`, but instead it's super class `Foo` does. And maybe some interface `IFoo` declares a method `bar()`, too. So it's not immediately clear, which actual `JavaCodeUnit` belongs to a call target. Nevertheless we can already derive some things like what parameters the target has.
`CodeUnitCallTarget` offers a method `resolve()` though, that tries to find the respective method that matches the target. So we could implement `HasThrowsDeclarations` (:wink:) here, too, and pipe to `resolve().getThrowsDeclarations()`. If multiple methods match `resolve()` (the case of several interfaces declaring the same method and no class implementing it up to the part in the hierarchy that is referenced) it seems like throws declarations should be intersected (if you have interface `A` and `B` both declaring `foo`, but with different throws declarations, then `C extends A, B` can per se neither declare the one or the other checked exception)
What do you think?

archunit/src/main/java/com/tngtech/archunit/core/importer/DomainBuilders.java
```diff
@@ -206,6 +206,7 @@ JavaField construct(JavaFieldBuilder builder, ClassesByTypeName importedClasses)
     public abstract static class JavaCodeUnitBuilder<OUTPUT, SELF extends JavaCodeUnitBuilder<OUTPUT, SELF>> extends JavaMemberBuilder<OUTPUT, SELF> {
         private JavaType returnType;
         private List<JavaType> parameters;
+        private List<JavaType> exceptions;
```
codecholeric: I would already do the translation here. ASM has `exceptions` but `JavaCodeUnitBuilder` should already have `throwsDeclarations`

archunit/src/main/java/com/tngtech/archunit/core/importer/DomainBuilders.java
```diff
@@ -219,15 +220,28 @@ SELF withParameters(List<JavaType> parameters) {
             this.parameters = parameters;
             return self();
         }
+        
+        SELF withExceptions(List<JavaType> exceptions) {
```
codecholeric: Same here, just because ASM calls it `exceptions` doesn't mean we have to pull this into `JavaCodeUnitBuilder`

archunit/src/main/java/com/tngtech/archunit/core/importer/JavaClassProcessor.java
```diff
@@ -225,6 +227,16 @@ public MethodVisitor visitMethod(int access, String name, String desc, String si
         return result;
     }
 
+    private List<JavaType> typesFrom(String[] exceptions) {
+        List<JavaType> result = new ArrayList<>();
+        if (exceptions != null) {
```
codecholeric: This really ever null? I hate those APIs :wink:

tngwoerleij: Yes 😕 See https://asm.ow2.io/javadoc/org/objectweb/asm/ClassVisitor.html#visitMethod-int-java.lang.String-java.lang.String-java.lang.String-java.lang.String:A-

codecholeric: Maybe it's for performance reasons, I still hate those API from an user's point of view...

tngwoerleij:
I considered all your comments, but please have a close look as I was not always sure if I am doing the right thing; especially with CodeUnitCallTarget implementing HasThrowsDeclaration.

Two things are still missing: 

* There are no test cases for the enhanced CodeUnitCallTarget.
* archunit-example is still missing.

I would like to take care of those two points as soon as the other changes are in an acceptable state. Afterwards I would squash all commits into one.

codecholeric:
I've looked over your changes, looks good to me, I added some more review commits

* first of all I found that the expression "throws clause" seems to be a good match for the collection of throws declarations :wink: 
* I also removed the implementation of `List` from the renamed `ThrowsClause`, because I don't think `List` is the API we want (remove makes no sense and arguably `get(i)` doesn't either, especially with the implementation at `CodeUnitCallTarget`; and yes, you can argue that this also holds for `JavaClassList`, my bad that's already public API :wink:)
* I also thought it would be a nice API if one could navigate back from `ThrowsDeclaration` to `JavaCodeUnit`. But that honestly let me down a rabbit hole deeper and deeper. Expecially since the location where a `ThrowsClause` is declared is not necessarily a `JavaCodeUnit`, now that `CodeUnitCallTarget` also implements `HasThrowsClause` :wink: **So I'd definitely be happy about some feedback here, if this is too crazy now!!** Other than that, I've kept your implementation there. Of course it got more complicated now that I wanted to navigate back from the `ThrowsDeclaration` and suddenly there is not necessarily a `JavaCodeUnit` to go to. So I thought the only consistent way here was to always use `CodeUnitCallTarget` as the 'location' where the `ThrowsClause` originates from, even if the resolution would be unique. In the end a caller can still call `resolve()` themselves, but then has to handle possible ambiguity as well.
I've added tests for the throws clause at `CodeUnitCallTarget`, you can check out those as well.

As for the example, what was your original use case? Maybe you can abstract that and add it to `archunit-examples`. If it was just about dependencies, we could just extend one dependency example, but I think some example explicitly asserting something about a throws clause would be neat (like "methods throwing IOException should be annotated with SafeIO" or whatever).

And yes, squash whatever you want (you can squash my commits, too, I don't care), the only thing I care about is that a) the commits are reasonable for research in the long term (i.e. I don't care for 10 commits refactoring back and forth, just reasonable units) and b) I can do a fast forward merge (i.e. you have to rebase your branch against upstream master; I've done this and force pushed right now, so if you're quick, you should be good to go :wink:)

tngwoerleij:
Thanks for the effort you put into reviewing and improving my code! I hope it wasn't a bigger effort for you than to write it yourself completely ;-) 

I'm perfectly fine with your changes including adding the location to ThrowsDeclaration.

I added my original use case to DaoRulesTest. Please have a quick look at it. Would you like another use cases included for demonstration?

I also squashed my commits fixing your review findings into my very-first commit.

codecholeric:
That example looks good to me :+1: And useful in daily life.

Don't worry, I would have started exactly the way you did and have gone the same way, so you definitely saved me a lot of time, even if I have added some more commits :wink: And in my opinion the result is always better, if more than one brain puts thoughts into it. 

So thanks a lot, I think this will make a very useful extension :smiley:

PR #138, opened 2018-12-21
Reimplement cycle detection using Johnson algorithm.

torfmaster:
# Summary
In this pull request I replace the BFS for cycle detection by Johson's algorithm (Siam Journal on Computating, Vol. 4, No. 1, March 1975). This establishes the best asymptotic bound for complexity known to me, namely O( (v+e)*c) where 
 - v is the number of vertices
 - e is the number of edges
 - c is the number of cycles.

The detection is run after constructing the full dependency graph. This allows limiting the number of results by interrupting the computation of cycles, not the construction of the graph.

# Implementation:
The algorithm is implemented in the exact way it is described in the paper. In order to reduce the constants all access operation are performed on an indexed copy of the graph.

# Tests:
Existing test are passing. In contrast to Tarjan's algorithm which reports only strongly connected components the crucial requirement on the algorithm here was to find *all* cycles (also subcycles). A test case was added to GraphTest. The Tarjan implementation used internally does not have tests on its own as it is not exposed as an independent component. If one would decide to reuse the component one would have to equip it with an adequate API and test it independently of the CycleFinder.

# Possible extensions
 - benchmarking on graph-level
 - benchmarking with sample projects


codecholeric:
Puh, as discussed, I think we're finally almost there :smile: I've looked into it an replaced some structures to use more primitives, would be great if you could cross-review my commits!

codecholeric:
Thank you so much for your hard work!! :smile: I've extended the commit message a little to emphasize how we map to a primitive structure in the beginning, then run the algorithm and then map back the result, other than that :+1:

hankem:
As the commit message of 8e75b91c068b99b4ccc417c7d82c58d186b6c04e had made me curious, I started the test with the old code and saw that it finished only after 54 min – the new code, in contrast, takes just 4 s; that's impressive! 😃

SlicePerformanceTest

codecholeric:
Haha, I never made it to the end :wink:

PR #139, opened 2019-01-06
#114: Expose if JavaClass is an array

### LINKED ISSUE #114: Expose if a JavaClass is an array
It is currently not possible to check if a `JavaClass` is an array and to obtain the type of the array.

So for a class like this:
```
public class Foo {
    public Bar[] getBars() { ... };
}
```

getting the return type of the method `getBars` should allow to query if it is an array and then I would like to query for the type of the Array (`Bar`).

wengertj:
This commit partially resolves #114 by adding a `isArray` method to `JavaClass`. 

Getting the return type of the array as mentioned by @wolfs in the issue is a different story as it  probably requires a more "gerneral" kind of api - not just related to arrays (e.g what should such a method return? A simple String or a more powerful kind of abstraction?)

Thanks for looking at this and for your awesome work 👍

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
@@ -149,6 +149,11 @@ public class JavaClass implements HasName, HasAnnotations, HasModifiers {
     }
 
     @PublicAPI(usage = ACCESS)
+    public boolean isArray() {
+        return javaType.isArray();
+    }
+
+    @PublicAPI(usage = ACCESS)
     public boolean isInnerClass() {
         return enclosingClass.isPresent();
     }
--- a/archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTest.java
@@ -21,6 +21,7 @@ import com.tngtech.archunit.core.domain.testobjects.AhavingMembersOfTypeB;
 import com.tngtech.archunit.core.domain.testobjects.AllPrimitiveDependencies;
 import com.tngtech.archunit.core.domain.testobjects.B;
 import com.tngtech.archunit.core.domain.testobjects.InterfaceForA;
+import com.tngtech.archunit.core.domain.testobjects.IsArrayTestClass;
 import com.tngtech.archunit.core.domain.testobjects.SuperA;
 import com.tngtech.java.junit.dataprovider.DataProvider;
 import com.tngtech.java.junit.dataprovider.DataProviderRunner;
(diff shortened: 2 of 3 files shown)


codecholeric:
Thanks a lot for your contribution!! :smiley: 
Can you do me one favor though and sign-off your commit? (compare https://github.com/TNG/ArchUnit/blob/master/CONTRIBUTING.md#commits) 
I think a simple `git commit -s --amend` followed by a force push should do it.

Regarding the component type of the array in question, I would probably model it along the Java Reflection API. I.e. add a method `JavaClass.getComponentType()`, add clear JavaDoc and either return the respective component type if it's an array or throw an exception if it is not (I don't like the Reflection API returning null, I'd rather have people check first and only retrieve the component type if it makes sense. Or make the method return `Optional`, but that kind of makes no sense if one has checked if it's an array before anyways, so I don't think I'd like that API either).

wengertj:
Ooops 🙈 Of course - I will 👍 

Thanks for the explanation - I was not aware of the `getComponentType()` method in the Reflection API. I will try to implement your suggestion within the next couple of days :)

wengertj:
I've added the `getComponentType()` method according to your suggestions. If my solution works for you then I think this PR can be merged 👍

wengertj:
Ok that is very understandable - I will dig further into this 👍 
What I would propse though is that I revert my last commit regarding `getComponentType()` and you only merge the `isArray()` part. I would then open a new PR with an enhanced `getComponentType()` solution (in case I am able to find one 😆 )

wengertj:
Just one quick follow up: 
`JavaType` does have a method `tryGetComponentType()`. Would this be a more robust approach? In the private class `ArrayType` `tryGetComponentType()` is already implemented.
So something like this?:
```java
public Class<?> getComponentType() {
        if (isArray()) {
            // Might need checks before calling get() and throw an exception if 
           // no value is present
            return javaType.tryGetComponentType().get().resolveClass();
        } else {
            throw new IllegalArgumentException(
                    String.format("Cannot get the componentType - %s is not an array", toString()));
        }
    }
```

codecholeric:
`JavaType.tryGetComponentType()` should be a good place to start, just that the signature of the getter within `JavaClass` should be `JavaClass getComponentType()`, which is probably where it gets trickier. At some place you'll need the mapping className -> `JavaClass` (since `JavaType` has IMHO not much more usable information than that). So the flow will probably be `JavaType` -> `tryGetComponentType()` -> ifPresent `allClasses.lookUp(componentType.getName())`

Probably `JavaClass` needs a new field `Optional<JavaClass> componentType` that can be filled on class construction by looking up `JavaType.getName()` within the map of all imported classes (during import there should exist some map FQN -> `JavaClass`)

wengertj:
Thanks a lot for the hints 😄 
I will still revert my last commit though so that you can merge the `isArray()` method - the `getComponentType()` will propably take me some time.

wengertj:
I reverted my last commit. Feel free to merge the PR - I need a bit of time for the componentType 😅

PR #157, opened 2019-03-13
Introduce and use DescribedPredicateAssertion

hankem:
This does not add a feature to the ArchUnit library, but simplifies its internal tests.

DescribedPredicateAssertion allows to encapsulate

            assertThat(predicate.apply(value1)).as("predicate matches").isTrue();
            assertThat(predicate.apply(value2)).as("predicate matches").isFalse();
            assertThat(predicate.getDescription())).as("description").isEqualTo(description);

in

            assertThat(predicate).accepts(value1).rejects(value2).hasDescription(description);

I've created the pull request for the current `add_member_syntax` branch (where it has been developed), but its actually not related to Issue 38.
You could thus probably also merge it to `master` after `add_member_syntax` has been merged.

### DIFF
--- a/archunit/src/test/java/com/tngtech/archunit/base/DescribedPredicateTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/base/DescribedPredicateTest.java
@@ -25,89 +25,96 @@ public class DescribedPredicateTest {
 
     @Test
     public void alwaysTrue_works() {
-        assertThat(alwaysTrue().apply(new Object())).isTrue();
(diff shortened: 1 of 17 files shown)


codecholeric:
Thanks a lot for all that work :smile: That really makes the tests more readable!
I plan on merging this branch back into master soon anyway, so I think it's good enough to merge this into the current branch for now (I'm gonna rebase and force push though before I merge, so the origin repo will be lost anyway...)

PR #159, opened 2019-03-15
Extended domain model in user guide for packages

goetzd:
This pull request extends the domain model graph in the user guide so that it also contains the `JavaPackage` class and its relation to `JavaClass`.


goetzd:
I just saw that the dimensions of all image tags in the generated `000_Index.html` file changed slightly, which seems weird. I would have expected the height of `domain-overview.png` to change, but nothing else.

Do you think we should revert all the other changes?

goetzd:
Ok, I found out why the dimensions changed: Apparently, the fonts settings used for the diagrams are rendered with different antialiasing settings on different machines. The font on my system looks slightly thicker/blurrier than the one that is currently used in the pngs in the repository. Because of this, the image dimensions also change slightly which is reflected in the `width` and `height` of the `img` tags.

Since I only committed `domain-overview.png` the images now look inconsistent.

I could commit the remaining images, but I actually think that the font looks slightly too blurry on my machine.

@codecholeric: Could you regenerate `domain-overview.png`?

codecholeric:
Thanks for researching that weirdness :wink: And of course the adjustment! I've re-rendered the user-guide / diagrams. I wish GitHub pages would whitelist the Jekyll asciidoctor-diagrams Gem, that would make all this so much easier :frowning_face:

codecholeric:
Actually I've just applied the change again in 4a413be7d78a37cd2f7ba9bbbc10dc6f8b54c0a1, didn't wanna wait for the Travis Build and having two commits for this doesn't really provide much value :wink:

PR #161, opened 2019-03-16
Make array type return package name of their component type

codecholeric:
This PR changes the behavior of the class import with respect to array class package names.
So far an imported array class like `com.foo.SomeClass[]` always had an empty package (similar to `getPackage() == null` of the Reflection API).
However this behavior is usually not useful, because normally an user would expect a class like 

```
package com.foo;
class SomeClass {
    com.bar.OtherClass[] array;
}
```

to cause a dependency to package `com.bar`.
From now on `access.getTargetOwner().getPackageName()` will report `com.bar` instead of empty in such cases.

### DIFF
--- a/archunit-example/example-plain/src/main/java/com/tngtech/archunit/example/cycle/complexcycles/slice1/UnrelatedEnum.java
+++ b/archunit-example/example-plain/src/main/java/com/tngtech/archunit/example/cycle/complexcycles/slice1/UnrelatedEnum.java
@@ -1,0 +1,14 @@
+package com.tngtech.archunit.example.cycle.complexcycles.slice1;
+
+// This was just added for the integration tests since it creates some synthetic byte code
+@SuppressWarnings("SwitchStatementWithTooFewBranches")
+public enum UnrelatedEnum {
+    INSTANCE;
+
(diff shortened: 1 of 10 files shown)


archunit-example/example-plain/src/main/java/com/tngtech/archunit/example/cycle/complexcycles/slice1/UnrelatedEnum.java
```diff
@@ -0,0 +1,14 @@
+package com.tngtech.archunit.example.cycle.complexcycles.slice1;
+
+// This was just added for the integration tests since it creates some synthetic byte code
+@SuppressWarnings("SwitchStatementWithTooFewBranches")
+public enum UnrelatedEnum {
```
hankem: Probably indeed unrelated to the package name of array types... 😉

codecholeric: Yeah, my mind might have jumped a bit :wink: Thought before I completely forget I better add it on the fly...

PR #164, opened 2019-03-28
static/final syntax

hankem:
This PR adds
* some User Guide improvements and
* the following syntax to ArchUnit's Lang API:
｛`fields`, `methods`｝`().that().are`［`Not`］｛`Static`, `Final`｝`.should().`｛`notBe`, `be`｝｛`Static`, `Final`｝`()`


codecholeric:
Very nice, thanks a lot :smile: Also that you improved the user guide!!
I pulled the `beStatic` stuff up from `FieldsShouldInternal` and `MethodsShouldInternal` the same way you already did it for `FieldsThatInternal` and `MethodsThatInternal`, don't know if you had a specific reason to not do that in the should case, but it seems to work so far. (except that it is a little confusing to have those methods without the `@Override`, but with the comment it's fine I think)

hankem:
Consolidating the ｛ `be`, `notBe`｝｛`Static`, `Final`｝`()` functions in `AbstractMembersShouldInternal` makes a lot of sense, thanks! I just didn't do so because I somehow didn't realize that `MethodsShouldInternal` was `FieldsShouldInternal`'s nephew... ☺

PR #167, opened 2019-04-05
Add classes().that().areAnyClass(...) to syntax

mobolajiadefope-toast:
Extended the API to allow assertions to made against one or multiple types using class objects. 
E.g. The declaration
```
theClass(Foo.class).should().onlyHaveDependentClassesThat().areAnyClass(Bar.class, Baz.class).check(importedClasses);
```
should pass if only `Bar` and `Baz` and any of their anonymous inner classes depend on `Foo`.

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
@@ -17,6 +17,7 @@ package com.tngtech.archunit.core.domain;
 
 import java.lang.annotation.Annotation;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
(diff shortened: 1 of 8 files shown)


codecholeric:
Hey, thanks a lot :smiley: Looks like a reasonable addition. Can you do me a favor though and sign off your commits according to the DCO? -> https://github.com/TNG/ArchUnit/blob/master/CONTRIBUTING.md#commits
With two commits it might be easier to just add the signed-off line (the one you'd get if you use `git commit -s`) to your messages by rewording.

mobolajiadefope-toast:
Retry Deepcode

codecholeric:
One thing that really gives me a stomach ache is the fact that `areAnyClass(..)` matches inner classes. Cause that does feel specific to your use case and might surprise in other places. The question is, if what you really want is classes with their inner cases, should it be a different name, or is this use case in the end too specific for the general API.
Because for specific things it would be better to use a custom predicate and write sth. like

```
onlyHaveDependentClassesThat(areAnyOfOrInnerClassesOf(Foo.class, Bar.class))
```

mobolajiadefope-toast:
It only matches anonymous inner classes. This is mostly to catch lambdas.

cakeface:
I think that for `areAnyClass(Bar.class, Baz.class)` it is more correct and intuitive to include lambdas within the class for architectural checking. I think we should clearly document the contract. 

I'm trying to come up with a situation where we want to match the class but not match anonymous inner classes and not coming up with much.

codecholeric:
I don't doubt that in this case you probably always want to cover lambdas as well. But inner classes also covers `Foo.Bar`, what about those? Named inner classes? Possibly static named inner classes? Does `static class Foo.Bar` count as `Foo`? Or is it a different class?

mobolajiadefope-toast:
Oops. Did not notice i was matching with non-anonymous inner classes. Will fix this.

mobolajiadefope-toast:
Lost push access to toasttab fork so created https://github.com/TNG/ArchUnit/pull/173

PR #180, opened 2019-05-28
Set log level debug for debug output

Bananeweizen:
The info level logging of classes causes huge amounts of unwanted log output in Maven. Change it to the same debug log level as being used for all the other logging in this class. See https://github.com/societe-generale/arch-unit-maven-plugin/issues/10
Also change the text to American English, since that is used in other parts of the code base.

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/importer/JavaClassProcessor.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/importer/JavaClassProcessor.java
@@ -98,7 +98,7 @@ class JavaClassProcessor extends ClassVisitor {
 
     @Override
     public void visit(int version, int access, String name, String signature, String superName, String[] interfaces) {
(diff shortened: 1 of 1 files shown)


codecholeric:
Seems reasonable, originally I thought which classes were imported is a different level than gory details of method import, etc. But I've found myself to not really looking at that INFO log either, so I guess it makes sense to reduce it to DEBUG.

codecholeric:
Can you do me a favor and

* Sign off your commit according to https://github.com/TNG/ArchUnit/blob/master/CONTRIBUTING.md#commits
* Rebase your branch against upstream master? I've tried to do it, but I have no permission to force push on your fork :wink: I'd like to do a fast forward merge to keep the history straight...

codecholeric:
Since I can't push to this branch and there is no further activity, I've implemented this myself. Thanks for the initiative though!! I think this is a reasonable adjustment.

PR #192, opened 2019-07-03
upgrade Gradle to 5.5 and adjust plugin versions

jangalinski:
I noticed that archunit uses a quite old gradle version ... so I updated to latest.


codecholeric:
Cool, thanks a lot :smiley: I would have expected more problems on the upgrade to 5.x, so I'm pleasantly surprised that these are the only things needed to change!

Can you do me a favor and sign off your commit according to https://github.com/TNG/ArchUnit/blob/master/CONTRIBUTING.md#commits ?

Thanks!! (I made a little review commit about adjusting the Spotbugs source set, so either reword your commit and manually add the `signed-off-by` line or feel free to squash my commit onto yours and use `git commit --amend -s`, whatever you prefer)

codecholeric:
I've just stumbled over https://github.com/TNG/junit-dataprovider/issues/119
We might have the same problem with ArchUnit since the signing / publishing mechanism is fairly similar to that project. Have to check that out when I'm back from holiday...

aaschmid:
@codecholeric https://github.com/TNG/junit-dataprovider/issues/119 is finished if you don't need the OSGi entries in `MANIFEST.MF` which I at least could not find so far.

I can apply the changes to your build file if you want.

codecholeric:
That would be great, thanks @aaschmid !!

aaschmid:
OK, some questions @codecholeric:
* Do you still need the TNG Nexus as deployment location?
* Do you still use `install` task? What's your use case?

codecholeric:
@aaschmid the answers are:

* I don't really use the TNG Nexus anymore. Originally I used it to make sure the release looked good, but meanwhile I have enough infrastructure to be fairly certain at Maven Central as well
* Yes, I do still use the `install` task, e.g. if I want to test the performance of a feature in the context of another project. I could do that differently, I think, but the ability to install it `mavenLocal()` is quite nice for quick tests (e.g. to include the current SNAPSHOT version in `hibernate-orm` or similar for testing)

Let me know if anything is unclear. The build infrastructure has grown a little ... historically :wink: It's actually not as simple as it should be :frowning_face:

aaschmid:
* Ok, will remove the TNG nexus
* instead of `install` the new `maven-publish` plugin offers `publishToMavenLocal` which should be sufficient, see https://docs.gradle.org/current/userguide/publishing_maven.html#publishing_maven:tasks

codecholeric:
@jangalinski Hey, sorry that this dragged on so long! I guess according to @aaschmid we should be able to upgrade to the latest Gradle version 5.x without breaking the release process.

I saw that you merged `master` into this branch, we generally try to avoid that. I.e. we always rebase against `upstream/master` and then force push, thus we can always fast forward merge feature branches and keep the history linear. I rebased this branch against `master` which in turn got rid of the merge commit.

I mentioned this a while time ago (compare https://github.com/TNG/ArchUnit/pull/192#issuecomment-508815872), it would be great if you could do me a favor to get this mergeable:

According to the DCO we need contributors to sign-off their commits (compare https://github.com/TNG/ArchUnit/blob/master/CONTRIBUTING.md#commits). You can see how that looks if you check out my commits on top of your's. Can change the message like this of your old commit? One way to do that would AFAIK be to do 

```
git rebase -i $commitBeforeYourCommit
> mark your commit as 'edit'
git commit -s --amend
git rebase --continue
```

Otherwise you could also patch the message of the commit, but I think the `commit -s` version is safer in general.

Thank you!!

codecholeric:
Superseded by #306 due to inactivity

PR #197, opened 2019-07-11
adding log statements around archUnit property file 

vincent-fuchs:
adding log statements to indicate whether archUnit property file is found or not : when adding a property file but nothing happens, it's useful to easily see in the logs if the file is picked up or not. 

Having a simple log statement can also be very helpful to somebody new to a project using archUnit, but not getting the expected behavior : it will indicate clearly that a property file has been picked up

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/ArchConfiguration.java
+++ b/archunit/src/main/java/com/tngtech/archunit/ArchConfiguration.java
@@ -29,6 +29,8 @@ import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableMap;
 import com.tngtech.archunit.base.Optional;
 import com.tngtech.archunit.core.importer.resolvers.ClassResolver;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


codecholeric:
Thanks a lot! Yes, I agree that can save some time :wink: 
If I look at the code, I wonder if it wouldn't make sense to also log a warning for an occurring `IOException`? I don't think it was a good idea back then to just secretly swallow it.
In any case, could you check out https://github.com/TNG/ArchUnit/blob/master/CONTRIBUTING.md?
In particular
* please sign the commits according to the DCO (easiest with one commit: `git commit -s --amend`)
* please format the code with the provided formatter (I hope you're using either Eclipse or IntelliJ :wink:)

Thank you!!

vincent-fuchs:
sorry for the delay - I fixed my commit and force-pushed. I also added a warn log statement as you suggested.

codecholeric:
Thanks :smiley: Now that I looked at it again I thought it might be useful to see the full resolved URL of `archunit.properties`. So I've rebased against upstream master and added a little review commit to resolve the URL separately and then print that within the log statement.

Issue #113, opened 2018-10-09
Expose parameter annotations 


wolfs:
For defining assertions that method parameters have certain annotations it would be great to expose method parameter annotations via some method on `JavaMethod`.

Currently, I need to do the following to get the annotations of the first parameter:
```
method.reflect().getParameterAnnotations()[0])
```

I used this to check if Java setters have matching `@Nullable` annotations with Java getters here: https://github.com/gradle/gradle/blob/master/subprojects/architecture-test/src/test/java/org/gradle/architecture/test/KotlinCompatibilityTest.java#L254-L256

wengertj:
@wolfs I might be wrong (or missunderstand your question) but I think this is already possible.
Looking at the [Javadoc for JavaMember](https://static.javadoc.io/com.tngtech.archunit/archunit/0.9.3/com/tngtech/archunit/core/domain/JavaMember.html) there is a method `getAnnotations()`.

So you should be able to do this:
```java
// Returns com.tngtech.archunit.core.domain.JavaAnnotation
for (final JavaAnnotation annotation : method.getAnnotations()) {
     // Do stuff with:
      annotation.getType();
      annotation.getProperties()
}
```

codecholeric:
I think the problem is 

```
@Foo
void method() { ... }
```

which is already implemented, vs.

```
void method(@Foo param) { ... }
```

which is unfortunately still missing... One bad thing is that there is a `getParameters()` method that returns an iterable of `JavaClass` while we would need sth. like `CodeUnitParameter` that offers `getType()` and `getAnnotations()` (or similar). So this will probably need some migration via `@Deprecated` to get there.

wengertj:
Oh ok I see - I really did not understand the question then :)

Farbauti89:
Hi everyone,

I'm interested in solving this issue.
At the "Hackergarten Berlin" [Philip Riecks](https://github.com/rieckpil) and I already started to work on it :)

remal:
Are there any news?

steftux:
We had another crack at it, unfortunately currently stuck at 
        public AnnotationVisitor visitAnnotation(String desc, boolean visible) {
            return new AnnotationProcessor(addAnnotationTo(annotations), annotationBuilderFor(desc));
        }
from MethodProcessor class from JavaClassProcessor.java somehow...

Farbauti89:
Yeah basically we are struggling with the integration of the asm Dependency right now.
[...]

Issue #115, opened 2018-10-09
Expose type parameters of JavaClass


wolfs:
It should be possible to obtain the type parameters for a JavaClass.

When given code like this:
```
public class Foo implements List<Bar> {
}
```

I now want to obtain the type parameters of the interface. Same goes for fields, method return types and parameters, etc.

Something like this would be great:
```
JavaClass genericType = <List<Foo>>
genericType.isAssignableTo(Iterable.class) // true, I guess that works already
genericType.getTypeParameters() == [Foo]
```

slq:
This might not be as easy as it seems at the first spot.

Due to the recursive nature of generic types (a generic type can be parameterized by a generic type – consider for example `List<List<String>>`) the grammar of type signatures is quite complex. We will need to support 

1. Add parameter to `JavaType` interface and its implementation
```java
public interface JavaType {
    ...
    JavaType getParameter();
    boolean isParametrizedType();
    ...
}
```
2. Rewrite `com.tngtech.archunit.core.importer.ImportedClasses` to use `JavaType` instead of `String` as cache key in `#directlyImported`, `#additionalClasses`.
3. Write implementation of `org.objectweb.asm.signature.SignatureVisitor` that will instantiate `JavaType` implementations and add type parameters.
```java
//com.tngtech.archunit.core.importer.JavaClassProcessor
public FieldVisitor visitField(int access, String name, String desc, String signature, Object value) {
        ...
        if (signature != null) {
            XXXSignatureVisitor visitor = new XXXSignatureVisitor();
            SignatureReader reader = new SignatureReader(signature);
            reader.acceptType(visitor);
            // visitor.getParametrizedType();
       }
       ...
}
```

I will try to proceed with implementation using smaller steps:
1. Implementation for simple case `List<String>`, `Set<Order>`, etc.
2. Add support for recursive generic types `List<List<List<String>>>`, `Map<String, List<Order>>`, etc.
3. Expose generic type as dependency (same as in https://github.com/TNG/ArchUnit/issues/136)

Few additional questions arise:
1. How to proceed with type parameters: `Set<T>`, `Set<? extends T>`, `Set<? super T>`?
2. How to proceed with parametrized types like: `Set<?>`, `Set<? extends Order>`, `Set<? super Order>`?

codecholeric:
Thanks for looking into that :smiley: Nobody ever said this is easy :wink: 
[...]

PR #202, opened 2019-07-19
WIP introducing new method to receive a `CodeUnitParameterList` from a JavaCodeUnit

Farbauti89:
This PR is still WIP.
I just wanted to check if the general Idea used here to solve Issue #113 is okay.

As explanation:
The Idea is that one of the Builders in DomainBuilders.java will be extended to return a `CodeUnitParameterList` (The name is based on the new `CodeUnitParameter` suggested in the issue). This List is then used to initialize the `JavaCodeUnit`.
From there on the the flow is the same as it is right now - The `JavaCodeUnit` gets a new method `getCodeUnitParameters` and so on.

The deprecated method `JavaCodeUnit::getRawParameterTypes` will create its return value based on the new `getCodeUnitParameters` Method. This way the backwards compatibility should be 
assured.

Any thoughts about it @codecholeric ?

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/CodeUnitParameter.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/CodeUnitParameter.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2019 TNG Technology Consulting GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


codecholeric:
Thanks for looking into this :smiley: 
Yes, I think this should work. I.e. Create `CodeUnitParameter`, then fill it with the information within `com.tngtech.archunit.core.importer.JavaClassProcessor`. It might be a little tricky though in the end, because I think there is some lazy mechanism with respect to annotations in place.
One thing I noticed about your PR so far is that

* ArchUnit domain objects should not be constructible by clients (-> public constructors)
* public ArchUnit methods may never contain parameter types from Guava (Guava gets relocated during build and ArchUnit's API should not depend on Guava in any way)

codecholeric:
Are you still working on this? Can I support you in any way? I think I will also need a `CodeUnitParameter` for solving #115 eventually (because we will e.g. need `method.getParameters().get(0).getTypeParameters()`)

Farbauti89:
Hi @codecholeric ,

I don't think I'am able to finish this issue :(
As mentioned in the issue I struggled to get the needed information from the asm Lib.

Also I currently can't find the time or motivation to push trough this.
Sorry :(

codecholeric:
@Farbauti89 No problem, thanks for letting me know! I know that it's not always easy to find the extra energy to push some OSS project :wink: But then at least it's clear that this issue is open for grabs again...

PR #204, opened 2019-07-22
add [haveOnlyPrivateConstructors]

sullis:


sullis:
please review @codecholeric

codecholeric:
Thanks for adding this :smiley: 
I've made some review adjustments, in particular unified the two conditions asserting that all of certain objects (fields, constructors, ...) have a certain modifier.
Can you check if that looks good to you? The semantics shouldn't have changed, only the reporting of the details, but I think it's better if it's consistent.
If you give me your thumbs up I'll merge it.

archunit/src/main/java/com/tngtech/archunit/lang/syntax/elements/ClassesShould.java
```diff
@@ -257,6 +257,14 @@
     @PublicAPI(usage = ACCESS)
     ClassesShouldConjunction haveOnlyFinalFields();
 
+    /**
+     * Asserts that classes have only private constructors.
+     *
+     * @return A syntax element that can either be used as working rule, or to continue specifying a more complex rule
+     */
+    @PublicAPI(usage = ACCESS)
+    ClassesShouldConjunction haveOnlyPrivateConstructors();
```
ratoaq2: It would have been much better to expose `haveOnlyConstructorsWithModifiers` as well

codecholeric: We're open for PR's :wink:
Should be easy to pipe this method to a new `haveOnlyConstructorsWithModifiers(..)` method.
Other than that you can of course already test this with the current state, just directly use the predicates / conditions and adjust them...

PR #217, opened 2019-08-20
Ensure correct simple name for local classes

### LINKED ISSUE #216: Simple name of local class is wrong
Test:
```java
@Test
void simpleNameOfLocalClass() {
    class LocalClass {
    }

    JavaClass javaClass = new ClassFileImporter().importClass(LocalClass.class);
    System.out.println(javaClass.getSimpleName());
    System.out.println(javaClass.getName());

    System.out.println(LocalClass.class.getSimpleName());
    System.out.println(LocalClass.class.getName());
}
```

Output:
```
1LocalClass
org.dummy.ExampleTest$1LocalClass
LocalClass
org.dummy.ExampleTest$1LocalClass
```

I would expect that the output of `javaClass.getSimpleName()` is also `LocalClass`.

rweisleder:
Resolves #216

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/Formatters.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/Formatters.java
@@ -19,7 +19,6 @@ import java.util.ArrayList;
 import java.util.List;
 
 import com.google.common.base.Joiner;
-import com.google.common.primitives.Ints;
 import com.tngtech.archunit.PublicAPI;
 import com.tngtech.archunit.core.domain.properties.HasName;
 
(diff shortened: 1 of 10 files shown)


codecholeric:
Hey, thanks for so much support these days :smile: 
Also I'll give you a badge for being the first person since I'm coding Java, who actually uses local classes :wink: (I never really understood the need for something between a 'member' class and an anonymous class)
I've looked into it and actually found, that there is a way to extract the simple name from the byte code with ASM (even though I still consider the visitor flow somewhat weird, that `visit(..)` and `visitInnerClass(..)` are called for the same class. So I've added that on top of your PR, because I think it's more robust. There can always be classes that contain a '$' within their simple class name (even though not super common for the classes one is interested in) and the heuristic just doesn't work for those classes (and I think it's an inherent problem, there is no way to fix this, if all your input data is a FQN)
So, maybe you can check out my commits, and if it seems reasonable to you, I'd press the button :wink:

rweisleder:
@codecholeric I looked into your commits. The new solution to identify the simple name looks more robust now. 👍 
Feel free to merge this into master. :-)

codecholeric:
Cool, I'll merge it then :smiley: And thanks again for fixing those things, I really appreciate it!

PR #220, opened 2019-08-24
Read "static" modifier of classes

### LINKED ISSUE #219: Modifier "static" of nested classes is not recognized
Currently, interfaces that are declared within another class and don't have the static keyword, are not considered as static. But those interfaces are implicitly static.

See [JLS, 8.5.1](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.5.1):
> A member interface is implicitly static (§9.1.1). It is permitted for the declaration of a member interface to redundantly specify the static modifier.

Test:
```java
class ExampleTest {

    @Test
    void modifierOfInterface() {
        JavaClass javaClass = new ClassFileImporter().importClass(NestedInterface.class);
        System.out.println(Predicates.modifier(STATIC).apply(javaClass)); // ArchUnit

        System.out.println(Modifier.isStatic(NestedInterface.class.getModifiers())); // JDK
    }

    interface NestedInterface {}
}
```

Output:
```
false
true
```

rweisleder:
Resolves #219

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaModifier.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaModifier.java
@@ -34,7 +34,7 @@ public enum JavaModifier {
     @PublicAPI(usage = ACCESS)
     PRIVATE(EnumSet.allOf(ApplicableType.class), Opcodes.ACC_PRIVATE),
     @PublicAPI(usage = ACCESS)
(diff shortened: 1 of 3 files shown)


codecholeric:
Ah, cool, thanks a lot :smiley: You're right of course, we need to read the static modifier from classes as well, guess I kind of had some tunnel vision when I originally implemented this :wink:

Issue #207, opened 2019-07-28
Extend Classes Syntax for anonymous / inner / outer classes


codecholeric:
It would be neat to have `classes().that().`

* `areAnonymous()`
* `areInnerClasses()` (includes anonymous classes, do we need `areNamedInnerClasses()`?)
* `areOuterClasses()` (i.e. no enclosing class / toplevel, I'm open for better naming :wink:)

Same holds for `classes().should().be...`

codecholeric:
Overlaps in parts with #172

rweisleder:
The JLS defines four different kinds of classes, see [chapter 8](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html).

With this reference I suggest to add the methods `classes().that().`
* `areTopLevelClasses()`
* `areNestedClasses()`
* `areLocalClasses()`
* `areAnonymousClasses()`

And of course the same for `classes().should().be...`

To be consistent with existing methods we should also add the negated methods like `areNotTopLevelClasses()` and `notBeTopLevelClasses()`.

codecholeric:
We should also make the determination of the nested class type more resilient. Currently `isAnonymous()` is implemented by checking the simple name for being empty. The class file import should be able though, to derive a type of the nested class (static nested, inner, local, anonymous) and pass that on to the `JavaClass` instead of applying some heuristics later (since for a local class this would be something shady like `javaClass.getName().matches(".*\\$\\d+\\w+")` or similar :fearful:)

rweisleder:
I started working on this issue.

PR #224, opened 2019-08-31
Clarify difference between nested and inner classes

rweisleder:
This is something we should clarify before implementing #207, so that we have consistent naming.

> "A nested class is any class whose declaration occurs within the body of another class or interface."

> "An inner class is a nested class that is not explicitly or implicitly declared static."

See [JLS, chapter 8](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html).

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
@@ -186,9 +186,22 @@ public class JavaClass implements HasName.AndFullName, HasAnnotations, HasModifi
         return javaType.isArray();
     }
 
+    /**
+     * @return Returns true if this class is declared within another class.
+     *         Returns false for top-level classes.
+     */
+    @PublicAPI(usage = ACCESS)
+    public boolean isNestedClass() {
+        return enclosingClass.isPresent();
+    }
+
+    /**
+     * @return Returns true if this class is a non-static nested class.
+     *         Returns false otherwise.
+     */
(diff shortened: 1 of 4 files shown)


codecholeric:
Thanks for the PR :smiley: I completely agree, that the ArchUnit terminology was a little sloppy there :disappointed: 
Unfortunately this is now the worst sort of breaking an API, it compiles fine, but the API semantically behaves differently :frowning_face: I've pondered how to mitigate this for a while, but I didn't come up with a good solution. I thought about printing a warning to `System.err` for one release, but that would be really annoying, if you really want to use `isInnerClass()`. Or creating a new method for now and deprecating `isInnerClass()` to give a clear signal, but `isInnerClass()` simply is the best name. So I guess in the end I'll stick to simply printing a dominant warning on top of the release notes for the next version, but it's pretty clear that there will be people not reading this and being surprised that their current rules behave differently (and in the worst case catching less than before).
BTW. I've moved the tests to `ClassFileImporterTest`, even though that test is pretty fat. But originally the idea was to have a black-box view on the import from there. Meanwhile the test scope got a little fuzzy in some places, because when the whole import matured I started to use it for simple unit tests as well, because the test setup was simply better and more realistic. I would see the point to put these tests in `JavaClassTest`, because `isInnerClass()` simply derives from two other attributes of `JavaClass`, but `ClassFileImporterTest` has a test for every other type of nested class, so I thought it's best to add it there.
Anyway, thanks again! I'm gonna merge it now...

PR #225, opened 2019-09-04
Add classes().that().areEnums() and .should().beEnums()

### LINKED ISSUE #172: orShould().beAnInterface().orShould().beAnEnum().orShould().beAnInnerClass()
First of all: ArchUnit is great! I've been looking for something like this for many years and it turns out ArchUnit is much better than everything I've used before and even better than what I had wished for. Keep going!

I'd like to have a rule that checks that the data package only consists of certain types of classes. That's pretty straightforward to do:

```java
    @ArchTest
    public static final ArchRule shouldOnlyHaveCertainClassesInDataPackage = 
            classes().that().resideInAPackage("..data..")
            .should().haveSimpleNameEndingWith("Repository")
            .orShould().beAnnotatedWith(Entity.class)
            .orShould().notBePublic();
```

Then I realized that there are some checks that are not supported out of the box. I'd like to have methods `beAnInterface()`, `beAnEnum()`, and `beAnInnerClass()`.

Fortunately it's easy to specify these conditions yourself:

```java
private static ArchCondition<JavaClass> beAnEnum() {
        return new ArchCondition<JavaClass>("be an Enum") {
            @Override
            public void check(JavaClass javaClass, ConditionEvents events) {
                String message = javaClass + (javaClass.isEnum() ? "is an Enum" : "is not an Enum");
                events.add(new SimpleConditionEvent(javaClass, javaClass.isEnum(), message));
            }
        };
    }
```

Nevertheless I feel like this should be part of the standard API.

rweisleder:
Resolves #172

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
@@ -1219,6 +1219,14 @@ public class JavaClass implements HasName.AndFullName, HasAnnotations, HasModifi
         };
 
         @PublicAPI(usage = ACCESS)
+        public static final DescribedPredicate<JavaClass> ENUMS = new DescribedPredicate<JavaClass>("enums") {
+            @Override
+            public boolean apply(JavaClass input) {
+                return input.isEnum();
+            }
+        };
+
+        @PublicAPI(usage = ACCESS)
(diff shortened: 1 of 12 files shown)


codecholeric:
Cool, thanks for being so diligent :smile: 
I've added a test for `ClassesShould`, other than that lgtm!!
I guess this resolves #172, since the only things left over are covered by #207 anyway.

Issue #228, opened 2019-09-05
Layered architecture by names


blackfoxoh:
Hi
I really like the idea of the layerd architecture check but sadly it seems to depend on the layers beeing identifyed by package.
We use packages by feature and have xxxFacade and xxxService in the same package per feature. Would it be possible to add something like `layer('Facade').definedByClassSimpleNameContaining('Facade')` ?
Or is there something I missed to achieve this with current version of ArchUnit?

codecholeric:
Honestly I think that if you care about clear layers within your features, you should introduce a subpackage structure modelling that :wink: (i.e. `com.myapp.order.facade` and `com.myapp.order.service`). Typically it's not just those toplevel classes called `*Facade` or `*Service`, but any parameters or utilities that those classes need, where the borders start to get fuzzy.
Nevertheless, I think it would be a good idea to make `LayeredArchitecture` more powerful. I think this could simply be accomplished by offering a more generic

```
// predicate.apply(..) == true <==> JavaClass in layer
layer("foo").definedBy(describedPredicate)
```

Then the existing `definedBy(packageIdentifier)` would simply be a special case of the generic API and everything should be consistent. And it would naturally be powerful enough to cover your use case.
However, I won't have time for a while to implement this. I think this would be a good start to contribute though, because it's a very limited scope and easy to describe. So I'll keep it open and see if anyone wants to help here :smile_cat:

PR #238, opened 2019-09-20
Add the possibility to define a layer using a DescribedPredicate - Issue 228

mikomatic:


mikomatic:
Sorry for all the force pushing :/  

This PR should not move anymore and is ready for review.

codecholeric:
Hey, thanks a lot :smile: Sorry that it took me so long to look into it!!
I took the liberty to adjust your commit messages a little, because I think commit messages should be self-contained and not depend on e.g. GitHub. If you just write "implemented issue #228", then the info what this commit was about is gone as soon as the issue on GitHub is gone. Also if one quickly looks into the history later, it's tedious to look for the number on GitHub, just to find out what the goal of the commit was.
I refactored the code a little, because I had the feeling that `LayeredArchitecture` did a lot of low level things with the layer definitions and the wrapped predicate.
Anyway, I'm gonna merge it now so it can go right into the upcoming release, thanks again!!

Issue #212, opened 2019-08-10
Volatile should also be applicable to Java Methods


ratoaq2:
I'm not sure if this is related to #87 

I noticed that several methods that are actually in an abstract super class was appearing in one of my tests as they were in the subclass. After some investigation I noticed that all these methods are marked as "volatile" by java reflect (`Modifier.isVolatile()`) but the JavaModifier from ArchUnit wasn't showing them as "volatile". It seems that ArchUnit only allow volatile at the field level:
`VOLATILE(EnumSet.of(ApplicableType.FIELD), Opcodes.ACC_VOLATILE),`

The only way to filter them out from my tests was to create a predicate that does a reflect and use standard java reflect API to check the modifiers (which cause the test to be much slower).

hankem:
Not allowing `volatile` methods is consistent with [the Java Language Specification](https://docs.oracle.com/javase/specs/jls/se12/html/jls-8.html#jls-MethodModifier). 😉

[`Modifier.isVolatile(int)`](https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/lang/reflect/Modifier.html#isVolatile(int)) and [`Modifier.toString(int)`](https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/lang/reflect/Modifier.html#toString(int)) cannot distinguish [`Modifier.VOLATILE`](https://docs.oracle.com/en/java/javase/12/docs/api/constant-values.html#java.lang.reflect.Modifier.VOLATILE) for fields from [`Modifier.BRIDGE`](http://hg.openjdk.java.net/jdk/jdk12/file/06222165c35f/src/java.base/share/classes/java/lang/reflect/Modifier.java#l337) (internal) for methods – cf., e.g., [Stack Overflow: 'volatile' in method signature?](https://stackoverflow.com/q/10370033)

ratoaq2:
Is there a way to filter these them out in ArchUnit.
I'm debugging now and I can see that I have 2 methods that looks the same. Same class, same name, same parameters, etc... but one with my annotation and the other without my annotation

ratoaq2:
I also noticed that sometimes the reflect for the 2 JavaMethod instances returns the method that has the volatile modifier and sometimes not. I get different results in different executions.

codecholeric:
As @hankem pointed out, by the JLS there is no `VOLATILE` for methods. If you look at the definition of `volatile` this also makes sense in my opinion :wink: 
[...]

Issue #87, opened 2018-07-09
Synthetic fields considered by matching rules


bedla:
Hi, I have found that synthetic fields are matched against specified rules. I think that this kind of fields (also methods?) should be skipped.
I can create PR, but first I want to discuss it...
Ivos

```java
    @Test
    public void xxx() {
        noClasses()
                .should()
                .accessClassesThat()
                .resideOutsideOfPackages(
                        "com.tngtech.archunit.lang.syntax.elements..",
                        "java..")
                .check(importClasses(Foo.class));
    }

    public enum Foo {
        A, B, C;

        static void doIt() {
            for (Foo value : values()) {
                System.out.println(value);
            }
        }
    }
```

Violation output is
```
java.lang.AssertionError: Architecture Violation [Priority: MEDIUM] - Rule 'no classes should access classes that reside outside of packages ['com.tngtech.archunit.lang.syntax.elements..', 'java..']' was violated (1 times):
Method <com.tngtech.archunit.lang.syntax.elements.ClassesShouldTest$Foo.values()> calls method <[Lcom.tngtech.archunit.lang.syntax.elements.ClassesShouldTest$Foo;.clone()> in (ClassesShouldTest.java:1175)
```

codecholeric:
This already came up recently: https://github.com/TNG/ArchUnit/issues/81

I do agree, that there is little value in checking synthetic methods for architecture tests, I'm just not completely sure, at which level this should be approached.

ArchUnit tries to mimic the Reflection API at its core layer, and the Reflection API allows access to synthetic method, thus up to now synthetic methods are imported.

I see two approaches:

1) Don't import any synthetic methods / fields at all
2) Filter synthetic accesses at the rule API level

With 1. the advantage is, that it is a very limited scope that needs to be adjusted, just at the import level. On the other hand, information will be missing if this should be relevant on some level for tests. I don't like that implication too much, but I have to admit that I can't really think of an use case for having access to synthetic members at the moment.
With 2. the advantage would be, that the core API stays as powerful as it has been up to now, but on the downside it's various places that need to be adjusted, and every user has to consider synthetic members in custom predicates.

An alternative might be, to filter on the import level, but make the behavior configurable, i.e. extend `ArchConfiguration` to allow specifying this behavior, with the default skipping any synthetic members from the import.

Any opinions or further ideas?

bedla:
[...]

PR #243, opened 2019-10-04
Handling of SYNTHETIC and BRIDGE modifiers added

alexfedorenchik:
Issue: #212
Issue: #87

Signed-off-by: Alexander Fedorenchik <alexander.fedorenchik@gmail.com>


codecholeric:
Thanks a lot :smiley: 
I've simplified the test a little bit and added some comments to the example class as to why synthetic members are actually created (just for the next time someone needs to understand the test :wink:)

Issue #187, opened 2019-06-25
Offer a way to retrieve the component type of a `JavaClass` that is an array


codecholeric:
As a follow up of #114 we should offer a way to get the component type of an array (compare the Java Reflection API -> `arrayClass.getComponentType()`)
For example a `JavaClass` representing an array `com.some.Foo[]` should return `com.some.Foo` as the component type. Likewise `com.some.Foo[][]` should return `com.some.Foo[]` as its component type.
On the contrary to the Reflection API and to be consistent with ArchUnit's API, we should call this method

```
Optional<JavaClass> tryGetComponentType()
```

and return `Optional.absent()` if the respective `JavaClass` is no array type. Additionally we could offer `JavaClass getComponentType()` throwing an exception if the `JavaClass` is no array type for cases where the nature of the `JavaClass` is already known.

**Note**: The `JavaType` within `JavaClass` already offers a method

```
Optional<JavaType> tryGetComponentType()
```

So to implement the feature during class creation the respective `JavaType.getName()` must be resolved against the matching `JavaClass`.

PR #244, opened 2019-10-05
Offer a way to retrieve the component type of a `JavaClass` that is an array

alexfedorenchik:
Issue: #187

Signed-off-by: Alexander Fedorenchik <alexander.fedorenchik@gmail.com>

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/ImportContext.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/ImportContext.java
@@ -57,5 +57,5 @@ public interface ImportContext {
 
     Set<ThrowsDeclaration<JavaConstructor>> getConstructorThrowsDeclarationsOfType(JavaClass javaClass);
 
-    Optional<JavaClass> resolveComponentType(JavaType type);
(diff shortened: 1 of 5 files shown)


alexfedorenchik:
Not sure that I find the best to propagate component type from JavaType to JavaClass,
so any feedback will be appreciated.

Also I suppose component type has to be considered in getDirectDependenciesFromSelf(), but may be it worth to do it with a separate PR.

And thanks a lot for the library.

codecholeric:
Hey, thanks a lot for your contribution!! Looks good to me :smile: 
I've only found one corner case, namely if the component type is not referenced in another place, then for multi-dimensional arrays the component type chain might be broken.
This is, because if the type is referenced no earlier than that point, `resolve` will create a stub (for sure, because we never have byte code for an array type) and that stub is not going to be processed any further. Thus I've added some commit to ensure the whole chain is present, i.e. just walk up the component type chain and ensure every successive component type is present.

And yes, component types should be added to the dependencies of a class, but also yes, that's outside of this issue. I'll create a new one for that!

So thanks again :smiley:

PR #245, opened 2019-10-06
Extend syntax to query different kinds of classes

### LINKED ISSUE #207: Extend Classes Syntax for anonymous / inner / outer classes

rweisleder:
This PR adds the following methods to JavaClass
* isTopLevelClass
* isMemberClass
* isAnonymousClass (as replacement for isAnonymous)
* isLocalClass

This PR also adds classes().that(). and classes().should().
* [are|be]TopLevelClasses
* [are|be]NestedClasses
* [are|be]MemberClasses
* [are|be]InnerClasses
* [are|be]AnonymousClasses
* [are|be]LocalClasses

and their corresponding negations.

The naming is based on the [Java Language Specification, chapter 8](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html).

Resolves #207

### DIFF
--- a/archunit-integration-test/src/test/java/com/tngtech/archunit/PublicAPIRules.java
+++ b/archunit-integration-test/src/test/java/com/tngtech/archunit/PublicAPIRules.java
@@ -29,6 +29,7 @@ import static com.tngtech.archunit.PublicAPI.Usage.INHERITANCE;
 import static com.tngtech.archunit.base.DescribedPredicate.anyElementThat;
 import static com.tngtech.archunit.base.DescribedPredicate.doNot;
 import static com.tngtech.archunit.base.DescribedPredicate.not;
+import static com.tngtech.archunit.core.domain.JavaClass.Predicates.ANONYMOUS_CLASSES;
 import static com.tngtech.archunit.core.domain.JavaClass.Predicates.assignableTo;
 import static com.tngtech.archunit.core.domain.JavaClass.Predicates.equivalentTo;
 import static com.tngtech.archunit.core.domain.JavaClass.Predicates.resideInAPackage;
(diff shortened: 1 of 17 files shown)


codecholeric:
Cool, thank you so much!! :smile: I'm sorry it took me so long to get back to this, my queue is a little overloaded at the moment :frowning_face: Since this PR will go into my "complex review queue", thus it will also take me a little longer to completely get through with this, since on top of that queue is now #171, which is super useful, but open for a ridiculously long time... I hope I'll be through with it soon. I'm planning on finally releasing 1.0, once #171 is merged, and I'll include this one, too.
If you have any other remarks about terminology, the switch to 1.0 would be the perfect time for breaking changes :wink: 
Even though the biggest break should probably have been #224...

codecholeric:
Now that #171 is merged, I finally got around to it :smile: Thank you so much for your support!
Really nice and clean :+1:
I added some more Javadoc, because I think there are many people out there that would not spontaneously be able to define a "nested class" vs a "member class", so I added more explanation and examples (and a link to the JLS, even though that's a little brittle because there are so many different versions).
Other than that it was really nice on my eyes to review :wink:
Since you continuously contribute so much to ArchUnit, do you want to add yourself to the list of developers? (I should probably ask / mention this more often to people that contribute, but somehow always forgot :frowning: This way I'm still alone in that section, even though there are quite some other people actively supporting ArchUnit)
If yes, you can create a PR and add yourself to the section in `build-steps/release/publish.gradle`, you'll then end up in the published POMs of the released artifacts (other than that you have hopefully recognized that I attribute every contribution in the release notes anyway :wink:)
If no, that's of course also fine, I just think it's nice to see the people continuously contributing to ArchUnit in that way, instead of the artifacts looking like a one man show when in fact I often just gladly merge in supportive PRs from the community :wink:

rweisleder:
Thanks for adding the Javadoc! I wasn't sure how much I should write or if the description is enough.

codecholeric:
Yeah, but in general I cannot remember when I though "that Javadoc is really too extensive" :wink: Except if it's obsolete duplication like `sets the foo property`. On the other hand I've often thought "`processes the result`, thanks for that info" :wink: Don't think it hurts to have some more docs, in the end it might save people from having to search online for the expressions...

PR #246, opened 2019-10-07
Overload JavaClass.getConstructor() and JavaClass.getMethod()

### LINKED ISSUE #236: JavaClass should offer getters for methods and constructors independent of the classpath
At the moment there are various options

```
javaClass.getCodeUnitWithParameterTypes(Class<?>...)
javaClass.getCodeUnitWithParameterTypeNames(String...)
```

but only

```
javaClass.getMethod(Class<?>...)
javaClass.getConstructor(Class<?>...)
```

It should be possible to retrieve methods and constructors, even if the parameter types are not in the classpath.

rweisleder:
This commit adds the methods
* JavaClass.getConstructor(String... parameters)
* JavaClass.getMethod(String name, String... parameters)
* JavaClass.tryGetMethod(String name, String... parameters)

to retrieve constructors/methods even if a parameter type is not present on the classpath.

The methods
* JavaClass.getConstructor()
* JavaClass.getMethod(String name)
* JavaClass.tryGetMethod(String name)

to retrieve constructors/methods without parameters where added to avoid ambiguous method calls.

Resolves #236

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
@@ -502,13 +502,33 @@ public class JavaClass implements HasName.AndFullName, HasAnnotations, HasModifi
     }
 
     @PublicAPI(usage = ACCESS)
+    public JavaMethod getMethod(String name) {
+        return findMatchingCodeUnit(methods, name, Collections.<String>emptyList());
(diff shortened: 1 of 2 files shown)


codecholeric:
Nice, thanks a lot for your support!! :smiley: I extracted some local variables in the test, since the clutter I originally added got even worse to read now :wink:

Issue #215, opened 2019-08-19
Method reference is not considered as a dependency


dpfg:
Lets say there are two packages:
- org.company.domain
- org.company.integration

where one of classes in integration package has the following code:
```java
Runnable a = DomainClass::methodName; // or DomainClass::new
```

and the following rule
```java
@ArchTest
  public ArchRule integration_does_not_access_domain =
      noClasses()
          .that()
          .resideInAnyPackage("org.company.integration..")
          .should()
          .accessClassesThat()
          .resideInAnyPackage("org.company.domain..");
```

Check passes successfully, while when used with regular construction of DomainClass inside IntegrationClass check fails. 

Seems like the root cause is the same as in #131. Dependency has been moved to the constant pool at compile time. In this case it gets replaced by invoke dynamic that refers to the constant pool. And this is a huge difference as now we are talking about legitimate not-dead code that violates constraints but not catched.

ASM [exposes](https://static.javadoc.io/org.ow2.asm/asm/5.0.4/org/objectweb/asm/ClassReader.html#readClass-int-char:A-) API to access constant pools so theoretically can be used to detect all references to classes. 

Does it make sense to mention this limitations/corner cases in the documentation? Seems like problem doesn't get solution in near future.

codecholeric:
Yeah, you're right, it's a shortcoming at the moment. ArchUnit uses the visitor-API though, but I still think this could be enhanced. I've looked into it, and I think even the case from #131 might be solvable with the current ASM version. For your case, I think the information could be retrieved by checking

```
org.objectweb.asm.MethodVisitor.visitInvokeDynamicInsn
```

While for #131, it might be possible to use

```
org.objectweb.asm.MethodVisitor.visitLdcInsn
```

At least with a quick check the first one had a reference to the target class of the method handle within the arguments, the second one was called specifically with the type of the target class. That would actually be cool to add those features somehow :smile: 

The info could easity be retrieved by extending `com.tngtech.archunit.core.importer.JavaClassProcessor`.

I guess a `JavaCodeUnit` could have `JavaConstantAccess` or something similar (like `JavaFieldAccess`), then `JavaConstantAccess` could be integrated into a `JavaClass`'s `Dependencies`.

Similar a `JavaCodeUnit` could also have a `JavaDynamicInvocation` or something like that, to solve your actual problem.

This should be feasible, even though it's not trivial. But at least it should be possible to integrate this into the current ArchUnit code base without having to change much existing code.

[...]

PR #249, opened 2019-10-16
Added override method visitInvokeDynamicInsn to handle lambda expression method references

Tegmango:
visitInvokeDynamicInsn(String name, String descriptor, Handle bootstrapMethodHandle,Object... bootstrapMethodArguments) in
package com.tngtech.archunit.core.importer in order to make ArchUnit recognize lambda expression references to methods (mentioned in issue TNG#215).


Tegmango:
I ran into the issue that ever since adding tests (which run successful and as expected in Eclipse), building the project using gradle fails at

`> Task :archunit:compileJava FAILED
C:\dev\ArchUnit_fork\archunit\src\main\java\com\tngtech\archunit\core\domain\AccessTarget.java:362: error: incompatible types: ImmutableList<ThrowsClause<CAP#1>> cannot be converted to List<ThrowsClause<JavaCodeUnit>>
                    .toList();
                           ^
  where CAP#1 is a fresh type-variable:
    CAP#1 extends JavaCodeUnit from capture of ? extends JavaCodeUnit
`

This is especially confusing since I haven't touched this file at all. Since the first commit (42479fa950b718b90d07fbc3df07cf7a9d6e93c3) built correctly, I can only imagine that adding tests impacts the build process. However, simply deleting these tests again still lets my gradlew build  fail, so maybe the issue is in some gradle configurations? I'm kind of at a loss here.

Tegmango:
Update: I was able to resolve the aforementioned issue by moving the tests I've written to `archunit/src/jdktest9` and adding 

`sourceCompatibility = JavaVersion.VERSION_1_7
targetCompatibility = JavaVersion.VERSION_1_7` 

to `archunit/build.gradle`. I found out the error I got before only happened when trying to compile the classes in the `archunit` project with a JDK version higher than 7. Since the tests I've added include lambda expressions, they needed to be compiled with JDK >=8, so I moved them to the `jdk9test` folder.

archunit/build.gradle
```diff
@@ -2,6 +2,9 @@ apply plugin: 'idea'
 
 ext.moduleName = 'com.tngtech.archunit'
 
+sourceCompatibility = JavaVersion.VERSION_1_7
```
codecholeric: why is this necessary, by `archunit/build.gradle > subprojects {...}` this should be the default for all modules that do not explicitly override it, isn't it?

archunit/src/jdk9test/java/com/tngtech/archunit/core/importer/LambdaExampleTest.java
```diff
@@ -0,0 +1,39 @@
+package com.tngtech.archunit.core.importer;
+
+import java.util.function.Supplier;
+
+
+public class LambdaExampleTest {
```
codecholeric: Did you want to commit this? It looks like it's just a little class to experiment? Since I don't see any assertions that would actually "test" anything?

archunit/build.gradle
codecholeric: Maybe your problem would also be gone without this now that you added the tests to `jdk9test`?

archunit/src/jdk9test/java/com/tngtech/archunit/core/importer/TestLambdaMethodCalls.java
```diff
@@ -0,0 +1,67 @@
...
+import com.tngtech.archunit.core.domain.JavaMethod;
+import com.tngtech.archunit.core.importer.ClassFileImporter;
+import com.tngtech.archunit.lang.ArchCondition;
+import com.tngtech.archunit.lang.ArchRule;
+import com.tngtech.archunit.lang.ConditionEvents;
+import com.tngtech.archunit.lang.SimpleConditionEvent;
+import com.tngtech.archunit.lang.syntax.ArchRuleDefinition;
+
+
+public class TestLambdaMethodCalls {
```
codecholeric: Same here, this test would simply fail and not assert anything, right?

archunit/src/jdk9test/java/com/tngtech/archunit/core/importer/TestLambdaMethodCalls.java
```diff
@@ -0,0 +1,67 @@
...
+	@Test
+	public void testName() throws Exception {
+		JavaClasses importedClasses = new ClassFileImporter()
+				.withImportOption(location -> location.contains("LambdaExampleTest"))
+				.importPackages("com.tngtech.archunit.core.importer");
+
```
codecholeric: Also unnecessary line breaks :wink:

archunit/src/jdk9test/java/com/tngtech/archunit/core/importer/TestLambdaMethodCalls.java
```diff
@@ -0,0 +1,67 @@
...
+import com.tngtech.archunit.lang.ArchCondition;
+import com.tngtech.archunit.lang.ArchRule;
+import com.tngtech.archunit.lang.ConditionEvents;
+import com.tngtech.archunit.lang.SimpleConditionEvent;
+import com.tngtech.archunit.lang.syntax.ArchRuleDefinition;
+
+
+public class TestLambdaMethodCalls {
+	@Test
+	public void testName() throws Exception {
```
codecholeric: And I wouldn't know exactly what this test wants to tell me

Tegmango:
I've cleaned the files and unnecessary parts up a bit for more clarification.

The main reason for this PR was that we wanted to include lambda method calls in our ArchUnit tests (e.g. to test whether a method is ever called by any other method), which ArchUnit is not able to do without the Override of visitInvokeDynamicInsn. 

The test was only added as a proof that it works - without the Override, the rule would not recognise that "getAge()" and "getName()" in the Person class are referenced by a lambda in the useMethodByLambda-Method and throw a rule violation since it assumes that these methods are not accessed. 
With the Override, the calls are recognised and the test succeeds.

archunit/src/jdk9test/java/com/tngtech/archunit/core/importer/TestLambdaMethodCalls.java
```diff
@@ -0,0 +1,89 @@
...
+import com.tngtech.archunit.lang.ArchRule;
+import com.tngtech.archunit.lang.ConditionEvents;
+import com.tngtech.archunit.lang.SimpleConditionEvent;
+import com.tngtech.archunit.lang.syntax.ArchRuleDefinition;
+
+public class TestLambdaMethodCalls {
+  @Test
+  public void testName() throws Exception {
+    JavaClasses importedClasses = new ClassFileImporter()
+        .withImportOption(location -> location.contains("TestLambdaMethodCalls.DummyClassWithLambdas"))
```
hankem: The updated test does actually not import `com.tngtech.archunit.core.importer.TestLambdaMethodCalls$DummyClassWithLambdas` and
`com.tngtech.archunit.core.importer.TestLambdaMethodCalls$DummyClassWithLambdas$Person`.

If it did, it would still fail because `useMethodsByLambda()` is not called – unlike in your earlier implementation where the methods mutually called each other.

archunit/src/jdk9test/java/com/tngtech/archunit/core/importer/TestLambdaMethodCalls.java
```diff
@@ -0,0 +1,89 @@
...
+    rule.check(importedClasses);
+  }
+
+  ArchCondition<JavaClass> haveOnlyNecessaryMethods = new ArchCondition<JavaClass>(
+      "Classes must only have methods that are called at least once") {
+    @Override
+    public void check(final JavaClass item, final ConditionEvents events) {
+      List<JavaMethod> collect = item.getMethods().stream().collect(Collectors.toList());
+
+      for (JavaMethod method : collect) {
```
hankem: You could directly use `item.getMethods()` and remove `collect`.

hankem:
The [Checks](https://github.com/TNG/ArchUnit/pull/249/checks) indicate that the commits have not been signed-off according to the [DCO](https://github.com/TNG/ArchUnit/blob/master/DCO), cf. [CONTRIBUTING.md](https://github.com/TNG/ArchUnit/blob/master/CONTRIBUTING.md).

Maybe you could rebase your PR and squash all commits in one?
```
git rebase -i origin/master
# replace all "pick" commands (except for the first one) with "fixup"
git commit --amend -s
```
In addition, your PR introduces plenty of empty `.gitignore` files, which can probably be removed?
This would be:
```
git rm archunit-example/example-junit4/bin/.gitignore archunit-example/example-junit5/bin/.gitignore archunit-example/example-plain/bin/.gitignore archunit-integration-test/bin/.gitignore archunit-junit/.gitignore archunit-junit/junit4/bin/.gitignore archunit-junit/junit5/api/bin/.gitignore archunit-junit/junit5/engine-api/bin/.gitignore archunit-junit/junit5/engine/bin/.gitignore archunit/bin/jdk9main/.gitignore archunit/bin/jdk9test/.gitignore archunit/bin/main/.gitignore archunit/bin/test/.gitignore
git commit -a --amend
```

At the end, you'll need to `git push --force-with-lease`.

archunit/src/jdk9test/java/com/tngtech/archunit/core/importer/TestLambdaMethodCalls.java
hankem: To extract the essence of this test more clearly, we could even simplify your example to `ArchRuleDefinition.methods().should(beCalled)` using
```java
    ArchCondition<JavaMethod> beCalled = new ArchCondition<JavaMethod>("be called") {
        @Override
        public void check(JavaMethod method, ConditionEvents events) {
            boolean satisfied = !method.getCallsOfSelf().isEmpty();
            String message = String.format("%s %s called", method.getDescription(), satisfied ? "is" : "is not");
            events.add(new SimpleConditionEvent(method, satisfied, message));
        }
    };
```

archunit/src/jdk9test/java/com/tngtech/archunit/core/importer/TestLambdaMethodCalls.java
hankem: But you can simply use
```java
JavaClasses importedClasses = new ClassFileImporter().importClasses(DummyClassWithLambdas.class);
```
when you can boil the essence of this test down to a single example class, e.g.
```java
    static class DummyClassWithLambdas {
        void methodCallingLambdas() {
            IntStream.range(0, 1).forEach(this::methodOnlyCalledViaMethodReference);
            IntStream.range(0, 1).forEach(i -> this.methodOnlyCalledViaLambda(i));
        }

        void methodOnlyCalledViaMethodReference(int i) {
        }

        void methodOnlyCalledViaLambda(int i) {
            methodCallingLambdas();
        }
    }
```

archunit/src/jdk9test/java/com/tngtech/archunit/core/importer/TestLambdaMethodCalls.java
hankem: Your example is a nice integration test, but at the end, we probably want to have a unit test (similar to `ClassFileImporterTest`'s `methods_know_callers`) that specifically tests the calls via method references or lambdas.

archunit/src/main/java/com/tngtech/archunit/core/importer/JavaClassProcessor.java
```diff
@@ -327,6 +328,19 @@ public void visitFieldInsn(int opcode, String owner, String name, String desc) {
...
+        
+        @Override
+		public void visitInvokeDynamicInsn(String name, String descriptor, Handle bootstrapMethodHandle,
+				Object... bootstrapMethodArguments) {
+			for (Object o : bootstrapMethodArguments) {
+				if (o instanceof Handle) {
+					Handle currentHandle = (Handle) o;
+					accessHandler.handleMethodInstruction(currentHandle.getOwner(), currentHandle.getName(),
+							currentHandle.getDesc());
+					break;
```
hankem: Can we break the loop because `bootstrapMethodArguments` cannot contain more than one `Handle`?
(I must admit that I'm not familiar with `visitInvokeDynamicInsn` at all, I've so far only seen [ASM's javadoc](https://asm.ow2.io/javadoc/org/objectweb/asm/MethodVisitor.html#visitInvokeDynamicInsn-java.lang.String-java.lang.String-org.objectweb.asm.Handle-java.lang.Object...-)...)

Tegmango:
Thanks for your hints and comments - I‘ll look more into them tomorrow/over the weekend

hankem:
You're welcome! I'm sorry if my rebase suggestion caused troubles... 😉
(It takes some time to get used to it, but IMO it's absolutely worth it.)

codecholeric:
@Tegmango just so I understand the intent of this PR, is this WIP to finally be merged to solve #215 ? Or is this mainly a POC that overriding this method in the visitor provides enough details to solve #215 ? If it is the second, do you still plan on adding anything? In other words, where should this PR go to, what is your goal?

codecholeric:
Superseded by #649

PR #259, opened 2019-11-05
Fix grammar in Javadoc

Bananeweizen:
Replace article "an" by "a" in cases where it is not appropriate.

Signed-off-by: Michael Keppler <Michael.Keppler@gmx.de>


codecholeric:
Thanks :smiley: And sorry for the delay!

PR #269, opened 2019-11-25
make default ViolationLinerMatcher ignore numbers after a "$" (#248)

### LINKED ISSUE #248: Freezing Architecture Tests doesn't work well with Lambdas
I have a problem with a frozen rule when building on my CI server.

I'm using the standard rule `NO_CLASSES_SHOULD_ACCESS_STANDARD_STREAMS` - but due to lots of legacy code, I've had to freeze the rule as a starting point:
```
@ArchTest
    public final ArchRule no_new_access_to_standard_streams = freeze(NO_CLASSES_SHOULD_ACCESS_STANDARD_STREAMS);
```

In the saved rules file, I have something like this (class names redacted):
```
Method <my.company.package.MyClass.lambda$myFunction$2()> calls method <java.io.IOException.printStackTrace()> in (MyClass.java:123)
```

However, when this builds on Jenkins, I'm getting a test failure like this:
```
Method <my.company.package.MyClass.lambda$myFunction$114()> calls method <java.io.IOException.printStackTrace()> in (MyClass.java:123)
```

I guess that something is slightly different on the CI server compared to my local build, and as a result the Lambda is getting a different ID.

Unfortunately, I can see that this would be a problem even with a local build - if I added a new lambda at the top of a class, I would expect the lambdas later on to get renumbered, and that would cause the freezing to fail.

Is there any way we can get the ID of lambdas to be ignored - in the same way that line numbers are ignored to support refactoring?

hankem:
Resolves #248.

I haven't checked the performance with a larger number of violations yet.

I was thinking of why we're using `Character.isDigit` instead of `'0' <= ch && ch <= '9'`, and wonder whether `SourceCodeLocation.formatLocation` shouldn't actually use 
```
String.format(Locale.ENGLISH, LOCATION_TEMPLATE, sourceFileName, lineNumber)
```
to avoid a dependency on `Locale.getDefault`.

It wouldn't matter for the current default `FuzzyLineNumberMatcher`, but a custom `ViolationLineMatcher` could benefit from violations not having localized numbers such as `(MyClass.java:١٢٣)` or `(MyClass.java:১২৩)`.

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/library/freeze/FreezingArchRule.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/freeze/FreezingArchRule.java
@@ -59,7 +59,8 @@ import static com.tngtech.archunit.library.freeze.ViolationStoreFactory.FREEZE_S
  * </li>
  * <li>
  *   a {@link ViolationLineMatcher} to decide which violations are "known", i.e. have already been present in the previous evaluation.<br>
- *   The default {@link ViolationLineMatcher} compares violations ignoring the line number of their source code location.<br>
+ *   The default {@link ViolationLineMatcher} compares violations ignoring the line number of their source code location
+ *   and auto-generated numbers of anonymous classes or lambda expressions.<br>
  *   A custom implementation can be configured in two ways.
  *   Again either programmatically via {@link #associateViolationLinesVia(ViolationLineMatcher)}, or within
  *   {@value com.tngtech.archunit.ArchConfiguration#ARCHUNIT_PROPERTIES_RESOURCE_NAME}, e.g.
--- a/archunit/src/main/java/com/tngtech/archunit/library/freeze/ViolationLineMatcherFactory.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/freeze/ViolationLineMatcherFactory.java
@@ -15,9 +15,6 @@
  */
 package com.tngtech.archunit.library.freeze;
 
-import java.util.Iterator;
-
-import com.google.common.base.Splitter;
 import com.tngtech.archunit.ArchConfiguration;
 import com.tngtech.archunit.core.MayResolveTypesViaReflection;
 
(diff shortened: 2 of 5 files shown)


hankem:
It seems that I'll be able to come up with a more performant implementation.

hankem:
Sure! 😃 (If you wanted to make `FuzzyViolationLineMatcher` fully compliant with `java.util.Iterator<String>`, its `next()` should get an
```
                if (!hasNext()) {
                    throw new NoSuchElementException();
                }
```
– but I wouldn't think that this is necessary for a private helper class.)

Regarding the performance: 
I thought that `relevantSubstringMatches`'s `this.end - this.start == other.end - other.start` would have provided a really cheap early exit. I'll check again how much more expensive the full `substring(equals(substring))` really is, but this shouldn't prevent you from merging the PR.

hankem:
FYI: I've repeated my "performance benchmark" measuring the time spent in `ViolationLineMatcher.matches` in an example project with 5.5k stored violations.
On the basis of 10 manual repetitions, the difference between your solution and mine is below the scatter of the data, whereas the regex-based solution is ~8× slower (and this measurement _is_ significant!).

codecholeric:
Thank you so much for your effort :smile: Did you try the thing with the early exit, if it improves performance? Otherwise I would merge this, in the end we can always add another performance improvement in a separate PR :wink:

hankem:
Ah, sorry for being unclear. With
> the difference between your solution and mine is below the scatter of the data

I meant that I cannot distinguish the performance of
* 83579a234f487eb72882f345e7294fd85e9126f0 with the early-exit "optimization", and
* e668fced9f297156c4551f6c4173774df29e320f with the simplified `substring.equals(substring)` code.

(Yes, I know... _"Premature optimization is the root of all evil!"_ 😉)

And the earlier comment
> this shouldn't prevent you from merging the PR.

was meant as: _"Thanks for the review. Please go ahead and merge."_ 😜

Issue #271, opened 2019-11-28
Allow to define Onion Architecture without Domain Services


pszymczyk:
Hi

I've just bump ArchUnit to 0.12.0 and one of my tests failed. It's because I have a module which has no Domain Services - only Application Services and Domain Model components. From my understanding  this situation is totally ok and still consistent with all Onion Architecture principles, that's why in my opinion Domain Services layer should be an optional layer. What it's your point of view? If you agree I can proceed with Pull Request solving this issue.

codecholeric:
Resolved by #273

Issue #177, opened 2019-05-22
empty layers should fail the test


Bananeweizen:
Copy the layered architecture example from the user guide into an arbitrary project and run it. The test passes, since none of the layer definitions maps to any class, field etc. in that project.

I suggest to fail every test where at least one layer definition does not contain any java artifact. This helps avoiding errors in multiple scenarios:
* When starting such tests initially and getting the layer setup wrong.
* When refactoring an existing code base and therefore no longer matching anything. E.g. rename the "persistence" package in the example to "database", and the layered architecture check stops doing anything useful, but it is still green in every test run.

codecholeric:
I think the reason it was originally not done like this, is the possibility to define your layered architecture right in the beginning, even if the layers are not filled yet. But I see your point, it might indeed be better to just fail and force people to only add layers once they also have a class in that layer :wink:

I'd be definitely open for a change here, if anybody wants to jump in and submit a PR. Otherwise I'll change it once I get around to it, but my task pipe is horribly crowded :frowning_face:

davidkarlsen:
Do you have a suggestion for workaround?
I used to have a common test included in all my modules, which stated the layers, among them that my top-layers could not be accessed by any layer.
e.g. layers:
- dao
- service
- common
- model
and then 
- batch
- web

batch cannot be called by any module (as it is a final application module) and the ws-module cannot be called by any else.

I could extract these tests into the layers (e.g. batch and web) - but of course these would never catch anything as a layer can never state anything about others layers access INTO this layer.

davidkarlsen:
ah - I see an `withOptionalLayers(true)` - I'll go with that

PR #273, opened 2019-12-02
Allow empty layers in LayeredArchitecture & OnionArchitecture when configured

### LINKED ISSUE #267: Empty layers of LayeredArchitecture in multi-module project
The `0.12.0` version forbids empty layers of LayeredArchitecture to avoid accidental misbehavior. 

Unfortunately, this breaks our use-case where we use ArchUnit in multi-module project. Our project consists of approx. 60 gradle modules where all share the same LayeredArchitecture configuration. But not all modules have all the layers as i.e. some features just don't need to persist any data to database. 

Currently we are not able to update to `0.12.0` version as this would force us to create and maintain separate LayeredArchitecture configuration for each of them.

Is there any other option we can leverage on in our use case?
If not, would it be possible to make this behavior configurable - enabled by default but allowed to be disabled?

hankem:
Resolves #267 & #271:

After #177 had prevented empty layers by default, this can now be overridden to re-allow empty layers in `LayeredArchitecture` & `OnionArchitecture`.

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/library/Architectures.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/Architectures.java
@@ -102,7 +102,7 @@ public final class Architectures {
         private final Set<LayerDependencySpecification> dependencySpecifications;
         private final PredicateAggregator<Dependency> irrelevantDependenciesPredicate;
         private final Optional<String> overriddenDescription;
-        private boolean allowEmptyLayers;
+        private boolean optionalLayers;
 
         private LayeredArchitecture() {
             this(new LayerDefinitions(),
@@ -116,22 +116,22 @@ public final class Architectures {
                 Set<LayerDependencySpecification> dependencySpecifications,
                 PredicateAggregator<Dependency> irrelevantDependenciesPredicate,
                 Optional<String> overriddenDescription,
-                boolean allowEmptyLayers) {
+                boolean optionalLayers) {
             this.layerDefinitions = layerDefinitions;
             this.dependencySpecifications = dependencySpecifications;
             this.irrelevantDependenciesPredicate = irrelevantDependenciesPredicate;
             this.overriddenDescription = overriddenDescription;
-            this.allowEmptyLayers = allowEmptyLayers;
+            this.optionalLayers = optionalLayers;
         }
 
         /**
          * By default, layers defined with {@link #layer(String)} must not be empty, i.e. contain at least one class.
-         * <code>allowEmptyLayers(true)</code> can be used to skip this consistency check for all layers.
+         * <code>withOptionalLayers(true)</code> can be used to make all layers optional.
          * @see #optionalLayer(String)
          */
         @PublicAPI(usage = ACCESS)
-        public LayeredArchitecture allowEmptyLayers(boolean allowEmptyLayers) {
-            this.allowEmptyLayers = allowEmptyLayers;
+        public LayeredArchitecture withOptionalLayers(boolean optionalLayers) {
+            this.optionalLayers = optionalLayers;
             return this;
         }
 
@@ -148,8 +148,7 @@ public final class Architectures {
         /**
          * Starts the definition of a new layer within the current {@link #layeredArchitecture() LayeredArchitecture}.
          * <br>
-         * Unless {@link #allowEmptyLayers(boolean) allowEmptyLayers(true)} is used, this layer must not be empty.
-         * @see #allowEmptyLayers(boolean)
+         * Unless {@link #withOptionalLayers(boolean) withOptionalLayers(true}} is used, this layer must not be empty.
          * @see #optionalLayer(String)
          */
         @PublicAPI(usage = ACCESS)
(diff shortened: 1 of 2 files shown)


archunit/src/test/java/com/tngtech/archunit/library/ArchitecturesTest.java
```diff
@@ -303,6 +312,32 @@ public void onion_architecture_gathers_all_violations() {
...
+        if (allowEmptyLayers != null) {
+            architecture.allowEmptyLayers(allowEmptyLayers);
+        }
+
+        JavaClasses classes = new ClassFileImporter().importPackages(absolute("onionarchitecture"));
+
+        EvaluationResult result = architecture.evaluate(classes);
+        boolean expectViolation = allowEmptyLayers != Boolean.TRUE;
+        assertThat(result.hasViolation()).as("result of evaluating empty layers has violation").isEqualTo(expectViolation);
+        if (expectViolation) {
```
codecholeric: I'm wondering, if it would be better to have either two tests, or specify given and the assertion in the data point. I find conditionals in assertions harder to read and harder to maintain in the long run. What do you think?
I would see as alternatives to either extract the common given into a method or pass more into the data point, e.g. pass a function pair `given{ architecture -> architecture.allowEmptyLayers(true) }` and `expected{ failureReport -> assertThat(failureReport.isEmpty()).as("failure report").isTrue(); }` as data point.

archunit/src/main/java/com/tngtech/archunit/library/Architectures.java
```diff
@@ -426,6 +438,12 @@ public OnionArchitecture adapter(String name, String... packageIdentifiers) {
             return this;
         }
 
+        @PublicAPI(usage = ACCESS)
+        public OnionArchitecture allowEmptyLayers(boolean allowEmptyLayers) {
```
codecholeric: I wonder if `allowEmptyLayers` is the right expression, when we switch the context to `onionArchitecture`. Because the whole point of the Onion Architecture infrastructure was pretty much to provide better naming than "layers". Unfortunately I don't have any spontaneous ideas for better wording :thinking: What is a common name for the parts of the Onion Architecture?

archunit/src/test/java/com/tngtech/archunit/library/ArchitecturesTest.java
```diff
@@ -134,24 +134,33 @@ public void layered_architecture_defining_constraint_on_non_existing_origin_is_r
...
 
     @Test
-    public void layered_architecture_defining_empty_layers_is_rejected() {
+    @DataProvider(value = {"true", "false", "null"})
+    public void layered_architecture_allows_or_rejects_empty_layers(Boolean allowEmptyLayers) {
         LayeredArchitecture architecture = layeredArchitecture()
                 .layer("Some").definedBy(absolute("should.not.be.found.."))
                 .layer("Other").definedBy(absolute("also.not.found"))
                 .layer("Okay").definedBy("..testclasses..");
+        if (allowEmptyLayers != null) {
```
codecholeric: Same as below. For me it is also hard to immediately grasp at the first look, that those two conditional blocks relate to each other and are only set in conjunction...

archunit/src/main/java/com/tngtech/archunit/library/Architectures.java
hankem: I believe that [The Onion Architecture](https://jeffreypalermo.com/2008/07/the-onion-architecture-part-1/) still consists of _"layers"_ – even though the architecture as a whole is considered a replacement for a _layered_ architecture.

archunit/src/test/java/com/tngtech/archunit/library/ArchitecturesTest.java
hankem: Alternatively, we can also extract helper methods such as
* `private LayeredArchitecture aLayeredArchitectureWithEmptyLayers()`
* `private void assertFailureLayeredArchitectureWithEmptyLayers(EvaluationResult result)`

and explicitly unroll the data points into
* `@Test public void onion_architecture_rejects_empty_layers_by_default()`
* `@Test public void onion_architecture_allows_empty_layers_if_configured_to_allow()`
* `@Test public void onion_architecture_rejects_empty_layers_if_explicitly_configured_to_not_allow()`

archunit/src/test/java/com/tngtech/archunit/library/ArchitecturesTest.java
hankem: There's no 1:1 correspondence, but an explicit test is probably still better than this implicit construction.

archunit/src/main/java/com/tngtech/archunit/library/Architectures.java
codecholeric: Hmm, I guess so, in the end an onion kind of has "layers" :wink:

codecholeric:
@hankem do you have any strong opinion regarding the discussion of #278 ? To me it seems that either both of these features, or none of them should add additional information to the description, since they are of similar type...

hankem:
No, I don't have a strong opinion.

I initially considered `allowEmptyLayers` rather as a technical configuration of the ArchRule, similar to `@AnalyzeClasses`'s `packages` or `importOptions`, which have IMO a similar relevance (Whether layers are empty or not obviously depends on which classes are analyzed...), but are not included in the test report either, so I hadn't even thought about it.

A difference between those and this configuration specific to `LayeredArchictecture` or `OnionArchitecture` is, however, that we could easily adapt the description in this case, so if @roxspring has good reasons to do so, I can try to adapt the description for `allowEmptyLayers` as well.

codecholeric:
Looks noice now, thank you so much :smile: 
Glad we can take away the pain the increased restrictions have caused!

PR #275, opened 2019-12-09
Add archunit-junit5 pom aggregator

### LINKED ISSUE #272: Provide an Aggregator POM for archunit-junit5
Imho it would be great if you'd provide a maven bill of materials. So you can import it into the dependencyManagement and have all versions defined in one place.

eddumelendez:
Resolves #272

### DIFF
--- a/archunit-example/example-junit5/build.gradle
+++ b/archunit-example/example-junit5/build.gradle
@@ -4,10 +4,8 @@ sourceCompatibility = JavaVersion.VERSION_1_8
 targetCompatibility = JavaVersion.VERSION_1_8
 
 dependencies {
-    testCompile project(path: ':archunit-junit5-api')
+    testCompile project(path: ':archunit-junit5')
     testCompile project(path: ':archunit-example:example-plain')
-
-    testRuntime project(path: ':archunit-junit5-engine')
 
     testRuntime dependency.log4j_api
     testRuntime dependency.log4j_core
--- a/archunit-junit/junit5/aggregator/build.gradle
+++ b/archunit-junit/junit5/aggregator/build.gradle
@@ -4,3 +4,7 @@ dependencies {
 	compile(project(":archunit-junit5-api"))
 	runtime(project(":archunit-junit5-engine"))
 }
+
+shadowJar {
+	exclude '**'
+}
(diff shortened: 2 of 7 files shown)


codecholeric:
Thanks a lot for looking into this :smiley: It seems to work, but I have two things to consider

a) I did get some exception during `archunit-junit5:shadowJar`, which actually did not let the build fail or broke the result, but we should still fix this (I did not see it in the console output of the Travis build though)
b) a bigger thing I wonder is, if we really want to create all those artifacts. I.e. now we create an empty JAR, and empty javadoc and an empty sources JAR. I've noticed that the JUnit team does the same, but I wonder why we would not simply use a POM of type `pom`? Similar to https://repo1.maven.org/maven2/com/google/guava/guava-parent/28.1-jre/. Shouldn't a simple POM without any artifacts attached to it also do the trick?

eddumelendez:
@codecholeric the exception is displayed building against java 11 at least in my machine. You can see my findings below:

a) The new commit resolves the issue building against java 11. I couldn't upgrade to `4.0.4` due to one issue which was supposed to be fixed in that version but it wasn't until version `5.x`. Upgrading to `5.x` requires gradle version. I think there is more to see regarding this dependency upgrade. You can see more about `shadow` plugin building against java 11 in the [release notes](https://github.com/johnrengelman/shadow/releases)

b) That would be something that gradle understand and can resolve the dependencies but maven users will need to specify an additional line `type` meanwhile gradle users will not.  Keeping the jar will benefit both.

codecholeric:
@eddumelendez ah, cool, thanks a lot for looking into this :smile: 

a) Yes, we need to upgrade Gradle :frowning_face: Luckily there is light at the horizon thanks to @aaschmid (see #192)
b) Okay, I didn't consider the implication with the additional `type` :wink: Since the aggregator POM strives only to make including ArchUnit as simple as possible, I agree! It's probably best to produce the triplet of JAR, sources JAR and javadoc JAR then...

codecholeric:
I have looked over it, looks good now :smile:
I noticed that the upgraded Shadow plugin now was "too good" :wink: (started to aggregate some 3rd party stuff into `archunit-junit5.jar`) I've adjusted the config to simply exclude everything. Seems reasonable now, three empty JARs with just a Manifest and a POM with the correct dependencies.
I've adjusted the Maven integration test to now test JUnit 5 with the new aggregator dependency and it seems to work.
So thank you so much for your support!! :smiley: Gonna merge it now...

PR #278, opened 2019-12-14
Add `ensureAllClassesAreContainedInLayers()` to Architectures

### LINKED ISSUE #222: Is there a way log/trace files that are not covered by any layer of a LayeredArchitecture?
Defining layers and allowed access is easy but how to check if all loaded classes are covered by at least one layer?

roxspring:
This will add possibilities to ensure that all classes under test are contained within the respective `LayeredArchitecture`/`OnionArchitecture`. It will help users to make sure they don't overlook any classes when defining their architectures. Furthermore, it will help with maintainability of the architectures when new classes are added to the code base later on. In detail the following methods have been added to `LayeredArchitecture` and `OnionArchitecture`:

* `ensureAllClassesAreContainedInArchitecture()`
* `ensureAllClassesAreContainedInArchitectureIgnoring(packageIdentifiers)`
* `ensureAllClassesAreContainedInArchitectureIgnoring(predicate)`

Resolves #222

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/library/Architectures.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/Architectures.java
@@ -766,6 +766,7 @@ public final class Architectures {
         private Map<String, DescribedPredicate<? super JavaClass>> adapterPredicates = new LinkedHashMap<>();
         private boolean optionalLayers = false;
         private List<IgnoredDependency> ignoredDependencies = new ArrayList<>();
+        private AllClassesAreContainedInArchitectureCheck allClassesAreContainedInArchitectureCheck = new AllClassesAreContainedInArchitectureCheck.Disabled();
 
         private OnionArchitecture() {
             overriddenDescription = Optional.empty();
@@ -856,6 +857,46 @@ public final class Architectures {
             return this;
         }
 
+        /**
+         * Ensure that all classes under test are contained within a defined onion architecture component.
+         *
+         * @see #ensureAllClassesAreContainedInArchitectureIgnoring(String...)
+         * @see #ensureAllClassesAreContainedInArchitectureIgnoring(DescribedPredicate)
+         */
+        @PublicAPI(usage = ACCESS)
+        public OnionArchitecture ensureAllClassesAreContainedInArchitecture() {
+            return ensureAllClassesAreContainedInArchitectureIgnoring(alwaysFalse());
+        }
+
+        /**
+         * Like {@link #ensureAllClassesAreContainedInArchitecture()} but will ignore classes in packages matching
+         * the specified {@link PackageMatcher packageIdentifiers}.
+         *
+         * @param packageIdentifiers {@link PackageMatcher packageIdentifiers} specifying which classes may live outside the architecture
+         *
+         * @see #ensureAllClassesAreContainedInArchitecture()
+         * @see #ensureAllClassesAreContainedInArchitectureIgnoring(DescribedPredicate)
+         */
+        @PublicAPI(usage = ACCESS)
+        public OnionArchitecture ensureAllClassesAreContainedInArchitectureIgnoring(String... packageIdentifiers) {
+            return ensureAllClassesAreContainedInArchitectureIgnoring(resideInAnyPackage(packageIdentifiers));
+        }
+
+        /**
+         * Like {@link #ensureAllClassesAreContainedInArchitecture()} but will ignore classes in packages matching
+         * the specified {@link DescribedPredicate predicate}.
+         *
+         * @param predicate {@link DescribedPredicate predicate} specifying which classes may live outside the architecture
+         *
+         * @see #ensureAllClassesAreContainedInArchitecture()
+         * @see #ensureAllClassesAreContainedInArchitectureIgnoring(String...)
+         */
+        @PublicAPI(usage = ACCESS)
+        public OnionArchitecture ensureAllClassesAreContainedInArchitectureIgnoring(DescribedPredicate<? super JavaClass> predicate) {
+            allClassesAreContainedInArchitectureCheck = new AllClassesAreContainedInArchitectureCheck.Enabled(predicate);
+            return this;
+        }
+
         private DescribedPredicate<JavaClass> byPackagePredicate(String[] packageIdentifiers) {
             return resideInAnyPackage(packageIdentifiers).as(joinSingleQuoted(packageIdentifiers));
         }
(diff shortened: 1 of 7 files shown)


roxspring:
@codecholeric can you take a look at this and let me know how it compares to what you imagined? I'm fairly new to ArchUnit and wanted to double check the direction before going much further.

archunit/src/test/java/com/tngtech/archunit/library/ArchitecturesTest.java
```diff
@@ -218,6 +219,55 @@ public void layered_architecture_combines_multiple_ignores() {
...
+
+    @Test
+    public void layered_architecture_where_all_classes_are_covered_with_ignoring() {
+        JavaClasses classes = new ClassFileImporter().importClasses(
+                FirstAnyPkgClass.class,
+                SecondThreeAnyClass.class);
+
+        LayeredArchitecture architecture = layeredArchitecture()
+                .layer("One").definedBy("..first..")
+                .layer("Two").definedBy("..second..")
```
codecholeric: Maybe it would be better if the test would fail in case `covered...ignoring...` would not work?

archunit/src/main/java/com/tngtech/archunit/library/Architectures.java
```diff
@@ -167,9 +175,65 @@ public EvaluationResult evaluate(JavaClasses classes) {
...
                 result.add(evaluateDependenciesShouldBeSatisfied(classes, specification));
             }
+            if (allClassesAreCovered.isPresent()) {
+                result.add(allClassesAreCovered.get().evaluate(classes, layerDefinitions));
+            }
             return result;
         }
 
+        @PublicAPI(usage = ACCESS)
+        public LayeredArchitecture whereAllClassesAreCovered() {
```
codecholeric: I'm not sure if the expression "covered" is in general easy to understand here. Maybe `whereAllClassesAreContainedInLayers`? Just thinking that "covered" is used nowhere else in ArchUnit, as far as I remember, but "contained" is used.
Or even name it `ensureAllClassesAreContainedInLayers` or similar, since all the `where` clauses are part of the specification, i.e. "this is how our layered architecture is defined", while this here is more like a configuration how this layered architecture should be tested against the classes supplied.

archunit/src/test/java/com/tngtech/archunit/library/ArchitecturesTest.java
```diff
@@ -218,6 +219,55 @@ public void layered_architecture_combines_multiple_ignores() {
...
+                .whereAllClassesAreCovered();
+
+        assertThat(architecture.getDescription()).contains("where all classes are covered");
```
codecholeric: Similar to my comment in `Architectures`, I'm not sure if we actually should adjust the description. In particular we should probably make it consistent to #273 where as far as I saw the description is not adjusted. But this is pretty much the same thing, a configuration of *how* the layered architecture should be tested against the input, not about the layered architecture itself.

codecholeric:
Thanks a lot for looking into this :smiley: I've added a couple of thoughts, in particular I think it is important that we compare this to #273, since this feels to me like two sides of the same coin, don't you think? Both deal with a kind of configuration of how I want to test the layered architecture.
Do I want to allow empty layers when I test it? Do I want to allow some classes not covered by any layer at all? This does not influence the basic structure of the layered architecture, it's more "meta-information" of how I want to test the supplied set of classes.

archunit/src/test/java/com/tngtech/archunit/library/ArchitecturesTest.java
roxspring: Thought I was covering the failure cases too but will check :)

archunit/src/main/java/com/tngtech/archunit/library/Architectures.java
roxspring: Fair enough. `ensureAllClassesAreContainedInLayers` and similar seem very verbose and so I was shooting for something more succinct - but apparently lost clarity in the process. Also hadn't picked up on the `where` prefix suggesting the spec rather than a test. Will rework based on your naming.

archunit/src/test/java/com/tngtech/archunit/library/ArchitecturesTest.java
roxspring: Interesting. Omitting this from the description seems counter-intuitive but I'll have a play and try and convince myself one way or the other.

roxspring:
Thanks for the review. I agree this relates very closely to #273 and that consistency is important. I'll take a closer look at the code there and try to bring my branch in line with that one.

archunit/src/test/java/com/tngtech/archunit/library/ArchitecturesTest.java
codecholeric: Sure! To me the description is like a report of the architecture spec. But I have no strong opinion, if this should be part of the spec or not. I just have a strong opinion that #273 and #278 should do it the same way, one way or the other :wink:

archunit/src/test/java/com/tngtech/archunit/library/ArchitecturesTest.java
roxspring: Doh. Fixed in 88cbd3e!

archunit/src/test/java/com/tngtech/archunit/library/ArchitecturesTest.java
roxspring: I've dropped the descriptions for now in f431a75. Perhaps later I'll open a separate merge request adding descriptions for both related cases when I've had chance to make a convincing argument.

archunit/src/main/java/com/tngtech/archunit/library/Architectures.java
roxspring: Done in a9740ba.

clojj:
Hi, nice PR... can anything be done here ?

codecholeric:
Hey, I took the liberty to fix up and finalize this PR, since it seemed a little orphaned to me :wink: Hope that's alright with you!!

roxspring:
Please feel free! I've changed jobs since starting this, and although I've introduced use of ArchUnit, we don't have a strong layered architecture so this has become a very low priority 😕

codecholeric:
No worries, I somehow completely lost track :see_no_evil: But I think this will be a really nice addition that users will appreciate :slightly_smiling_face:

GandalfIX (approved):
This is a nice and simple addition which will help in our project as we are right within a transition from layer to onion :-)

clojj:
Nice!👌

PR #284, opened 2019-12-21
Migrate build to "maven-publish" plugin

aaschmid:
Finally got it running @codecholeric. But haven't had a look at the new "shadow" plugin and it's possibilities in order to get rid of dependencies removal in published poms, yet.

See commit message for still open points and tell me what you think about these as well as with I should get fixed.

You can check the results of a `publish` run at https://oss.sonatype.org/content/repositories/snapshots/com/tngtech/archunit/, version `0.13.0-SNAPSHOT` or run `publishToMavenLocal` yourself. I also created a release (which is again dropped) and run `checkUAr -PtngRepoId=1202` besides some pom formattings - at least as far as I could see manually ;-) - quite successfully IMO.
Note: Only thing was that I needed to run `./gradlew --no-parallel publish` multiple times to have all artifacts in the OSS Sonatype repo but could not find out why this is the case ...

Unfortunately, the upgrade to Gradle version 5 (first commit) is required unless you want to add `enableFeaturePreview('STABLE_PUBLISHING')` to your settings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: java
 
 script:
-- ./gradlew --no-daemon -PallTests -PscanBuild clean build install runMavenTest
+- ./gradlew --no-daemon -PallTests -PscanBuild clean build publishToMavenLocal runMavenTest
 
 jdk:
   - openjdk11
--- a/archunit-junit/build.gradle
+++ b/archunit-junit/build.gradle
@@ -84,8 +84,15 @@ ext.configureJUnitArchive = { configureDependencies ->
         delegate.with addCleanThirdPartyTask
         compileJava.dependsOn project(':archunit-junit').finishArchive
 
-        install.repositories.mavenInstaller.pom.whenConfigured configureDependencies
-        uploadArchives.repositories.mavenDeployer.pom.whenConfigured configureDependencies
+        publishing{
+            publications{
+                mavenJava{
+                    pom.withXml {
+                        configureDependencies(asNode().dependencies.first()) // there is only one "dependencies" element
+                    }
+                }
+            }
+        }
     }
 }
 
--- a/archunit-junit/junit4/build.gradle
+++ b/archunit-junit/junit4/build.gradle
@@ -43,10 +43,9 @@ shadowJar {
     }
 }
 
-def configureDependencies = { pom ->
-    pom.dependencies.removeAll {
-        it.scope != 'compile' || !(it.artifactId in ['archunit', 'archunit-junit5-api', 'archunit-junit5-engine-api', 'junit'])
+def configureDependencies = { deps ->
+    deps.children().removeIf { dep ->
+        dep.scope.text() != 'compile' || !(dep.artifactId.text() in ['archunit', 'archunit-junit5-api', 'archunit-junit5-engine-api', 'junit'])
     }
-    pom.dependencies.find { it.artifactId == 'archunit' }.classifier = null
 }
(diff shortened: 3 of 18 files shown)


aaschmid:
BTW: I like the check logic for the upload quite much. Maybe I will use it for the [/TNG/junit-dataprovider](junit-dataprovider) as well some day ...

aaschmid:
@codecholeric: Any feedback on this?

codecholeric:
Thanks a lot for your support! :smiley: 

I've rebased this against master and fixed some parts. To answer the questions from your commit message:

* I think we should keep the parallel check, because I've had many problems with broken uploads otherwise. I've improved it though, that it also detects if you run with `--no-parallel`
* I removed the distributed `repackagesAsm`. Seemed like a good idea at the time, but in fact it will always only be `archunit` that repackages ASM, any other ArchUnit artifact would simply depend on that within `archunit`. And it only affects license handling, so I think it's fine to keep it locally within `license.gradle` (the way you had it)
* I don't think we need the `signing.dependsOn finishArchive`, because the order seems correct anyway. `publish` depends on `build` and `build` on `finishArchive`, so I don't think we have a problem there anymore.
* I did not completely get your last point about "refactor code to use JUnit* module mechanism"

There is one problem left (maybe you know how to fix this out of your head?): In the JUnit 5 aggregator POM the `scope` of the `runtime` dependency `archunit-junit5-engine` is now suddenly `compile` instead of `runtime` like before :frowning_face:

aaschmid:
Thanks for the rebase and the reply.

> There is one problem left (maybe you know how to fix this out of your head?): In the JUnit 5 aggregator POM the scope of the runtime dependency archunit-junit5-engine is now suddenly compile instead of runtime like before.

I think I know how to fix and will try during the next days.

aaschmid:
868b6aa fixes your last issue. I (only) tested it locally by using your `checkUploadedArtifacts` with `file:/home/user/.m2/repository/...` :-)

aaschmid:
Addtional note:

I had to use `--no-build-cache` in order to get reproducable results. As soon as I used the build cache I got blinking "tests" (= failing/succeeding `checkUploadedArtifcats` tasks) with:
```
There exists a third party folder. Expression: (jarFile.getEntry(com/tngtech/archunit/thirdparty) == null)
```
I haven't looked at the root cause intensively but on a quick look it turns up that `addCleanThirdPartyTask` could likely cause this behavior.

aaschmid:
Follow up PR upgrading to Gradle 6 can be find on https://github.com/TNG/ArchUnit/tree/gradle.6

codecholeric:
Looks good now :smile: I've tested it a little more as well, artifacts, manifests and POMs look good and the project where I tried the aggregator POM worked fine as well.
It's well possible, that there is something shifty going on with the `cleanThirdPartyTask`, that was always a little hacky :frowning_face: Maybe once we've migrated everything to Gradle 6 and `java-library` we can model this in a correct way and get rid of this abomination...
Anyway, thank you so much for your support! I'll look into the Gradle 6 upgrade ASAP, then we're finally current again :smile:

aaschmid:
@codecholeric: I just rebased the https://github.com/TNG/ArchUnit/tree/gradle.6 branch. I haven't had time to try to cleanup the dependencies more properly, especially test dependencies but IMO this should be done also.
The build is working on Gradle 6 branch and locally tests on the deployment also. Feel free to integrate it :-)

However, trying to further use the shadow plugin provided features was not successful yet. And I also wanted to clean up / simplifiy the test dependencies which was not time for yet.

PR #289, opened 2019-12-31
Add GeneralCodingRule to discourage use of field injection

### LINKED ISSUE #288: Default condition to avoid field Injection and haveConstructor Injection
I'm fascinated by the API and have started using it our project.Lately i discovered that few 
people in our project were using fieldInjection and not constructor injections 

So we thought of using ArchRule and was expecting a default rule as below .

         @ArchTest
    public static final ArchRule NO_FIELD_WITH_AUTOWIRED_ANNOTATION_AND_HAVE_CDI = 
                   classes().that().resideInAPackage("xxx.xxx..xxx.xxx")
            .should(NO_AUTOWIRING).andShould(HAVE_CONSTRUCTOR_DEPENDANCY_INJECTION);

But by default it was not available ,so we wrote a conditions to it,

 Avoiding Field Injection
   
          @Override
    public void check(JavaClass javaClass, ConditionEvents conditionEvents) {
         javaClass.getAllFields().stream().filter(javaField -> 
           javaField.isAnnotatedWith(Autowired.class))
              .forEach(conditionEvent -> 
       conditionEvents.add(SimpleConditionEvent.violated(conditionEvent,
                      String.format("class %s contains a Field %s with AutoWired",javaClass.getName() , 
           conditionEvent))));
           }

Ensuring CDI

         @Override
    public void check(JavaClass javaClass, ConditionEvents conditionEvents) {

        List<Dependency> classDependencies = 
           javaClass.getDirectDependenciesFromSelf().stream().collect(Collectors.toList());

        List<String> constructorParams = getConstructorParams(classDependencies);

        List<JavaField> classFields = javaClass.getAllFields().stream().collect(Collectors.toList());
        classFields.stream().filter(classField ->  
           !constructorParams.contains(classField.getRawType().getName()))
                .forEach(conditionEvent -> 
         conditionEvents.add(SimpleConditionEvent.violated(conditionEvent,
                        String.format("class %s  has dependency on %s which is not properly injected in 
         the constructor", javaClass.getName(), conditionEvent))));
       }

     private List<String> getConstructorParams(List<Dependency> classDependencies) {
        return classDependencies.stream()
                .filter(classDependency -> 
           !classDependency.getTargetClass().isEquivalentTo(Object.class))
                .filter(classDependency -> 
          classDependency.getDescription().startsWith("Constructor"))
                .map(filteredClassDependency -> 
           filteredClassDependency.getTargetClass().getName()).collect(Collectors.toList());
     }

Would like to know if it can be added as default behaviour in the API.

Regards,
Faizal

rweisleder:
Resolves #288


archunit/src/main/java/com/tngtech/archunit/library/GeneralCodingRules.java
```diff
@@ -151,4 +154,28 @@ private GeneralCodingRules() {
...
     public static final ArchRule NO_CLASSES_SHOULD_USE_JODATIME =
             noClasses().should(USE_JODATIME).because("modern Java projects use the [java.time] API instead");
+
+    /**
+     * For information about checking this condition, refer to {@link GeneralCodingRules}.
+     */
+    @PublicAPI(usage = ACCESS)
+    public static final ArchCondition<JavaField> USE_FIELD_INJECTION = useFieldInjection().as("use field injection");
+
+    private static ArchCondition<JavaField> useFieldInjection() {
```
codecholeric: technically this is not `useFieldInjection`, buy `annotatedWithAnInjectionAnnotation` or similar, isn't it? :wink:

archunit/src/main/java/com/tngtech/archunit/library/GeneralCodingRules.java
```diff
@@ -151,4 +154,28 @@ private GeneralCodingRules() {
...
+    }
+
+    /**
+     * Field injection is seen as an anti-pattern.
+     * It is a good practice to use constructor injection for mandatory dependencies and setter injection for optional dependencies.
+     * <br>
+     * For information about checking this rule, refer to {@link GeneralCodingRules}.
+     */
+    @PublicAPI(usage = ACCESS)
+    public static final ArchRule NO_CLASSES_SHOULD_USE_FIELD_INJECTION =
```
codecholeric: The text of the rule and field doesn't really match now, right? I would either call both "no fields should be injected" or "no classes should use field injection", but now the field name is "no classes..." but the rule text will then be "no fields..." or am I overlooking something here? Also "no fields should use field injection" sounds a little strange to me :wink:

archunit/src/main/java/com/tngtech/archunit/library/GeneralCodingRules.java
```diff
@@ -151,4 +154,28 @@ private GeneralCodingRules() {
...
+            noFields().should(USE_FIELD_INJECTION).because("field injection is considered harmful");
```
codecholeric: Should we add more reason here? Maybe some link with more information? Because this way the "because" sounds more like a dogma "you should not do foo, because it's bad" :wink:

archunit/src/main/java/com/tngtech/archunit/library/GeneralCodingRules.java
```diff
@@ -151,4 +154,28 @@ private GeneralCodingRules() {
...
             noClasses().should(USE_JODATIME).because("modern Java projects use the [java.time] API instead");
+
+    /**
+     * For information about checking this condition, refer to {@link GeneralCodingRules}.
+     */
+    @PublicAPI(usage = ACCESS)
+    public static final ArchCondition<JavaField> USE_FIELD_INJECTION = useFieldInjection().as("use field injection");
+
+    private static ArchCondition<JavaField> useFieldInjection() {
+        ArchCondition<JavaField> annotatedWithAutowired = beAnnotatedWith("org.springframework.beans.factory.annotation.Autowired");
```
codecholeric: I wrote this already in the issue, it's too bad that this is so framework dependent. Since we're already on it, should we add `com.google.inject.Inject` as well? Or are we sure the kids of today don't use that anymore in favor of `javax.inject.Inject`? But I once was in a project where we could not migrate away, because `com.google.inject.Inject` allows `optional = true` which the standard does not allow...

build.gradle
```diff
@@ -50,7 +50,11 @@ ext {
             junit_dataprovider  : 'com.tngtech.java:junit-dataprovider:1.11.0',
             mockito             : 'org.mockito:mockito-core:2.27.0',
             assertj             : 'org.assertj:assertj-core:2.9.1',
-            assertj_guava       : 'org.assertj:assertj-guava:2.0.1'
+            assertj_guava       : 'org.assertj:assertj-guava:2.0.1',
+
+            springBeans         : 'org.springframework:spring-beans:5.2.2.RELEASE',
+            jakartaInject       : 'jakarta.inject:jakarta.inject-api:1.0',
```
codecholeric: Is this where `javax.inject.Inject` is from? I think there is also a slim standard dependency `javax.inject:javax.inject:1` which we could maybe use?

codecholeric: Even though maybe it's better to take everything from `jakarta`, since `@Resource` is not part of any standard dependency AFAIK.

archunit/src/main/java/com/tngtech/archunit/library/GeneralCodingRules.java
```diff
@@ -151,4 +154,28 @@ private GeneralCodingRules() {
     @PublicAPI(usage = ACCESS)
     public static final ArchRule NO_CLASSES_SHOULD_USE_JODATIME =
             noClasses().should(USE_JODATIME).because("modern Java projects use the [java.time] API instead");
+
+    /**
+     * For information about checking this condition, refer to {@link GeneralCodingRules}.
+     */
+    @PublicAPI(usage = ACCESS)
+    public static final ArchCondition<JavaField> USE_FIELD_INJECTION = useFieldInjection().as("use field injection");
```
codecholeric: Shouldn't overwriting the description happen within a factory method of a condition?

archunit/src/main/java/com/tngtech/archunit/library/GeneralCodingRules.java
hankem: We could also add the hints from the JavaDoc (_"use constructor injection for mandatory dependencies and setter injection for optional dependencies"_) to the `.because` to give the specific advice when the rule fails.

archunit/src/main/java/com/tngtech/archunit/library/GeneralCodingRules.java
```diff
@@ -151,4 +154,28 @@ private GeneralCodingRules() {
...
+
+    private static ArchCondition<JavaField> useFieldInjection() {
+        ArchCondition<JavaField> annotatedWithAutowired = beAnnotatedWith("org.springframework.beans.factory.annotation.Autowired");
+        ArchCondition<JavaField> annotatedWithValue = beAnnotatedWith("org.springframework.beans.factory.annotation.Value");
```
remal: I'm afraid it would be a bit inconvenient, as some standard Spring properties don't have corresponding classes annotated by `@ConfigurationProperties`. For example: `spring.application.name` - this property is usually injected by `@Value`

rweisleder: `@Value` can be used on constructor parameters:
```java
public InfoBean(@Value("spring.application.name") String applicationName) {
  // ...
}
```

So there is no need to use `@Value` on fields.

remal: Oh... I see... I'm OK with it.

build.gradle
rweisleder: `jakarta.inject:jakarta.inject-api` contains `javax.inject.Inject`, and `jakarta.annotation:jakarta.annotation-api` contains `javax.annotation.Resource`.
I took Jakarta because they provide the latest version of these annotations.

archunit/src/main/java/com/tngtech/archunit/library/GeneralCodingRules.java
rweisleder: I added `com.google.inject.Inject`, the more the better 😉

PR #290, opened 2020-01-04
Add API methods to JavaPackage to read package-info.java

### LINKED ISSUE #263: Add rules for packages
I'd like to check that every package has `package-info.java` with `NonNullApi` and `NonNullFields` annotations:
```java
@NonNullApi
@NonNullFields
package com.moex.mp.sys.payment.libs.commontest;

import org.springframework.lang.NonNullApi;
import org.springframework.lang.NonNullFields;
```

Currently I can do it only via reflection, not via bytecode.

Please add a possibility of checking packages' annotations.

rweisleder:
Resolves #263

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaAnnotation.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaAnnotation.java
@@ -80,6 +80,14 @@ public final class JavaAnnotation<OWNER extends HasDescription> implements HasTy
     private final String description;
     private final Map<String, Object> values;
 
+    private JavaAnnotation(JavaClass type, OWNER owner, CanBeAnnotated annotatedElement, String description, Map<String, Object> values) {
+        this.type = checkNotNull(type);
+        this.owner = checkNotNull(owner);
+        this.annotatedElement = checkNotNull(annotatedElement);
+        this.description = checkNotNull(description);
+        this.values = checkNotNull(values);
+    }
+
     JavaAnnotation(OWNER owner, JavaAnnotationBuilder builder) {
         this.type = checkNotNull(builder.getType());
         this.owner = checkNotNull(owner);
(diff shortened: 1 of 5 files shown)


codecholeric:
Thanks a lot!! :smile: Now there is only one problem, with #171 merged, `JavaAnnotation` now has a type parameter of the type of the element that is annotated. Since `package-info` is a "class" for ArchUnit and `HasAnnotations` now also wants the type of the annotated element, this feels a little inconsistent. In particular because one would expect that `JavaPackage.getAnnotations()` would return `JavaAnnotation<JavaPackage>`, but at the moment it could only return `JavaAnnotation<JavaClass>`.
Have to ponder a little how to implement this best, I guess. In particular whether it should be moved closer to the core where `JavaAnnotation<JavaPackage>` could be created, or if it is actually okay like this (or `JavaAnnotation<?>` would also be sufficient as a return type, don't know. But then it's maybe still strange to get the `JavaClass<package-info>` as the `JavaAnnotation.getOwner()`.
What's your opinion here?

rweisleder:
As a user of the API, I would expect
```java
JavaAnnotation<JavaPackage> a = javaPackage.getAnnotationOfType(PackageAnnotation.class);
JavaPackage owner = a.getOwner(); // same as javaPackage
```

And it feels strange to handle package-info as a class because it's not really a class. We should get rid of this:
https://github.com/TNG/ArchUnit/blob/f1d7da342558abc769200020350f478d2c35bec2/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaPackage.java#L63

rweisleder:
If we wanna break the user's code we could make the methods of `HasAnnotations` consistent. 😉 
Currently, we have this:
https://github.com/TNG/ArchUnit/blob/a15d47a64ef6fff4a7bc5f3d8e0f932c7da85016/archunit/src/main/java/com/tngtech/archunit/core/domain/properties/HasAnnotations.java#L33-L43

Some methods return `JavaAnnotation` and other methods with the same name return `Annotation`.

rweisleder:
Yes, looks good to me 👍 thanks, @codecholeric

codecholeric:
Closing and reopening to trigger a new Travis build (the old one seems stuck)

PR #299, opened 2020-01-22
Official Gradle Wrapper Validation GitHub Action

JLLeitschuh:
See: https://github.com/gradle/wrapper-validation-action


codecholeric:
Thanks a lot for making OSS safer!! :smile: This seems indeed to be a completely reasonable check...

PR #308, opened 2020-02-07
Fix version of Gradle in wrapper task

rweisleder:
Belongs to #306 😉

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -109,7 +109,7 @@ ext {
 }
 
 wrapper {
-    gradleVersion = '4.7'
+    gradleVersion = '5.6.4'
 }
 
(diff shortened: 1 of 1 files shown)


codecholeric:
Thanks for paying attention :smile: I've completely removed the Gradle Wrapper task configuration in https://github.com/TNG/ArchUnit/commit/802d90f956715a1888cf850fa8b5ba9534cf7793 though, because I think this is a leftover from old times.
When we upgrade the Gradle Wrapper, we'll do it like `./gradlew wrapper --gradle-version=5.6.4` anyway, and afterwards it's checked in, so I don't really see any use to configure this task with a default version at all.

PR #311, opened 2020-02-09
Add support for records (Java 14 preview feature)

### LINKED ISSUE #295: Add support for records (Java 14 preview feature)
Java 14 will bring us [JEP 359](https://openjdk.java.net/jeps/359) (preview feature) which introduces records as a new kind of type declaration.

We should support records like we support enums. I suggest to add the following methods:
* JavaClass.isRecord
* classes().that().are[Not]Records()
* classes().should().[not]BeRecords()

ASM supports records since 7.3.1 (released yesterday). The [release notes](https://asm.ow2.io/versions.html) are not available yet, but the code contains the [merge request for records](https://gitlab.ow2.org/asm/asm/commit/db802e6eb64784567b81277f28734421b9dc878e). So I guess records should be part of that version.

So that we are able to test the implementation, we need to build ArchUnit against JDK 14 with enabled preview features.
Do we need an additional testing source set in our Gradle build? Or could this be done inside src/test/java?

rweisleder:
I didn't touch the build environment and first implemented the API only.

For testing purposes, I committed a compiled class file, see ClassFileImporterTest: https://github.com/rweisleder/ArchUnit/blob/d8e79f72f0d93251ed67d98330acfdd31619f095/archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterTest.java#L310-L320
As suggested in the issue, we should add a specific source set `src/jdk14test/java`. Then we could move the test to this source set. But therefore we need to make our Gradle build work with JDK 14.

Resolves #295


archunit/src/main/java/com/tngtech/archunit/core/importer/JavaClassProcessor.java
```diff
@@ -141,6 +142,13 @@ private boolean importAborted() {
         return javaClassBuilder == null;
     }
 
+    @Deprecated
+    @Override
+    public RecordComponentVisitor visitRecordComponentExperimental(int access, String name, String descriptor, String signature) {
```
hankem, line 147: With ASM 8 (released [lately](https://asm.ow2.io/versions.html)), this needs to be:
```
public RecordComponentVisitor visitRecordComponent(String name, String descriptor, String signature) {
```

rweisleder: Thanks for pointing out that ASM 8 got released, @hankem! I updated this PR.

codecholeric:
Hey @rweisleder, I've merged the PR to upgrade Gradle to 6.3 :smiley: In theory it should now be possible to build ArchUnit with JDK 14. What do you think, would it make sense to adjust this PR to have the test example in `src/jdk14test/java` instead of checking in a bytecode blob? Might be nicer to see what actually happens there and maybe catch a problem when we upgrade to JDK 15 or similar, if we actually compile that class... WDYT?

rweisleder:
Great, thanks! The upgrade to Gradle 6.3 was the last missing piece to run the build with JDK 14. I'll give it a try later on.

rweisleder:
I updated the PR so that we have a running example in `src/jdk14test/java`.

For `src/jdk14test/java` I just copy-pasted the snippets I found for `jdk9test`, but didn't quite understand what I did there. 😅 
Additionally, the task `jdk14Test` fails with
```
    org.gradle.api.GradleException: JUnit Categories defined but declared JUnit version does not support Categories.
        at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.verifyJUnitCategorySupport(JUnitTestClassExecutor.java:132)
        at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.runTestClass(JUnitTestClassExecutor.java:73)
        at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:58)
        at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:38)
        at org.gradle.api.internal.tasks.testing.junit.AbstractJUnitTestClassProcessor.processTestClass(AbstractJUnitTestClassProcessor.java:62)
        at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.processTestClass(SuiteTestClassProcessor.java:51)
```
This exception also happens when executing `./gradlew archunit:jdk9Test` against the current master. 🤔 
@codecholeric: Maybe you can help here?

rweisleder:
I finally found the time to rebase this PR against the current master and to fix the last issues.
@codecholeric, this PR is now ready for review.

bgalek (approved):
LGTM!

marcelbaumann:
It would be great if support for the record feature is coming soon. Records will be released in JDK 16 in a few weeks.
Our code already has quite a few record declarations and we had to disable all our ArchUnit tests due to that.

I am also a part-time lecturer and using ArchUnit in technical university bachelor lectures. And also teaching how to use records for immutability and functional programming in Java.

I am ready to provide support for testing if desired.

Sincerely

codecholeric:
This is my fault, since I was focusing on generics support recently, and found no time to look into this. I'll try to get this ready soon!

codecholeric:
Okay, I'm really sorry that it took me that ridiculously long :disappointed: I'm trying to balance my priorities always for what affects the most users right now versus what will affect users in the long run. And given my experiences how quickly new Java features are adopted out there I was putting more effort in the direction of generics for the recent months, because I felt that more users were affected by the lack of generics support.
On the other hand this was really nicely prepared, so it took me a lot less time than doing it from scratch :thinking:

Anyway, I finally got around to it, I rebased it, I reviewed it, and I put some stuff I noticed in Fixup commits! Would be super cool if you could go over it and see if it makes sense.

I would also restructure the commits a little bit, for once squash the first 3 commits. Because I feel just raising the min required Java version by itself doesn't provide any value and neither does adding an empty Java 14 source set. I would try to keep the commits that go into the `main` branch each as a self-contained unit that provides value and motivates its content, if possible. So adding records to the domain in one shot together with the Gradle source sets and the build adjustments would be fine I think. Of course the message should still contain all that nice explanation :smiley: Also the very last commit could go to the beginning I think, because it fixes an existing problem that just was revealed by these changes.

This Gradle issue that it tries to parse the bytecode and fails for records in the test source set is rather annoying :disappointed: But I guess we just have to be patient until records are fully supported (and hope that not too much will need to be adjusted for these tests anyway)

One weird thing you can see in the commits is that ASM somewhat only detects the `FINAL` modifier for records, but not the `STATIC` modifier :thinking: At least the test failed once I used the assertion that compares to the Java Reflection API. I have added one commit that adds those modifiers always, because the JLS specifies all `Records` are implicitly `STATIC` and `FINAL`, and it seems consistent with the Reflection API.

## Remaining Issue

I'll look into it more the next days, but there is still some build problem regarding JDK 14 vs JDK 15 testing :thinking: When we use a build JDK 14 and then a test JDK 15 the `--enable-preview` seems to clash :disappointed: The result is something like

```
java.lang.UnsupportedClassVersionError: com/tngtech/archunit/core/importer/ClassFileImporterRecordsTest (class file version 58.65535) was compiled with preview features that are unsupported. This version of the Java Runtime only recognizes preview features for class file version 59.65535
```

Out of the box I don't know how to solve this, if you know, let me know :wink: Otherwise I'll look into it, if worse comes to worse we can only run the tests with the same JDK that compiled the classes :thinking: And then fix that with JDK 16 when it is not preview anymore.

codecholeric:
Note: I just saw that `ACC_STATIC` supposedly only applies to fields and methods, but somehow it is part of `access` when we enter `visitInnerClass` (and is correctly derived for nested classes from that according to all tests at least :thinking:). 
So I don't know exactly what the intention is there :thinking: But maybe I should adjust the comment then in `visitRecord` :thinking:

archunit/src/jdk14test/java/com/tngtech/archunit/core/importer/ClassFileImporterRecordsTest.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Test
+    public void imports_record_component_getters() throws Exception {
+        record RecordToImport(String component1, int component2) {
+        }
+
+        JavaClass javaClass = new ClassFileImporter().importClass(RecordToImport.class);
+        JavaMethod component1Method = javaClass.getMethod("component1");
+        assertThat(component1Method).isEquivalentTo(RecordToImport.class.getMethod("component1"));
```
hankem, line 57: Given that we use `getDeclaredConstructor` and `getDeclaredField` above: should we use `getDeclaredMethod` here as well?

rweisleder:
Thanks for reviewing! No need to apologize for taking so long. For a "one man show", that's absolutely okay. :smiley: I looked at your fixup commits. They all look reasonable to me.

About the remaining issue
> When we use a build JDK 14 and then a test JDK 15 the `--enable-preview` seems to clash

I found that it's not possible to run the same classes with different JDKs. :confused:
> When experimenting with preview features, you need a compiler and a JVM from the same major version that introduced that feature. For example, you can't use 13's compiler to compile code using 12's switch expressions.

(source: [Evolving Java With `--enable–preview` aka Preview Language Features](https://nipafx.dev/enable-preview-language-features/#same-version-for-feature-compiler-and-jvm))

The simple solution would be that we test records only with JDK16+ when it's no preview feature anymore. But therefore we have to wait one more week until JDK 16 gets released. The usage of ArchUnit with records as preview feature in JDK 14/15 is then not supported by us.

Another idea that comes to my mind is that we could have two types of test sources for each JDK. The first type is the one we already have (e.g. `src/jdk9test/java`) which is executed with the declared JDK version and all newer. As second type we could introduce a test source for preview features (e.g. `src/jdk14-preview-test/java`) which is compiled with and only executed with the declared JDK version. In the concrete example of records, this means that we would have three test sources
* `src/jdk14-preview-test/java` which needs to be compiled with and only runs with JDK 14
* `src/jdk15-preview-test/java` which needs to be compiled with and only runs with JDK 15
* `src/jdk16test/java` which runs with JDK 16+

I think that Gradle's "Toolchains for JVM projects" (see #477) might be helpful here. On the downside, we would have to triple the test code for records (or use obscure gradle configurations).

codecholeric (approved):

codecholeric:
Boom :boom: Thank you so much for all your support to make ArchUnit better :smiley: :tada: BTW. one week until JDK 16 is released I guess :joy: But it shouldn't be too bad to lift the tests to JDK 16 once and for all then and drop the preview flag :smiley:

PR #320, opened 2020-02-19
 support OnionArchitecture.ignoreDependency

### LINKED ISSUE #318: Onion Architecture main class
Hello,
I tried out the `onionArchitecture()` style on a test project.
My test code (Kotlin):
```
    fun onionTest(importedClasses: JavaClasses) {
        val root = "hexarch_test"
        val rule = onionArchitecture()
            .domainModels("$root.domain.model..")
            .domainServices("$root.domain.service..")
            .applicationServices("$root.application..")
            .adapter("persistence", "$root.adapter.persistence..")
        rule.check(importedClasses)
    }
```
I put a main class which bootstraps the application the base package.
Now archunit complains that this main class accesses adapter code (which is the case).

I tried to add an exception for that class using: `importedClasses.that(doNot(equivalentTo(Main::class.java)))` but it would still complain that the adapter packages are accessed.

Maybe this issue does not occur if I were using DI where the adapter code would not be accessed directly (but actually still are indirectly / at runtime).

What's a good way to solve this? Using a DI-framework should not be mandatory IMO.

hankem:
resolves #318

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,8 +4,9 @@ Contributions are very welcome. The following will provide some helpful guidelin
 
 ## How to build the project
 
-ArchUnit requires at least JDK 9 to build. The following is just an example input/output from a 
-Unix command line. Windows users should use `gradlew.bat` instead.
+ArchUnit requires at least JDK 11 to build.
+The following is just an example input/output from a Unix command line.
+Windows users should use `gradlew.bat` instead.
 
 ```
 $ cd /path/to/git/clone/of/ArchUnit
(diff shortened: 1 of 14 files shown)


codecholeric:
Thanks a lot for your support and welcome to the official ArchUnit contributors :wink: 
LGTM, I just made a tiny adjustment to the test and added an example, because I think it's good to document how to ignore violations. We did the same for `LayeredArchitectureTest`, I think it's nice if people look into the example `OnionArchitectureTest` and see how to ignore a dependency right away :smiley:

PR #326, opened 2020-02-29
Reduce number of details shown for cycle violations

codecholeric:
So far in highly coupled legacy code bases there was no way to reduce the amount of details when evaluating

```
slices().matching("com.myapp.root.(*)..").should().beFreeOfCycles()
```

This lead to the only indicator of "you really have a cycle problem" being that the test does not terminate or causes a `StackOverFlowError` because the amount of violations to print overloads the configured heap (assume 100 chars * 250 dependencies * 10 edges * 5000 cycles >= 1.25GB heap with an overly optimistic 1B per char).

This PR introduces default limits (max 100 cycles to detect, max 20 dependencies per edge to report) and makes these limits configurable via `archunit.properties` in case they don't match users' needs.


archunit/src/main/java/com/tngtech/archunit/library/dependencies/SliceCycleArchCondition.java
```diff
@@ -195,16 +219,25 @@ private String createDetails(Map<String, Edge<Slice, Dependency>> descriptionsTo
...
-            for (Dependency dependency : new TreeSet<>(edgeWithDescription.getValue().getAttachments())) {
+            int maxDependencies = cycleConfiguration.getMaxNumberOfDependenciesToShowPerEdge();
+            List<Dependency> allDependencies = edge.getAttachments();
+            List<Dependency> limitedDependencies = allDependencies.size() > maxDependencies
+                    ? allDependencies.subList(0, maxDependencies)
+                    : allDependencies;
+            for (Dependency dependency : limitedDependencies) {
                 result.add(dependency.getDescription());
             }
+            if (limitedDependencies.size() != allDependencies.size()) {
```
torfmaster: See here

archunit/src/main/java/com/tngtech/archunit/lang/ConditionEvents.java
```diff
@@ -36,12 +38,26 @@ public ConditionEvents() {
     }
 
     private final Multimap<Type, ConditionEvent> eventsByViolation = ArrayListMultimap.create();
+    private Optional<String> informationAboutNumberOfViolations = Optional.absent();
```
torfmaster: As far as I can judge `informationThatViolationsHaveBeenLimited` would be a better description.

archunit/src/main/java/com/tngtech/archunit/library/dependencies/CycleConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.tngtech.archunit.library.dependencies;
+
+import com.tngtech.archunit.ArchConfiguration;
+
+final class CycleConfiguration {
+    static final String MAX_NUMBER_OF_CYCLES_TO_DETECT_PROPERTY_NAME = "cycles.maxNumberToDetect";
+    private static final String MAX_NUMBER_OF_CYCLES_TO_DETECT_DEFAULT_VALUE = "100";
```
torfmaster: This is a rather low number as a default I think even 1000 will not hit any heap limits on current machines (or build agents).

archunit/src/main/java/com/tngtech/archunit/library/dependencies/CycleConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+    static final String MAX_NUMBER_OF_DEPENDENCIES_TO_SHOW_PER_EDGE_PROPERTY_NAME = "cycles.maxNumberOfDependenciesPerEdge";
+    private static final String MAX_NUMBER_OF_DEPENDENCIES_TO_SHOW_PER_EDGE_DEFAULT_VALUE = "20";
```
torfmaster: Same applies here.

archunit/src/test/java/com/tngtech/archunit/library/dependencies/EdgeTest.java
```diff
@@ -11,19 +14,21 @@
     @Test
     public void edges_are_equal_iff_from_and_to_are_equal() {
 
-        assertThat(new Edge<>(from, to)).isEqualTo(new Edge<>(from, to));
-        assertThat(new Edge<>(from, to)).isNotEqualTo(new Edge<>(new Object(), to));
-        assertThat(new Edge<>(from, to)).isNotEqualTo(new Edge<>(from, new Object()));
+        assertThat(newEdge(from, to)).isEqualTo(newEdge(from, to));
```
torfmaster: I find calling the constructor more expressive here.

archunit/src/test/java/com/tngtech/archunit/library/dependencies/GraphTest.java
```diff
@@ -128,24 +132,37 @@ public void complete_graph() {
...
-                new Edge<Integer, Integer>(0, 4),
-                new Edge<Integer, Integer>(1, 0),
-                new Edge<Integer, Integer>(1, 5),
-                new Edge<Integer, Integer>(2, 1),
-                new Edge<Integer, Integer>(3, 1),
-                new Edge<Integer, Integer>(3, 5),
-                new Edge<Integer, Integer>(4, 3),
-                new Edge<Integer, Integer>(5, 1),
-                new Edge<Integer, Integer>(5, 2)
+                newEdge(0, 4),
```
torfmaster: Here, using `newEdge` is fine. Maybe it should be moved to `GraphTest`

archunit/src/test/java/com/tngtech/archunit/library/dependencies/GraphTest.java
```diff
@@ -128,24 +132,37 @@ public void complete_graph() {
...
+                newEdge(3, 5),
+                newEdge(4, 3),
+                newEdge(5, 1),
+                newEdge(5, 2)
         ));
 
         assertThat(graph.findCycles()).isNotEmpty();
     }
 
+    // This test covers some edge cases, e.g. if too many nodes stay blocked
```
torfmaster: This edge case is already covered by `graph_which_causes_error_when_dependently_blocked_nodes_are_not_cleared_after_unblocking`

archunit/src/test/java/com/tngtech/archunit/library/dependencies/RealLifeGraph.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.Map;
+import java.util.Set;
+
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Multimap;
+
+import static com.tngtech.archunit.library.dependencies.EdgeTest.newEdge;
+
+public class RealLifeGraph {
```
torfmaster: Maybe the edge cases covered by this graph should be described.

docs/userguide/008_The_Library_API.adoc
```diff
@@ -176,6 +176,24 @@ SliceAssignment legacyPackageStructure = new SliceAssignment() {
...
+They can be configured via `archunit.properties` (compare <<Advanced Configuration>>).
+
+[source,options="nowrap"]
+.archunit.properties
+----
+# this will limit the maximum number of cycles to detect and thus required CPU and heap
+# default is 100
+cycles.maxNumberToDetect=50
+
+# this will limit the maximum number of dependencies to report per cycle edge
```
torfmaster: A note that this affects only the output not the searching (i.e. edges will be respected but not reported) would be good to be prevent the user from increasing this number without need.

archunit/src/test/java/com/tngtech/archunit/library/dependencies/testexamples/cyclewithunbalanceddependencies/CycleWithUnbalancedDependenciesRoot.java
```diff
@@ -0,0 +1,4 @@
+package com.tngtech.archunit.library.dependencies.testexamples.cyclewithunbalanceddependencies;
+
```
torfmaster: A brief note about the need of this class (making the tests refactoring safe) would be helpful.

archunit/src/test/java/com/tngtech/archunit/library/dependencies/SliceRuleTest.java
```diff
@@ -0,0 +1,215 @@
...
+    /**
+     * To calculate the number of expected cycles check the following (we denote binomial coefficients "n over k" as (n k) ):
+     * For any k-cycle (cycle of k nodes within a graph of n nodes) we can denote
+     * n1 -> n2 -> n3 -> ... -> n[k-1] -> n[k] -> n1
+     * Fix n1 and we have (k-1)! possibilities to order the other k-1 nodes.
+     * Now we have (n k) possibilities to choose k nodes to form a k-cycle and for each choice (k-1)! permutations.
+     * Thus we retrieve the sum of all cycles as
+     * SUM(k=2, n) (n k) * (k-1)! = SUM(k=2, n) n! / ((n-k)! * k)
+     */
+    static int getNumberOfCyclesInCompleteGraph(int numberOfNodes) {
```
torfmaster: Is the exact value really needed here?

archunit/src/test/java/com/tngtech/archunit/library/dependencies/SliceRulePerformanceTest.java
```diff
@@ -27,31 +34,12 @@ public void searching_for_cycles_should_terminate_reasonably_fast_for_complete_g
...
-     * SUM(k=2, n) (n k) * (k-1)! = SUM(k=2, n) n! / ((n-k)! * k)
-     */
-    private int getNumberOfCyclesInCompleteGraph(int numberOfNodes) {
-        int numberOfCycles = 0;
-        for (int k = 2; k <= numberOfNodes; k++) {
-            numberOfCycles += factorial(numberOfNodes) / (factorial(numberOfNodes - k) * k);
-        }
-        return numberOfCycles;
+        int numberOfDetectedCycles = countCyclesInMessage(violations);
+        assertThat(numberOfDetectedCycles).as("number of cycles detected").isEqualTo(expectedNumberOfCycles);
```
torfmaster: Should we really assert that in this test category? I would expect the assertion on the number of cycles in the unit test `GraphTest` if at all.

archunit/src/main/java/com/tngtech/archunit/library/dependencies/CycleConfiguration.java
codecholeric: Yes, but what is the info if your Unit test fails and reports 1000 cycles? This was less about "is it doable" as more "is it useful test output". And I think by default to see 100 cycles is way enough to get a feeling for the dimension :wink: 
In particular because of the explosion cycles * edges * violations per edge. Would you really want to read 1000 cycles x 10 edges x 100 dependencies = 1000000 lines of test failure message? Besides that IntelliJ cannot print this anymore as I found out :wink:

archunit/src/main/java/com/tngtech/archunit/lang/ConditionEvents.java
codecholeric: But it is more generic, right? You don't have to specify "the information has been limited" you could also say "the heuristically determined number of violations reported does not match my custom violations, I'll count and specify the number of violations myself". I'm just not using that within ArchUnit itself, but this is public API that users of ArchUnit could use themselves...

archunit/src/main/java/com/tngtech/archunit/library/dependencies/SliceCycleArchCondition.java
```diff
@@ -195,16 +219,25 @@ private String createDetails(Map<String, Edge<Slice, Dependency>> descriptionsTo
...
             return Joiner.on(System.lineSeparator()).join(details);
         }
 
-        private List<String> dependenciesDescription(Map.Entry<String, Edge<Slice, Dependency>> edgeWithDescription) {
+        private List<String> dependenciesDescription(Edge<Slice, Dependency> edge) {
             List<String> result = new ArrayList<>();
-            for (Dependency dependency : new TreeSet<>(edgeWithDescription.getValue().getAttachments())) {
+            int maxDependencies = cycleConfiguration.getMaxNumberOfDependenciesToShowPerEdge();
+            List<Dependency> allDependencies = edge.getAttachments();
+            List<Dependency> limitedDependencies = allDependencies.size() > maxDependencies
```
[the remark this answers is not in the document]
codecholeric: Right, gonna extract and reuse that, then it's gonna be slightly quicker to read...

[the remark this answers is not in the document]
codecholeric: done :+1:

archunit/src/main/java/com/tngtech/archunit/library/dependencies/CycleConfiguration.java
codecholeric: Same argument as above applies here :wink: Do I really want to read so many dependencies per edge? I would guess normally there is one direction that is intended, where there are a lot of dependencies, and then there is this wrong edge that creates the cycle, which maybe has only a couple of dependencies (yes, in legacy projects that's probably too optimistic :wink:)
I mean, this is only a default, if anyone doesn't like it it can be reconfigured. But remember that this is going to be the setting **for every edge**, so it quickly explodes.

archunit/src/test/java/com/tngtech/archunit/library/dependencies/EdgeTest.java
codecholeric: Hmm, I'm ambiguous here. Because then you now have to specify `new Edge<>(from, to, Collections.emptySet())` or similar for every edge. Seems to also obscure a little what is the essence of the test?

codecholeric: But maybe you are right, in particular in the second part of the test, where `newEdge` is compared to `new Edge(..)`. Okay, convinced, gonna inline it there and move the method :wink:

codecholeric: Done

archunit/src/test/java/com/tngtech/archunit/library/dependencies/GraphTest.java
codecholeric: done

archunit/src/test/java/com/tngtech/archunit/library/dependencies/GraphTest.java
codecholeric: Hmm, to be honest, when I was done I tested with `org.hibernate` and it crashed because I had made some wrong refactoring with respect to unblocking. Yet all existing tests were green. Since I felt like I had already invested much time, I decided to not dig deeper to reduce the problem to a minimal example, but instead just use the Hibernate graph that caused the problem.
So I don't exactly know the difference to the other test, I just know that the way I broke unblocking did not break the existing test :frowning:
I'm still not sure if it's worth it to invest the time though, but yes, maybe the comment should then be adjusted? Or do you think we should dig deeper?

codecholeric: The error here was that the whole node was not unblocked though, not that the dependently blocked nodes weren't cleared correctly. So it is also not the same case...

codecholeric: So maybe the comment is fine then, because if you carefully read it, the one test says "when dependently blocked nodes are not cleared" while the comment says "if nodes stay blocked". But maybe it could be written clearer?

archunit/src/test/java/com/tngtech/archunit/library/dependencies/RealLifeGraph.java
codecholeric: If I would know the exact edge case, then yes :wink: But unfortunately this graph is super complex and I don't know exactly why all our other graphs did not uncover that unblocking issue I've introduced. I just knew it would probably cost me a couple of hours to find out exactly the constellation causing the issue and I did not want to spend that time.
Or do you have an idea what the problem could be?

docs/userguide/008_The_Library_API.adoc
codecholeric: Ah, yes, I didn't think about that :+1: I'll write this more clear

codecholeric: done

archunit/src/test/java/com/tngtech/archunit/library/dependencies/SliceRuleTest.java
codecholeric: Hey, you just wrote `assertThat(numberOfCycles).isEqualTo(1442525)` :rofl:
Yes, I think I would want to assert it once blackbox against the rule text, just to assert that really all cycles make it into the message. You can argue about the performance test :wink:

archunit/src/test/java/com/tngtech/archunit/library/dependencies/SliceRulePerformanceTest.java
codecholeric: As mentioned, can argue about that. I mean the point is that the test terminates in time, I know. But a test without assertion seems confusing, too, somehow.
We could write a comment instead, like "if we reach this point in reasonable time, we consider the performance test passed"? :wink: 
Or could simply assert `violations != null`, but that would also be weird...

archunit/src/test/java/com/tngtech/archunit/library/dependencies/testexamples/cyclewithunbalanceddependencies/CycleWithUnbalancedDependenciesRoot.java
codecholeric: Yeah, you're right, I've added it analogously to the others... :+1:

archunit/src/main/java/com/tngtech/archunit/lang/ConditionEvents.java
torfmaster: I agree.

archunit/src/main/java/com/tngtech/archunit/library/dependencies/CycleConfiguration.java
torfmaster: I agree. I didn't think human beeings as consumers of the test results.

archunit/src/test/java/com/tngtech/archunit/library/dependencies/SliceRulePerformanceTest.java
torfmaster: I see the point of the test without assertion. We can leave that here (as the test hopefully never breaks because of deviation of numbers here).

archunit/src/main/java/com/tngtech/archunit/library/dependencies/CycleConfiguration.java
torfmaster: As below I didn't think of human beeings as consumers of test results in my request.

archunit/src/test/java/com/tngtech/archunit/library/dependencies/GraphTest.java
torfmaster: No, I just didn't know the background and didn't get the lack of the word "dependently".

archunit/src/test/java/com/tngtech/archunit/library/dependencies/SliceRuleTest.java
torfmaster: That makes to me (see also below)

archunit/src/test/java/com/tngtech/archunit/library/dependencies/RealLifeGraph.java
torfmaster: I see no obvious reason why this should fail (but I don't know your erroneous refactoring either). I would propose leaving this as it is. Maybe you _could_ describe the error you introduced during refactoring, but this makes only sense, if this can be explained in two or three sentences.

PR #328, opened 2020-03-07
Reuse supplied import options in ClassFileImporter.importClasspath()

### LINKED ISSUE #296: ClassFileImporter.importClasspath() ignores .withImportOption
The [user guide](https://www.archunit.org/userguide/html/000_Index.html#_import) contains the following snippet:

```java
new ClassFileImporter()
    .withImportOption(ImportOption.Predefined.DO_NOT_INCLUDE_JARS)
    .withImportOption(ImportOption.Predefined.DO_NOT_INCLUDE_TESTS)
    .importClasspath();
```

Unfortunately `ClassFileImporter.importClasspath()` seems to override the import options and includes test classes:

```java
importClasspath(new ImportOptions().with(ImportOption.Predefined.DO_NOT_INCLUDE_ARCHIVES));
```
The following code works as expected:

```java
new ClassFileImporter().importClasspath(new ImportOptions()
    .with(ImportOption.Predefined.DO_NOT_INCLUDE_JARS)
    .with(ImportOption.Predefined.DO_NOT_INCLUDE_TESTS)
);
```

rweisleder:
Until now any `ImportOption` supplied via `classFileImporter.withImportOptions(importOptions)` was disregarded, if `importClasspath()` was used (that method just used new `ImportOptions` with some default).
Now any previously added `ImportOptions` will be merged with the defaults resulting in a more consistent and less surprising development experience.

Fixes #296

### DIFF
--- a/archunit-junit/build.gradle
+++ b/archunit-junit/build.gradle
@@ -15,10 +15,6 @@ sourceSets {
     }
 }
 
-spotbugs {
-    sourceSets += [project.sourceSets.api]
-}
-
 dependencies {
     compileOnly sourceSets.api.output
     testCompile sourceSets.api.output
--- a/archunit/build.gradle
+++ b/archunit/build.gradle
@@ -89,6 +89,8 @@ task jdk9Test(type: Test) {
 
 [jar, test]*.dependsOn compileJdk9mainJava
 
+spotbugsJdk9test.enabled = false
+
 idea {
     module {
         sourceDirs += jdk9MainDirs.collect { file(it) }
--- a/archunit/src/main/java/com/tngtech/archunit/core/importer/ClassFileImporter.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/importer/ClassFileImporter.java
@@ -248,7 +248,7 @@ public final class ClassFileImporter {
      */
     @PublicAPI(usage = ACCESS)
     public JavaClasses importClasspath() {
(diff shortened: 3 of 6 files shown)


codecholeric:
Thanks a lot! :smiley: I adjusted the test name in your commit a little to better reflect the point of it (at least how I see it). Also I think the one test you adjusted actually should stay as it was, because the whole point was to get classes from the JDK, JARs and the project together. But unfortunately when I introduced the test for packages, I sloppily renamed that method to `importJavaBase()` (because that was what I needed for the package test), even though it actually imports more than that :disappointed:
So actually the name of that method was wrong, not what the test tested. So I inlined the old version of the import into the test method and kept the test as it was.
But I agree to your other comment that whole role of `importClasspath(..)` behaving differently than all the other import methods is rather unfortunate :frowning: We should fix that, if only there wouldn't be this API compatibility issue :wink:

codecholeric:
Hmm, I think Travis CI is broken at the moment :frowning: First I thought our Spotbugs version is too old (because it complained about some beta-version of a dependency missing?? :open_mouth: Does not raise my trust into the dependency management of Spotbugs...) So I've upgraded Spotbugs to the current version, but now it fails about some other dependency supposedly missing in Maven Central. If I copy the URL that supposedly cannot be reached, I can access it, so I think there is something wrong in Travis CI.
Gonna wait a little and then try to rerun again...

Issue #337, opened 2020-04-07
Arch Unit configuration with lombok [Error]


JaimLus:
Hi everyone,

I need help setting up the arch unit with lombok.

I am testing that my domain models are not annotated with @Data from lombok library, but I run the test and I get this error:

![image](https://user-images.githubusercontent.com/12102076/78619847-0b3edd80-7844-11ea-8774-261e899aa98b.png)

![image](https://user-images.githubusercontent.com/12102076/78619866-185bcc80-7844-11ea-9e27-e25a8097a38d.png)

![image](https://user-images.githubusercontent.com/12102076/78619882-24478e80-7844-11ea-822a-e01a787b998a.png)

**What recommendation do you suggest?**

This is test:
![image](https://user-images.githubusercontent.com/12102076/78619978-61138580-7844-11ea-837c-335f3cdfd4e2.png)

hankem:
As ArchUnit is (currently) built solely on byte code, there is unfortunately no way to check whether a class is annotated with an annoation with `@Retention(`[`SOURCE`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/annotation/RetentionPolicy.html#SOURCE)`)`, because – just as ArchUnit's consistency check states – _that information is gone after compile_.

So from ArchUnit's (current) point of view, it doesn't matter how your bytecode is generated.

I understand that you may still want to avoid the dependency on lombok; so maybe there is another signature in the generated bytecode that you can look for?
(I don't have a lombok example project at hand, but if you post the result of 
```
javap -c -p -v AMinimalClassUsingALombokDataAnnotation.class
```
we can have a look at the bytecode together.)

JaimLus:
`Classfile /D:/Documentos/Personales/Spring_boot/Git/arch-unit/build/classes/java/main/com/jaimlus/archunit/demo/application/dto/OneDto.class
  Last modified 7/04/2020; size 877 bytes
  MD5 checksum 41d05fe7f9ce43002e62cd155633f09a
  Compiled from "OneDto.java"
public class com.jaimlus.archunit.demo.application.dto.OneDto
  minor version: 0
  major version: 52
  flags: ACC_PUBLIC, ACC_SUPER
Constant pool:
   #1 = Methodref          #5.#30         // java/lang/Object."<init>":()V
   #2 = Class              #31            // com/jaimlus/archunit/demo/application/dto/OneDto
[...]

Issue #75, opened 2018-06-01
Expose line number of the definition of a JavaMember


rweisleder:
For a custom ArchCondition that checks the definition of a JavaClass/JavaMethod/JavaField it would be useful to append the declaration line to the error message as clickable link.

Example:
```java
private ArchCondition<JavaMethod> haveNameStartingWithLowerCase() {
    return new ArchCondition<JavaMethod>("have name starting with lower case") {
        @Override
        public void check(JavaMethod item, ConditionEvents events) {
            boolean satisfied = Character.isLowerCase(item.getName().charAt(0));
            String message = item.getFullName() + " " + Formatters.formatLocation(item.getOwner(), item.getDeclarationLineNumber());
            events.add(new SimpleConditionEvent(item, satisfied, message));
        }
    };
}
```
would print
```
java.lang.AssertionError: Architecture Violation [Priority: MEDIUM] - Rule 'methods should have name starting with lower case' was violated (1 times):
demo.GreetEndpoint.greet(java.lang.String) (GreetEndpoint.java:23)
```

Relates to #38

codecholeric:
I know, this issue came up in the past. Surprisingly, there doesn't seem to be a clean way to determine the line number from Java bytecode (I would already have added it otherwise).
Every solution I found (I looked at Javassist, etc.) at best had some workaround.
The problem is, that there is no trace of the line number of a member within the bytecode as far as I know, so the only way is, to look at the first statement of a method, and do something shady like "subtract 1" to get to the declaration (for fields there is not even such a workaround AFAIK)
I'm open to suggestions though, if someone knows a solution to this!! But I think it just is an intrinsic limitation of Java bytecode.
Also compare: https://stackoverflow.com/questions/49986872/how-to-determine-line-number-for-the-method-with-java-asm

bedla:
My poor-man's solution is
```
        final int lineNumber = Stream.of(javaCodeUnit.getMethodCallsFromSelf(),
                javaCodeUnit.getConstructorCallsFromSelf(),
                javaCodeUnit.getFieldAccesses())
                .flatMap(Collection::stream)
                .mapToInt(JavaAccess::getLineNumber)
                .min()
                .orElse(0);
```

codecholeric:
Yeah, for methods that would work, you could subtract 1, hope that nobody includes a comment or empty line, and it would match the method declaration in most cases :wink: 
[...]

PR #344, opened 2020-04-10
Use first line number from LineNumberTable for JavaCodeUnit's sourceCodeLocation

hankem:
So far, the `sourceCodeLocation` of `JavaMember`s does not contain a line number, as it cannot reliably be inferred from the byte code (cf. #75).

However, if the class file contains a [`LineNumberTable`](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.7.12) for a method, using the smallest line number from the `LineNumberTable` for the member's `sourceCodeLocation` – even if this probably corresponds to the first executable statement and not to the method header definition – seems to be more useful to me than using the fallback line number `0` in any case.

So for example, I would infer the line number `10` from the following byte code:
```
  void methodWithBodyStartingInLine10();
    descriptor: ()V
    flags: (0x0000)
    Code:
      stack=2, locals=1, args_size=1
         0: getstatic     #2                  // Field java/lang/System.out:Ljava/io/PrintStream;
         3: bipush        10
         5: invokevirtual #3                  // Method java/io/PrintStream.println:(I)V
         8: getstatic     #2                  // Field java/lang/System.out:Ljava/io/PrintStream;
        11: bipush        11
        13: invokevirtual #3                  // Method java/io/PrintStream.println:(I)V
        16: getstatic     #2                  // Field java/lang/System.out:Ljava/io/PrintStream;
        19: bipush        12
        21: invokevirtual #3                  // Method java/io/PrintStream.println:(I)V
        24: return
      LineNumberTable:
        line 10: 0
        line 11: 8
        line 12: 16
        line 13: 24
```
(My example's method header is actually defined in line 9, but I prefer 10 as opposed to 0... 😉)

Note that I do even get a line number for an empty method from the following byte code:
```
  void emptyMethodDefinedInLine15();
    descriptor: ()V
    flags: (0x0000)
    Code:
      stack=0, locals=1, args_size=1
         0: return
      LineNumberTable:
        line 15: 0
```

Even if line numbers inferred in this way do not exactly point to the method header definition, they can be used to  compare different methods (i.e., the ordering should be correct). This would allow for new rules like, for example (irrespective of whether I'd personally want to have them as arch rules or not):
* Public methods are defined _before_ private methods.
* `equals`, `hashCode` and `toString` are not generated by a framework (or a developer... 🙈) in the _same_ line (cf. #337)

With this pull request, line numbers are recorded for `JavaCodeUnit`s (`JavaMethod`s, `JavaConstructor`s, `JavaStaticInitializer`s) if the class file has a `LineNumberTable`.
(The reported line number `0` for `JavaField`s is unchanged, as it cannot be inferred from byte code.)

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/importer/DomainBuilders.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/importer/DomainBuilders.java
@@ -118,6 +118,7 @@ public final class DomainBuilders {
         private Set<JavaModifier> modifiers;
         private JavaClass owner;
         private ClassesByTypeName importedClasses;
+        private int firstLineNumber;
 
         private JavaMemberBuilder() {
         }
@@ -140,6 +141,10 @@ public final class DomainBuilders {
         SELF withModifiers(Set<JavaModifier> modifiers) {
             this.modifiers = modifiers;
             return self();
+        }
+
+        void recordLineNumber(int lineNumber) {
+            this.firstLineNumber = this.firstLineNumber == 0 ? lineNumber : Math.min(this.firstLineNumber, lineNumber);
         }
 
         @SuppressWarnings("unchecked")
@@ -176,6 +181,10 @@ public final class DomainBuilders {
 
         public JavaClass getOwner() {
             return owner;
+        }
+
+        public int getFirstLineNumber() {
+            return firstLineNumber;
         }
 
         @Override
--- a/archunit-integration-test/src/test/java/com/tngtech/archunit/integration/ExamplesIntegrationTest.java
+++ b/archunit-integration-test/src/test/java/com/tngtech/archunit/integration/ExamplesIntegrationTest.java
@@ -911,8 +911,8 @@ class ExamplesIntegrationTest {
 
                 .ofRule("no code units that are declared in classes that reside in a package '..persistence..' "
                         + "should be annotated with @" + Secured.class.getSimpleName())
-                .by(ExpectedConstructor.of(SomeJpa.class).beingAnnotatedWith(Secured.class))
-                .by(ExpectedMethod.of(OtherJpa.class, "getEntityManager").beingAnnotatedWith(Secured.class))
+                .by(ExpectedConstructor.of(SomeJpa.class).inLine(15).beingAnnotatedWith(Secured.class))
+                .by(ExpectedMethod.of(OtherJpa.class, "getEntityManager").inLine(31).beingAnnotatedWith(Secured.class))
 
                 .toDynamicTests();
     }
--- a/archunit-integration-test/src/test/java/com/tngtech/archunit/testutils/ExpectedConstructor.java
+++ b/archunit-integration-test/src/test/java/com/tngtech/archunit/testutils/ExpectedConstructor.java
@@ -15,17 +15,23 @@ public class ExpectedConstructor {
     public static class Creator {
         private final Class<?> clazz;
         private final Class<?>[] params;
+        private int lineNumber;
 
         private Creator(Class<?> clazz, Class<?>[] params) {
             this.clazz = clazz;
             this.params = params;
         }
 
+        public  Creator inLine(int lineNumber) {
+            this.lineNumber = lineNumber;
+            return this;
+        }
+
         public ExpectedMessage beingAnnotatedWith(Class<? extends Annotation> annotationType) {
-            return new ExpectedMessage(String.format("Constructor <%s> is annotated with @%s in (%s.java:0)",
+            return new ExpectedMessage(String.format("Constructor <%s> is annotated with @%s in (%s.java:%d)",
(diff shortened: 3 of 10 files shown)


codecholeric:
Nice, thanks a lot :smiley: 
I still feel a little uneasy about the "heuristic over the thumb" but I agree that it is probably more useful than a constant 0 :wink: 
It's particularly nice that you get this number also for void-methods with empty body :smile:
This makes it at least pretty consistent between methods and constructors I guess.
Too bad there is no way to determine it for fields :disappointed: 
But regarding the heuristic, is there any way to distinguish this empty case when the line number is actually exact (at least as far as I could tell?) and the case where it is the first line of the body? Because if yes I would wonder if a best guess `- 1` wouldn't be even better in that case, even though it might sometimes still be a little off? (e.g. if there is a comment in the first line)

hankem:
I don't think there's a way to get what you'd consider _exact_... Consider this example:
```
17    void emptyMethodEndingInLine19() {
18
19    }
```
The auto-generated `return` (the only statement in the byte code) gets placed in line 19, presumably at the `}`.

If we had instead used 
```
17    void emptyMethodEndingInLine19() {
18        return;
19    }
```
it would of course be in line 18... 🤷‍♂️

I would anyways refrain from playing any "line number – 1" trick, but just stick to the data that is actually there (to me, the only canonical options seem to be either the first or the last line number of the statements in the byte code), because
* we don't know, as you already mention, whether "the first line of a statement – 1" really hits the method header declaration, due to, e.g.,
  * the method body starting with a comment (which can be arbitrarily long)
  * the method body starting with one or several empty lines 
  * a code style demanding a method such as
```
    public void visit(int version, int access, String name, String signature, String superName, String[] interfaces) {
```
to be written as
```
    public void visit(
            int version,
            int access,
            String name,
            String signature,
            String superName,
            String[] interfaces
    ) {
```
* simple, well defined behavior can IMO more easily be explained/understood
(I'm sure that people will ask what the line number means...)
* we don't pretend that we could solve a problem that we know we cannot. 😉

codecholeric:
Okay, you have convinced me :wink: I agree, we can't solve it perfectly, and an educated guess close to the declaration is still a lot nicer than line number 0 which will definitely make you search for the declaration (in any non-trivial class). And yes, keeping it simple and directly providing the info from the bytecode seems to be the way to go. Too bad we have no estimate whatsoever for fields :frowning_face:

archunit/src/main/java/com/tngtech/archunit/core/importer/DomainBuilders.java
```diff
@@ -142,6 +143,11 @@ SELF withModifiers(Set<JavaModifier> modifiers) {
             return self();
         }
 
+        SELF withLineNumber(int lineNumber) {
```
codecholeric, line 146: Should we name this method in a way that reflects that it does more than just stupidly adding properties like other `with`-methods? Because it does not seem to be the way of least surprise to me, that I might set `withLineNumber(9)`, but afterwards `lineNumber == 7` because I actually ignored that element? I don't know what the best name would be, maybe something like `recordMethodBodyLineNumber(..)`? Or `addMethodBodyLineNumber(..)`? (but that one suggests to me I would keep them all somehow) Or `registerMBLN`? Or `storeMinimumMethod{Body}LineNumber(..)`?

hankem: Hehehe, I actually started with the `JavaMemberBuilder`'s field called `lineNumber` and had the following JavaDoc for `withLineNumber`:
```
/**
 * records a visited line number.
 * The actual line number of the JavaMember being build will be the smallest lineNumber != 0 ever visited.
 */
```
But then I thought that `withLineNumber` might not need that explanation if the field is called differently, i.e. `firstLineNumber`. 

If we need to keep all line numbers for some reason (e.g. add them to a `Set<Integer> lineNumbers`), we'd only need to change the implementation, but could keep the interface of this method.

I wouldn't include `Method`(`Body`) in the name, as the builder can be used for all members and is actually used for all code units. (It's not the builder's fault that we don't use call it for fields... 😜)

**TL;DR**: I think `recordLineNumber` would be the best alternative, but I'd still think that `withLineNumber` fits better for a builder.

hankem:
>  I feel we should squash these two commits.

Fair enough. (I just kept them separate as I wasn't sure whether we should make the existing failure report tests somehow ignore the line number of methods [which actually didn't work out easily] rather than encode the line number in the tests... 😉)

archunit/src/main/java/com/tngtech/archunit/core/importer/DomainBuilders.java
codecholeric: I'll buy `recordLineNumber` then :wink: For me that "`withLineNumber` fits for a builder" makes it even worse :rofl: Exactly because it looks like a typical builder method that you can quickly skim over but it actually does not behave like it. I'll rename that and merge it!

PR #345, opened 2020-04-14
Upgrade Gradle to 6.3

codecholeric:
This upgrades Gradle to the current version. One notable change is the migration to use the Java-Library-Plugin instead of the Java-Plugin.
Unfortunately the combination of IntelliJ 2020.1 and Gradle 6 does not work as well as it used to anymore. In particular IntelliJ cannot natively build this project anymore and without some explicit hack could also not run the tests with the IntelliJ test runner anymore. Thanks to this one hack (declaring a redundant dependency), it is now possible again to run tests with the IntelliJ test runner and only delegate the build to Gradle.
Altogether this seemed to be a good enough compromise to us, since with the current Gradle version we can now compile Java 14.


codecholeric:
@aaschmid Looks good to me and ran without problems in test projects where I tried the deployed SNAPSHOT artifacts from Maven Central :smiley: Thanks a lot for all the preparation so it was pretty smooth sailing for me (just some minor hacks :wink:)

PR #357, opened 2020-05-06
Normalize backslashes of Windows paths

### LINKED ISSUE #356: Import of class files is broken on Windows
I noticed that some tests fail on my Windows with the current master. For example, the test https://github.com/TNG/ArchUnit/blob/0c23de0d5bdf943df3c79a7026f877061eb4b709/archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterTest.java#L2084-L2089 fails with
```
Expecting:
  <[]>
to contain exactly (and in same order):
  <["com.tngtech.archunit.core.importer.testexamples.simpleimport.ClassToImportOne",
    "com.tngtech.archunit.core.importer.testexamples.simpleimport.ClassToImportTwo"]>
but could not find the following elements:
  <["com.tngtech.archunit.core.importer.testexamples.simpleimport.ClassToImportOne",
    "com.tngtech.archunit.core.importer.testexamples.simpleimport.ClassToImportTwo"]>
```

This error came in after merging #347 into master (commit b7bf88a4d21bd9ac73175ac81b51b60ad9077b84). The state before (commit 06f94b208d18f0f88e37bfd833205151859d900f) looks good.

rweisleder:
Now that we uniformly search all classpath resources by iterating the classpath URLs and matching them against the specified classes to import (where we used `ClassLoader.getResources(..)` in the past), we also need to normalize URIs by replacing Windows path separators `\` with Unix path separators `/`. Otherwise the mechanism does not work for file URI classpath resources like `file:/C:\foo\bar\..`.

Fixes #356

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/importer/NormalizedResourceName.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/importer/NormalizedResourceName.java
@@ -21,8 +21,13 @@ class NormalizedResourceName {
     private final String resourceName;
 
     private NormalizedResourceName(String resourceName) {
-        resourceName = resourceName.replaceAll("^/*", "").replaceAll("/*$", "");
-        this.resourceName = resourceName;
+        this.resourceName = resourceName
+                // normalize Windows backslashes
+                .replace('\\', '/')
+                // remove leading slashes
+                .replaceAll("^/*", "")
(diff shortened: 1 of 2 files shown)


codecholeric:
Why am I not surprised it is something like this :rofl: Thanks for the quick fix, I'll look into it later!

PR #362, opened 2020-05-22
Use Github Actions for build

spanierm:
Use Github actions to check whether the build runs on all OS.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,28 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        os:
+          - ubuntu-latest
+          - macos-latest
+          - windows-latest
+        java_version:
+          - 11
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java_version }}
+      - name: Build
+        uses: eskatos/gradle-command-action@v1
+        with:
+          arguments: build -PallTests


codecholeric:
Thanks a lot for the PR :smiley: 
I just don't think that it will be a "build matrix" with respect to the JDK version :thinking: Because soon the build will only work with JDK 14 anyway I think.
We can probably hard code the build JDK version to the newest supported one, but then we need to check out a whole palette of JDKs to `runMavenIntegrationTest` against. Do you think that is possible?

.github/workflows/build.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java_version }}
+      - name: Build
+        run: ./gradlew build
```
codecholeric, line 26: the build command should probably be `build -PallTests` to also run the slower tests...

.github/workflows/build.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java_version }}
```
codecholeric, line 24: Can we just set up all the Java-Versions from 7-14 here and then choose a specific one for the Gradle build? E.g. `JAVA_HOME=${howToFindOut?} gradlew build`?

.github/workflows/build.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Build
```
codecholeric, line 25: Should we use something like `eskatos/gradle-command-action@v1` here? Because on Windows the command `./gradlew` should fail, shouldn't it?

spanierm: It worked with `./gradlew build` as well though I am not sure what kind of magic helped us here :see_no_evil:

spanierm:
I though about hard coding the JDK version as well. The only reason to have several versions is in case we update the JDK version. During the migration/testing phase, this might help.

.github/workflows/build.yml
spanierm: Hmmm, what benefit do you expect from this solution? I think it is easier to let the matrix build take care of these env vars and use it accordingly rather than using it as a kind of "option provider that requires manual interaction".

spanierm: Ah, I see. You refer to the `runMavenIntegrationTest` task. Here, I thought about publishing each successful build into the Github private registry and then use the real artifact for tests with different JDKs. But I am still investigating whether this is easily possible as it might require some refactoring of the test run.

codecholeric: Ah, okay, yes, this makes sense :+1: It will be the real `publish` task then to publish to the GitHub registry? That would give an additional level of testing which would be really neat :smiley:

.github/workflows/build.yml
codecholeric: That's interesting, maybe this is some Microsoft magic to help stupid Linux users, now that GitHub belongs to MS :rofl:

codecholeric:
Yeah, I don't mind keeping it as a matrix, might be nice to be able to run the build with two different JDK versions in an easy way :+1:

codecholeric:
So would the build job then get an additional publish step? And would the `mavenIntegrationTest` then be a separate job defined in the same `yml`?

codecholeric:
I'm gonna go ahead and merge this, we can extend it to run the Maven integration test on various platforms and with various JDK versions later on. Meanwhile to build across different platforms already is a big improvement. Thanks a lot for your support!! :smile:

PR #368, opened 2020-05-28
Run integration tests for all JDKs in Github action for CI

spanierm:
Simply changing the JDK (and thus `JAVA_HOME`) using the `actions/setup-java@v1`does not work for Java 7as Gradle requires at least Java 8. Thus we now add a adedicated env var `JAVA_HOME_MAVEN` to override the JDK in the Maven integration tests. Unfortunately, due to the need of env vars and the different handling in Unix and Windows, we needed to duplicate the build logic in two different jobs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,57 +7,21 @@ on:
       - release-*
   pull_request:
 
+env:
+  build_java_version: 11
+
 jobs:
-  build-unix:
+  integration-test:
     strategy:
       matrix:
         os:
           - ubuntu-latest
           - macos-latest
-        java_version:
-          -  7
-          -  8
-          -  9
-          - 10
-          - 11
-          - 12
-          - 13
-          - 14
-    runs-on: ${{ matrix.os }}
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


spanierm:
Ok, still not running the correct JDKs. Need to debug further... It worked some time ago :(

spanierm:
Now it works with Gradle properties as you suggested, @codecholeric. Thanks for your hints about passing `java<version>Home` as property :)

codecholeric:
Nice, thanks a lot for your support!! I've had my share of trouble with GH actions :disappointed: I'll try to unify this with #367 to get a somewhat consistent solution.
There are just these couple of WTFs that make it hard :wink: Like the `setup-java` action setting environment variables containing `.` in their identifier that can then be read with `env`, but not with `System.env` within Gradle :disappointed: I didn't think of the `$env` version though, so maybe I can simply use that :smiley:

codecholeric:
Thanks again! I unified it a little with the lessons I learned from #367. That way we don't need the separate configurations for Windows and Unix anymore :smiley: I also took the liberty to squash all your commits, since it seemed to be a lot of experimentation / adding debug output, etc. :wink:

PR #376, opened 2020-06-09
Prevent excessive logging when supplying system properties

### LINKED ISSUE #375: A ton of log lines generated when specifying system properties during test run
We use ArchUnit extensively and we're loving it. Recently after we added additional tests some of the runs during our CI started failing due to excess logging. 

```
14:27:41.711 [INFO ] com.tngtech.archunit.ArchConfiguration.getMergedProperties() @315 - Merging properties: The following properties have been overwritten by system properties: {freeze.store.default.allowStoreUpdate=false, freeze.store.default.allowStoreCreation=false}
 14:27:41.712 [INFO ] com.tngtec
 Job's log exceeded limit of 4194304 bytes.
```

The `The following properties have been overwritten by system properties: ` is repeatedly logged which seems to be due to the code in https://github.com/TNG/ArchUnit/blob/master/archunit/src/main/java/com/tngtech/archunit/ArchConfiguration.java#L315

```
Properties getMergedProperties() {
            Properties result = createProperties(this.properties);
            Properties overwritten = getSubProperties("archunit", System.getProperties());
            if (!overwritten.isEmpty()) {
                LOG.info("Merging properties: The following properties have been overwritten by system properties: {}", overwritten);
            }
            result.putAll(overwritten);
            return result;
        }
```

Running maven/surefire with debugging enabled and specifying some system properties as described in https://www.archunit.org/userguide/html/000_Index.html#_overriding_configuration seems to hit the log line repeatedly when using a breakpoint on line 315 (as referenced in the code file above)

 ```
mvn -Dmaven.surefire.debug -Darchunit.freeze.store.default.allowStoreUpdate=false -Darchunit.freeze.store.default.allowStoreCreation=false test
```

We use the overriding of configuration during CI to prevent accidentally committing one of the configuration files with properties to allowing freeze store updates to master (the files are committed to the repo, but should be committed with `false`. In the event we make a mistake we rely on the configuration override to not allow store updates during the CI run, causing tests to fail.) https://www.archunit.org/userguide/html/000_Index.html#_configuration

It seems valid that this snippet of code is hit repeatedly. I don't think however it's meant to log this line so excessively.

I'll happily provide a PR with a solution for this, but I'd like to get your opinion as to what would be a fitting solution. Should the log line be deleted, should the log level be configurable, should it only fire x times or something entirely different?

### PR DESCRIPTION
**Pull request by ldebruijn**
Prevent excessive logging when supplying system properties by only logging if the system properties have changed.

I didn't add test case for the change since the `ArchConfigurationTest#allows_to_override_any_property_via_system_property` already covers the scenario.

Resolves #375

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/ArchConfiguration.java
+++ b/archunit/src/main/java/com/tngtech/archunit/ArchConfiguration.java
@@ -273,15 +273,16 @@ public final class ArchConfiguration {
                 ENABLE_MD5_IN_CLASS_SOURCES, Boolean.FALSE.toString()
         ));
 
-        private final Properties properties = createProperties(PROPERTY_DEFAULTS);
+        private final Properties baseProperties = createProperties(PROPERTY_DEFAULTS);
+        private final Properties overwrittenProperties = new Properties();
 
         void clear() {
-            properties.clear();
-            properties.putAll(PROPERTY_DEFAULTS);
+            replaceProperties(baseProperties, PROPERTY_DEFAULTS);
+            overwrittenProperties.clear();
         }
 
         void load(InputStream inputStream) throws IOException {
-            properties.load(inputStream);
+            baseProperties.load(inputStream);
         }
 
         Set<String> stringPropertyNames() {
@@ -301,21 +302,31 @@ public final class ArchConfiguration {
         }
 
         void setProperty(String propertyName, String value) {
-            properties.setProperty(propertyName, value);
+            baseProperties.setProperty(propertyName, value);
         }
 
         void remove(String propertyName) {
-            properties.remove(propertyName);
+            baseProperties.remove(propertyName);
         }
 
         Properties getMergedProperties() {
-            Properties result = createProperties(this.properties);
-            Properties overwritten = getSubProperties("archunit", System.getProperties());
-            if (!overwritten.isEmpty()) {
-                LOG.info("Merging properties: The following properties have been overwritten by system properties: {}", overwritten);
+            Properties result = createProperties(baseProperties);
+            Properties currentlyOverwritten = getSubProperties("archunit", System.getProperties());
+            result.putAll(currentlyOverwritten);
+
+            if (!overwrittenProperties.equals(currentlyOverwritten)) {
+                replaceProperties(overwrittenProperties, currentlyOverwritten);
+                if (!currentlyOverwritten.isEmpty()) {
+                    LOG.info("Merging properties: The following properties have been overwritten by system properties: {}", currentlyOverwritten);
+                }
             }
-            result.putAll(overwritten);
+
             return result;
+        }
+
+        private static void replaceProperties(Properties properties, Properties newProperties) {
+            properties.clear();
+            properties.putAll(newProperties);
         }
 
         private static Properties createProperties(Map<?, ?> entries) {
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/DomainObjectCreationContext.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/DomainObjectCreationContext.java
@@ -135,8 +135,8 @@ public class DomainObjectCreationContext {
         return new JavaEnumConstant(builder);
     }
 
-    public static Source createSource(URI uri, Optional<String> sourceFileName) {
-        return new Source(uri, sourceFileName);
+    public static Source createSource(URI uri, Optional<String> sourceFileName, boolean md5InClassSourcesEnabled) {
+        return new Source(uri, sourceFileName, md5InClassSourcesEnabled);
     }
 
     public static <CODE_UNIT extends JavaCodeUnit> ThrowsClause<CODE_UNIT> createThrowsClause(CODE_UNIT codeUnit, List<JavaClass> types) {
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/Source.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/Source.java
@@ -48,10 +48,10 @@ public class Source {
     private final Optional<String> fileName;
     private final Md5sum md5sum;
 
-    Source(URI uri, Optional<String> fileName) {
+    Source(URI uri, Optional<String> fileName, boolean md5InClassSourcesEnabled) {
         this.uri = checkNotNull(uri);
         this.fileName = checkNotNull(fileName);
-        md5sum = Md5sum.of(uri);
+        md5sum = md5InClassSourcesEnabled ? Md5sum.of(uri) : Md5sum.DISABLED;
     }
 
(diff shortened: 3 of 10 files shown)

### CODE REVIEW

File: archunit/src/main/java/com/tngtech/archunit/ArchConfiguration.java [hunk 1]
```diff
@@ -273,10 +273,12 @@ public String getPropertyOrDefault(String propertyName, String defaultValue) {
                 ENABLE_MD5_IN_CLASS_SOURCES, Boolean.FALSE.toString()
         ));
 
-        private final Properties properties = createProperties(PROPERTY_DEFAULTS);
+        private Properties properties = createProperties(PROPERTY_DEFAULTS);
```
codecholeric, line 276: Why did you make these non-`final`? In the original version the `Properties properties` never changed anyway, but only the `overwrittenProperties`, right?

archunit/src/main/java/com/tngtech/archunit/ArchConfiguration.java
```diff
@@ -311,13 +313,23 @@ void remove(String propertyName) {
...
+
+            if (overwrittenPropertiesHaveNotChanged(overwritten)) {
+                return result;
             }
+
+            LOG.info("Merging properties: The following properties have been overwritten by system properties: {}", overwritten);
+            this.overwrittenProperties = overwritten;
+
             result.putAll(overwritten);
+            this.properties = result;
```
codecholeric, line 325: Why do you overwrite the original `properties`? Isn't that way the basic configuration gone afterwards?
I would expect the behavior
```
getProperty("foo") -> bar from archunit.properties
System.setProperty("archunit.foo", "baz")
getProperty("foo") -> baz from System properties
System.clearProperty("archunit.foo")
getProperty("foo") -> again bar from archunit.properties
```

That behavior would now be broken, wouldn't it? Because the initial properties will never be restored to the ones from the `archunit.properties` if system properties are unset?

codecholeric:
Actually now that I looked into it, it also seems like the properties are accessed very often, so maybe scrap my comment about generating them on the fly. I always thought it would read a couple of properties for a run, but actually it reads at least one property for every class. I'll profile a little to see if it would be worth it to cache it somehow.
Nevertheless I think it is a good idea to also have the `finalProperties` as a field and only regenerate that one if the `overwrittenProperties` have changed :wink:

codecholeric:
Funny, when I profile it, it hardly makes a difference, if the properties are generated thousands of times or just a couple :joy: The JVM is a fascinating thing...
But anyway, I guess it might be better to only regenerate these properties if they actually change...

ldebruijn:
I indeed misunderstood your proposal 😃 . I think this new commit should be more in line with what you described. 

I'd have expected at least some benefit by using some sort of cache 🤔 fascinating indeed 😄 .

I also added a test case to cover the behavior you described in one of your comments:
```
getProperty("foo") -> bar from archunit.properties
System.setProperty("archunit.foo", "baz")
getProperty("foo") -> baz from System properties
System.clearProperty("archunit.foo")
getProperty("foo") -> again bar from archunit.properties
```

archunit/src/main/java/com/tngtech/archunit/ArchConfiguration.java
ldebruijn: I did this because I misunderstood the structure of the code. After your explanation in one of these comments I see it does indeed not make sense

archunit/src/main/java/com/tngtech/archunit/ArchConfiguration.java
ldebruijn: I did this because I misunderstood the structure of the code. After your explanation in one of these comments I see it does indeed not make sense

codecholeric:
Thanks for adding the additional test :smile: That should have already been tested when the original change to override properties was introduced :disappointed: 

I added some review commits, because I think it's easier that way. I looked at it again and I 
realized that the `finalProperties` does not really provide any value the way it is now, because we need to build the full properties, then compare it to `finalProperties`. Then if equal return `finalProperties` otherwise update, but by that we don't really save anything :thinking: So I thought that maybe just using the merged properties on the fly is good enough then (and it cannot go out of sync as easily). Also I think we should never log empty overwritten properties, but that can happen if we log every time the overwritten properties change, no matter the change...

One thing I also did is refactor the import to only look up the `md5Enabled` property once for the whole run. I don't think it makes sense to support updating it during the import. That by itself would probably already have solved your original problem :rofl: But in the end I think it does make sense to not repeatedly log the overwritten properties anyway :+1: 

Okay, anyway, can you look at my commits and tell me what you think? If the current state looks good to you I would rebase against master and squash the commits fixing the properties together, i.e. creating a clear history with the relevant changes for the future :wink:

archunit/src/main/java/com/tngtech/archunit/ArchConfiguration.java
```diff
@@ -311,16 +310,23 @@ void remove(String propertyName) {
...
-
-            result.putAll(overwritten);
-            if (!this.finalProperties.equals(result)) {
-                LOG.info("Merging properties: The following properties have been overwritten by system properties: {}", overwritten);
-                this.finalProperties = result;
+            Properties result = createProperties(baseProperties);
+            Properties currentlyOverwritten = getSubProperties("archunit", System.getProperties());
+            result.putAll(currentlyOverwritten);
+
+            if (!overwrittenProperties.equals(currentlyOverwritten)) {
```
ldebruijn, line 317: I'm unsure about the goal of the `overwrittenProperties` field. Is it used to compare against no overrides vs what's currently available on system overrides. Would it make sense to update the field?

ldebruijn:
Nice, Apart from the single question I added to one of your changes I do agree the current state looks good. If you could answer that I'll rebase, squash and update the PR :)

archunit/src/main/java/com/tngtech/archunit/ArchConfiguration.java
codecholeric: `overwrittenProperties` is there to only log on change :thinking: I.e. if you request properties a second time, but you still have the same `overwrittenProperties` (i.e. `overwrittenProperties` == `currentlyOverwritten`) then we don't want to log, right? After checking you could theoretically always update `overwrittenProperties`, but if they are equal anyway, it is unnecessary, isn't it? If not equal, then `overwrittenProperties` is updated in the next line...
I hope I understood your question? Do you think it would be easier to understand, if we would always update `overwrittenProperties`, even if they are equal to `currentlyOverwritten` anyways?

codecholeric:
When you squash, please keep `only check if MD5 is enabled once per import` as a separate commit :wink: (that one is complex enough to deserve its own entry with explanation in the history :joy:)

ldebruijn:
Squashed and ready to be merged! :)

PR #377, opened 2020-06-11
Allow usage of TestSets in Gradle

ebramirez:
When using TestSets in Gradle, tests are separated in different test folders, this change allows the use of `DoNotIncludeTests` to filter out those tests.


ebramirez:
Need to create a new version.

PR #378, opened 2020-06-11
Skip tests in Gradle TestSets

ebramirez:
When using TestSets in Gradle, tests are separated in different test folders, this change allows the use of `DoNotIncludeTests` import option to filter out those tests.


ebramirez:
Sorry about the DCO, if it is not possible to fix it, I will create the PR again.

hankem:
> about the DCO, if it is not possible to fix it

That can easily be done: there is a a hint on [the DCO check page](https://github.com/TNG/ArchUnit/pull/378/checks?check_run_id=760867100) on the `git rebase` command for the `--signoff`. Maybe you can also use this opportunity to squash the two commits in one?

PR #379, opened 2020-06-11
Exclude Gradle TestSet with DoNotIncludeTests

ebramirez:
When using TestSets in Gradle, tests are separated in different test folders, this change allows the use of DoNotIncludeTests import option to filter out those tests.

Signed-off-by: Eduardo Ramirez <eramirez@redhat.com>

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
@@ -77,9 +77,10 @@ public interface ImportOption {
     final class DoNotIncludeTests implements ImportOption {
         private static final Pattern MAVEN_PATTERN = Pattern.compile(".*/target/test-classes/.*");
         private static final Pattern GRADLE_PATTERN = Pattern.compile(".*/build/classes/([^/]+/)?test/.*");
+        private static final Pattern GRADLE_TESTSET_PATTERN = Pattern.compile( ".*/bin/([^/]+/)?test.*" );
         private static final Pattern INTELLIJ_PATTERN = Pattern.compile(".*/out/test/classes/.*");
 
-        private static final Set<Pattern> EXCLUDED_PATTERN = ImmutableSet.of(MAVEN_PATTERN, GRADLE_PATTERN, INTELLIJ_PATTERN);
+        private static final Set<Pattern> EXCLUDED_PATTERN = ImmutableSet.of(MAVEN_PATTERN, GRADLE_PATTERN, GRADLE_TESTSET_PATTERN, INTELLIJ_PATTERN);
 
         @Override
         public boolean includes(Location location) {


codecholeric:
Thanks a lot for your support! :smiley: Can you elaborate a little, in which case this exactly happens? I.e. which Gradle configuration causes this behavior?

ebramirez:
Sure!

[TestSets](https://github.com/unbroken-dome/gradle-testsets-plugin) is a Gradle plugin that allows treating folders as tests sets, that is: If you create a `src/integrationTest` folder and a typical structure of `src/integrationTest/java` and `src/integrationTest/resources`, then you can do a `gradle integrationTest` to run the tests inside that folder.

I know this plugin is not part of the default Gradle behavior, but this change is transparent to those not using it, but helps a lot those using it.

Thanks for your hard work!

codecholeric:
While I really appreciate the initiative, I'm not convinced it is a good idea to add this. Here is my reasoning:
1) Yes, technically it is transparent to add this for people not using the plugin. But the moment we merge it, we support this use case (and as you said, it's not default Gradle behavior). So if the maintainers of the plugin decide to change their conventions in 6 months, then we will have an issue that ArchUnit suddenly "does not work anymore" with this plugin. And then we need to fix it (again, it is probably a small thing, but it still costs resources and commitment on our side)
2) While the impact is probably quite slim, each additional pattern to check also costs a little bit of performance. Since only a very small part of the user base probably uses this plugin, I'm not convinced this is a good deal.

Moreover ArchUnit was in particular designed to be as extensible as possible so users could easily support their use cases without the need for ArchUnit to have all possible use cases integrated.
For example you could easily write your custom `ImportOption`

```
static class DoNotIncludeTestSets implements ImportOption {
    private static final Pattern GRADLE_TESTSET_PATTERN = Pattern.compile( ".*/bin/([^/]+/)?test.*" );

    @Override
        return !location.matches(GRADLE_TESTSET_PATTERN);
    }
}
```

Then either use it as a substitute for `DoNotIncludeTests` or simply combine it

```
@AnalyzeClasses(..., importOptions = {DoNotIncludeTests.class, DoNotIncludeTestSets.class})
public class CodingRulesTest {
```

Or write your own `ImportOption` that internally combines `DoNotIncludeTests` and `DoNotIncludeTestSets` if you want to have one consistent interface.

And if you think this is something that is useful for others, you can also share it on GitHub so people find it if they search for "ArchUnit gradle testsets", etc.! (I am sure you're not the only one this is useful to, I'm just not convinced the user base is big enough to justify integrating it into the very core of ArchUnit).

I hope you understand my point of view. If not, feel free to discuss :wink:

ebramirez:
I understand you reluctance to having to maintain another use case.

I created this PR because I wanted everyone to have a seamless experience and avoid "Why is this not working?" moment for anyone using archunit with the plugin, but I guess that if this hasn't cam up before means that its not a common combination (Only two projects on GH).

And yes, creating my own ImportOption is what I did as a temporary workaround.

codecholeric:
Yeah, as I said, I'm super happy about any support and I really appreciate your thinking to make this better for everyone! :smiley: 
But I'm glad you also see my point. If this should become widely used and some sort of standard for testing Gradle projects, I'm more than happy to merge it into the core!
I really hope you're not discouraged from supporting ArchUnit in the future if you have an idea to make the project better!! (because I know there is still a lot to do :wink:)

PR #381, opened 2020-06-11
Add note about JUnit's `@Disabled` to documentation

### LINKED ISSUE #380: JUnit 5 @Disabled annotation has no effect
With JUnit 5, ArchUnit should not run tests annotated with `@Disabled`.

rweisleder:
From now on, users should be able to find `@ArchIgnore` if they search for one of "skip", "disable" or "ignore".

Resolves #380

### DIFF
--- a/docs/userguide/009_JUnit_Support.adoc
+++ b/docs/userguide/009_JUnit_Support.adoc
@@ -156,6 +156,9 @@ public class ArchitectureTest {
 }
 ----
 
+Note for users of JUnit 5: the annotation `@Disabled` has no effect here.
+Instead, `@ArchIgnore` should be used.
+
 ==== Grouping Rules
 
 Often a project might end up with different categories of rules, for example "service rules"


docs/userguide/009_JUnit_Support.adoc
```diff
@@ -156,6 +156,8 @@ public class ArchitectureTest {
 }
 ----
 
+Note that JUnit's `@Disabled` annotation is not supported. Instead, `@ArchIgnore` should be used.
```
hankem, line 159: Thanks for the fast reaction! 👍

Just one thought: As the section is about JUnit **4 & 5**, I wonder whether we should more specfically refer to
> Junit 5's `@Disabled`

here? On the other hand, this probably shouldn't confuse JUnit 4 users, so I don't mind keeping the note simple.

codecholeric: Extending the docs is a good idea :+1: Since this section is a bout JUnit 4 and 5, what about writing sth. like "For JUnit 5 note that ArchUnit does not support JUnit Jupiter annotations like `@Disabled`, but provides its own API" or something like that? :thinking: I think it would be good to somehow make it clear that this is a pure JUnit 5 comment. But maybe that same sentence with a prefix "Note for JUnit 5 users:" or similar would be good enough?

Issue #371, opened 2020-06-02
`instanceof` calls are not recogniced as dependencies


t-h-e:
If a class contains an `instanceof X` call, then `X` is not part of the dependencies of that class.

Following example class:
```java
public class DependsOnMyExceptionInstanceOf {

    private void bla(Exception e) {
        if (e instanceof MyException) {
            System.out.println(e);
        }
    }
}
```

with the following arch test fails:
```java
@ArchTest
static final ArchRule dependsOnMyExceptionInstanceOf = theClass(DependsOnMyExceptionInstanceOf.class)
    .should()
    .dependOnClassesThat()
    .areAssignableTo(MyException.class);
```

I checked the ArchUnit code and noticed that `MethodProcessor` does not override `visitTypeInsn` which recognises `instanceof` calls.

If I naively add the following code, `instanceof`
```java
@Override
public void visitTypeInsn(int opcode, String type) {
    if (opcode == Opcodes.INSTANCEOF) {
        accessHandler.handleMethodInstruction(type, CONSTRUCTOR_NAME, "()V");
    }
}
```

As I'm new to the codebase of ArchUnit this might not be the best solution 😅 
Any suggestions how to address this issue?

codecholeric:
Yeah, I'm afraid you're right about ArchUnit not detecting this yet. From what I see, any `JavaCodeUnit` can check for `instanceof`, so I would guess to fit this in the existing design it could be something like

* instead of using `accessHandler`, simply add `JavaCodeUnitBuilder.instanceofChecks` (similar to `throwsDeclarations`)
* in the place you found add `codeUnitBuilder.withInstanceofCheck(JavaTypeImporter.createFromAsmObjectTypeName(type)`
* add something like a `JavaCodeUnit.instanceofChecks: Set<InstanceofCheck>` where `class InstanceofCheck{ JavaClass origin; JavaClass target }`
* in the constructor of `JavaCodeUnit` assign `codeUnitBuilder.getInstanceofChecks()` (which internally can use the meanwhile resolved `importedClasses` to resolve those types within the `instanceofChecks` that you have registered earlier)
* add `codeUnit.instanceofChecks` to the set returned by `codeUnit.getDirectDependencies{from/to}self()` through sth. like `Dependency.tryCreateFromInstanceofCheck(instanceofCheck)`

At least that's what comes to my mind without further research :thinking: 
If you want to give it a shot with a PR, I'm happy to support you with any questions :smiley:

[...]

PR #383, opened 2020-06-12
Consider classes referenced in instanceof

t-h-e:
Resolves: https://github.com/TNG/ArchUnit/issues/371

One failing test is still in the code due to another issue.

Feedback is highly appriciated 😃

### DIFF
--- a/archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterTest.java
@@ -172,6 +172,7 @@ import com.tngtech.archunit.core.importer.testexamples.simpleimport.InterfaceToI
 import com.tngtech.archunit.core.importer.testexamples.simplenames.SimpleNameExamples;
 import com.tngtech.archunit.core.importer.testexamples.specialtargets.ClassCallingSpecialTarget;
 import com.tngtech.archunit.core.importer.testexamples.syntheticimport.ClassWithSynthetics;
+import com.tngtech.archunit.testutil.ArchConfigurationRule;
 import com.tngtech.archunit.testutil.LogTestRule;
 import com.tngtech.archunit.testutil.OutsideOfClassPathRule;
 import com.tngtech.java.junit.dataprovider.DataProvider;
@@ -180,7 +181,6 @@ import com.tngtech.java.junit.dataprovider.UseDataProvider;
 import org.apache.logging.log4j.Level;
 import org.assertj.core.api.Condition;
 import org.assertj.core.util.Objects;
-import org.junit.After;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
@@ -242,11 +242,8 @@ public class ClassFileImporterTest {
     public final LogTestRule logTest = new LogTestRule();
     @Rule
     public final IndependentClasspathRule independentClasspathRule = new IndependentClasspathRule();
-
-    @After
-    public void tearDown() {
-        ArchConfiguration.get().reset();
-    }
+    @Rule
+    public final ArchConfigurationRule archConfigurationRule = new ArchConfigurationRule();
 
     @Test
     public void imports_simple_package() throws Exception {


archunit/src/main/java/com/tngtech/archunit/core/importer/DomainBuilders.java
```diff
@@ -222,8 +225,10 @@ JavaField construct(JavaFieldBuilder builder, ClassesByTypeName importedClasses)
         private JavaType returnType;
         private List<JavaType> parameters;
         private List<JavaType> throwsDeclarations;
+        private List<JavaType> instanceOfChecks;
 
         private JavaCodeUnitBuilder() {
+            instanceOfChecks = new ArrayList<>();
```
codecholeric, line 231: Why not add this right to the field declaration?

archunit/src/main/java/com/tngtech/archunit/core/domain/DomainObjectCreationContext.java
```diff
@@ -143,6 +143,14 @@ public static Source createSource(URI uri, Optional<String> sourceFileName) {
         return ThrowsClause.from(codeUnit, types);
     }
 
+    public static Set<InstanceofCheck> createInstanceofCheck(JavaCodeUnit codeUnit, List<JavaClass> types) {
```
codecholeric, line 146: I think this method should use the plural, i.e. be called `createInstanceofChecks(..)`, since it creates a set of `InstanceofCheck`

archunit/src/main/java/com/tngtech/archunit/core/importer/JavaClassProcessor.java
```diff
@@ -324,6 +324,13 @@ public void visitMethodInsn(int opcode, String owner, String name, String desc,
             accessHandler.handleMethodInstruction(owner, name, desc);
         }
 
+        @Override
+        public void visitTypeInsn(int opcode, String type) {
+            if (opcode == Opcodes.INSTANCEOF) {
+                codeUnitBuilder.addInstanceOfCheck(JavaTypeImporter.createFromAsmObjectTypeName(type));
```
codecholeric, line 330: As far as I can see, you can even pass in the `actualLineNumber` here, which will be really neat, because failure messages can report the exact location where the `instanceof` check has appeared! :smile: 
You might need something like `RawInstanceofCheck` though, to transport the info until you have all the classes imported...

archunit/src/test/java/com/tngtech/archunit/core/domain/DependencyTest.java
```diff
@@ -248,6 +262,14 @@ void method() throws IOException {
         }
     }
 
+    private static class ClassWithDependencyOnInstanceofCheck {
```
codecholeric, line 265: Would be nice to add some @SuppressWarnings("unused") here like for the other little test class snippets, so the IDE doesn't always give false positive warnings.

archunit/src/test/java/com/tngtech/archunit/core/domain/DependencyTest.java
```diff
@@ -71,6 +71,20 @@ public void Dependency_from_throws_declaration() {
...
+        JavaMethod origin = importClassesWithContext(ClassWithDependencyOnInstanceofCheck.class, SomeMemberType.class)
+                .get(ClassWithDependencyOnInstanceofCheck.class).getMethod("method", Object.class);
+        InstanceofCheck instanceofCheck = getOnlyElement(origin.getInstanceofChecks());
+
+        Dependency dependency = Dependency.tryCreateFromInstanceofCheck(instanceofCheck).get();
+
+        assertThat(dependency.getOriginClass()).matches(ClassWithDependencyOnInstanceofCheck.class);
+        assertThat(dependency.getTargetClass()).matches(SomeMemberType.class);
+        assertThat(dependency.getDescription()).as("description")
+                .contains("Method <" + origin.getFullName() + "> checks instanceof <" + SomeMemberType.class.getName() + ">");
```
codecholeric, line 85: If you add the line number to `InstanceofCheck` you can add a `SourceCodeLocation` to `Dependency` (compare method calls, etc.). Then you could assert the correct location link snippet here (`Foo.java:123`)

archunit/src/test/java/com/tngtech/archunit/core/domain/InstanceofCheckTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+import org.assertj.core.api.iterable.Extractor;
+import org.junit.Ignore;
+import org.junit.Test;
+
+import static com.tngtech.archunit.core.domain.TestUtils.importClasses;
+import static com.tngtech.archunit.core.domain.TestUtils.importMethod;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class InstanceofCheckTest {
```
codecholeric, line 13: This test looks a little inconsistent to me. For example there is also no `ThrowsDeclarationTest`. ArchUnit does not have any convention, that "every class must have a unit test class" :wink: 
For example, you assert dependencies of `JavaClass`, so why are these tests here and not within `JavaClassTest`? There are already tests for other sorts of dependencies there, I would add it to that. But then I feel the first test asserting only one dependency provides no value against the detailed test in `DependencyTest` and thus there are no meaningful tests left :wink:

archunit/src/test/java/com/tngtech/archunit/core/domain/InstanceofCheckTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @Test
+    public void class_contains_all_instanceof_dependencies() {
+        JavaClass javaClass = importClasses(SomeClass.class).get(SomeClass.class);
+
+        assertThat(javaClass.getDirectDependenciesFromSelf()).extracting(simpleNameOfTargetClass()).contains(A.class.getSimpleName(), B.class.getSimpleName(), C.class.getSimpleName());
+    }
+
+    @Test
+    @Ignore("Static initializers are not registered in MemberDependenciesByTarget")
```
codecholeric, line 34: The reason is, that it either made no sense (e.g. when parameters, return types, throws declarations, etc., are added, none of those are existing for a static initializer), or there is not enough info at the time (accesses of various type, at the point of registering those, we only have some raw string `desc` but the real "target" `JavaMember` is constructed at a way later point in time; you can see how that is handled within `JavaClassProcessor.AccessHandler` and `DomainObjectCreationContext.AccessContext`)
`InstanceofChecks` are just different in semantics. While the other ones are part of the declaration (parameters, return type, ...) `InstanceofChecks` are statements and reside within the bode of a `JavaCodeUnit`.
What do you think about adding a method `ClassGraphCreator.MemberDependenciesByTarget.registerStaticInitializer(..)` and registering only the `InstanceofChecks` by target there?
You are definitely right that this test should pass, so good catch!! :+1:

codecholeric:
If you want, you can also add an instanceof dependency to one of the `archunit-example` classes and then assert the additional dependency in `com.tngtech.archunit.integration.ExamplesIntegrationTest` :smiley: That way someone executing the respective example rule (e.g. `layeredArchitecture`) could see how `instanceof` checks are detected and what they look like.

archunit-integration-test/src/test/java/com/tngtech/archunit/ImporterRules.java
```diff
@@ -26,6 +26,7 @@
                     .that().resideOutsideOfPackage(THIRDPARTY_PACKAGE_IDENTIFIER)
                     .and(not(belongToAnyOf(JavaType.class)))
                     .and().doNotHaveFullyQualifiedName("com.tngtech.archunit.core.importer.JavaTypeImporter")
+                    .and().doNotHaveFullyQualifiedName("com.tngtech.archunit.core.importer.JavaClassProcessor$AnnotationArrayProcessor")
```
codecholeric, line 29: This should not be necessary, and actually it is an intended rule :wink: If it would really fail it would mean you're actually not doing it the way intended. In the past I did have some trouble when I found corner cases where `..asm..Type` would not behave the way I needed it to. So I encapsulated all that knowledge within `JavaTypeImporter` and forbid direct access to `..asm..Type` so I would not run into the same pitfall again :wink:

archunit/src/main/java/com/tngtech/archunit/core/importer/JavaClassProcessor.java
t-h-e: Seems to work, although I'm not completely sure why there is a difference between `lineNumber` and `sourceCodeLocation` in `Dependency`.

archunit-integration-test/src/test/java/com/tngtech/archunit/ImporterRules.java
t-h-e: I thought that it is there for a reason and I was confused myself why this test failed. But `JavaClassProcessor` was always using `..asm..Type`. It is just with this change that it becomes apparent, because `instanceOf Type` is checked :D

t-h-e:
> If you want, you can also add an instanceof dependency to one of the `archunit-example` classes and then assert the additional dependency in `com.tngtech.archunit.integration.ExamplesIntegrationTest` 😃 That way someone executing the respective example rule (e.g. `layeredArchitecture`) could see how `instanceof` checks are detected and what they look like.

I tried to, but I could not get the test to work properly. I will put an example in a separate commit.

t-h-e:
And apparently I am not able to work with Github properly and always destroy my pull requests.
I forgot to sign off on the commits and when I follow the instructions of the "Sign-Off" Bot the pull request is closed. 😞 
I will try again tomorrow...

archunit-integration-test/src/test/java/com/tngtech/archunit/ImporterRules.java
codecholeric: Oh yes, I remember seeing this and thinking "not so much harm and not detected anyway at the moment" :joy: Because what I was worried about couldn't happen by an `instanceof` check :wink: I think even this thing should then probably be encapsulated in `JavaTypeImporter` or `JavaType`. I really want to avoid as many dependencies on `..asm..Type` outside of these classes as possible as it is a source of error as mentioned...

PR #384, opened 2020-06-13
enhance CompositeArchRule.of

sullis:
- new method:  of(Collection)

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/lang/CompositeArchRule.java
+++ b/archunit/src/main/java/com/tngtech/archunit/lang/CompositeArchRule.java
@@ -15,6 +15,7 @@
  */
 package com.tngtech.archunit.lang;
 
+import java.util.Iterator;
 import java.util.List;
 
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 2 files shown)


sullis:
WDYT? @codecholeric

codecholeric:
Thanks a lot for the PR! :smiley: The only thing I'm wondering is: Wouldn't a single method that takes an `Iterable<ArchRule>` be sufficient? It would also swallow `Collection<ArchRule>` then. But I can't really imagine a scenario where one would only have  an `Iterator<ArchRule>` :thinking: That's not a type you usually really use, is it?

sullis:
> Thanks a lot for the PR! 😃 The only thing I'm wondering is: Wouldn't a single method that takes an `Iterable<ArchRule>` be sufficient? It would also swallow `Collection<ArchRule>` then. But I can't really imagine a scenario where one would only have an `Iterator<ArchRule>` 🤔 That's not a type you usually really use, is it?

I removed of(Iterator).  I plan to use of(Collection) is one my internal projects.

sullis:
@codecholeric WDYT?

sullis:
@codecholeric  are there any plans to release a new version of ArchUnit?   (This PR was completed in July but there has not been any releases since May)

sullis:
@codecholeric are there any plans to release a new version of ArchUnit? (This PR was completed in July 2020 but there have not been any releases since May 2020)

codecholeric:
Yes, I know I'm super slow this time :disappointed: In short, yes, I'm waiting for a review on #484 and I want to implement the counter part of #428 to not have a difference between dependencies from self and to self. I actually planned to release this weekend, but didn't completely make it. Need to make sure to release quicker iterations in the future, this time the interval really is way too big :disappointed:

PR #385, opened 2020-06-17
Fix example output for configured JDK in Contributing Guide

rweisleder:
> ArchUnit requires at least JDK 11 to build.

So the example output should also show at least version.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ Windows users should use `gradlew.bat` instead.
 ```
 $ cd /path/to/git/clone/of/ArchUnit
 $ ./gradlew showJdkVersion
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -11,7 +11,7 @@ Windows users should use `gradlew.bat` instead.
 ```
 $ cd /path/to/git/clone/of/ArchUnit
 $ ./gradlew showJdkVersion
-Configured JDK: 1.9
+Configured JDK: 11
```
codecholeric, line 14: For the sake of maintainability, what about just making it 14? :rofl: I tought you're working on having a separate test source set for Record tests which will then require the whole build to be at least JDK 14? :wink:

PR #387, opened 2020-06-19
enable Dependabot v2

sullis:
reference:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "bundler"
+    directory: "/docs"
+    schedule:
+      interval: "daily"


codecholeric:
Sorry that I'm so late, thanks for the initiative! I think it makes sense. Do you have any strong opinion about Dependabot vs. Renovate? Because there also was a Renovate PR, but it did scare me a little that it claimed "if added Renovate will open 21 Pull Requests" :joy:
It felt like Dependabot might be a little bit more conservative, since after all this is a library, so we cannot always just go to the newest version no matter compatibility issues. But ArchUnit meanwhile has a really good test coverage on various platforms with various JDKs, so I would be pretty confident that if the CI suite runs through, the version upgrade is safe to merge...
Another note, do we have to add the Ruby (Gem? Jekyll? ...) ecosystem for the ArchUnit docs, too? Because the old Dependabot prototype I've installed so far does add PRs for the website as well (see e.g. #420)

sullis:
> Sorry that I'm so late, thanks for the initiative! I think it makes sense. Do you have any strong opinion about Dependabot vs. Renovate? Because there also was a Renovate PR, but it did scare me a little that it claimed "if added Renovate will open 21 Pull Requests" 😂
> It felt like Dependabot might be a little bit more conservative, since after all this is a library, so we cannot always just go to the newest version no matter compatibility issues. But ArchUnit meanwhile has a really good test coverage on various platforms with various JDKs, so I would be pretty confident that if the CI suite runs through, the version upgrade is safe to merge...
> Another note, do we have to add the Ruby (Gem? Jekyll? ...) ecosystem for the ArchUnit docs, too? Because the old Dependabot prototype I've installed so far does add PRs for the website as well (see e.g. #420)


I have not used Renovate. 

 I've been using Dependabot v2 on all of my open source projects.

Dependabot v2 will upgrade:
 - Maven plugins
 - Maven dependencies
 - Gradle plugins
 - Gradle dependencies
 - GitHub Actions

codecholeric (approved):
Okay, I've added `bundler` on `docs` to the checks to keep the website/docs up to date. I'm curious if it will really work correctly, cause the good old Java 7 compatibility still causes some restrictions where certain dependencies cannot simply be upgraded. But we will see :wink:

PR #392, opened 2020-07-02
Add JavaClass.tryGetConstructor(..) methods

### LINKED ISSUE #386: Add JavaClass.tryGetConstructor methods
Currently, we have
* `JavaClass.getField(String)` and `JavaClass.tryGetField(String)`
* `JavaClass.getMethod(String)` and `JavaClass.tryGetMethod(String)`
* `JavaClass.getConstructor()` but no `JavaClass.tryGetConstructor()`

To be able to query constructors without handling exceptions, it would be nice to have
* `JavaClass.tryGetConstructor()` for the default constructor
* `JavaClass.tryGetConstructor(Class<?>... parameters)`
* `JavaClass.tryGetConstructor(String... parameters)`

Maybe I will find the time to create a PR in the next days.

rweisleder:
Before this commit, JavaClass had the methods
* `getField(..)` and `tryGetField(..)`
* `getMethod(..)` and `tryGetMethod(..)`
* `getConstructor(..)`

This PR adds the equivalent `tryGetConstructor(..)` methods.

Resolves #386

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
@@ -673,11 +673,18 @@ public class JavaClass implements HasName.AndFullName, HasAnnotations<JavaClass>
         return allFields.get();
     }
 
+    /**
+     * @return The field with the given name.
+     * @throws IllegalArgumentException If this class does not have such a field.
+     */
     @PublicAPI(usage = ACCESS)
     public JavaField getField(String name) {
         return tryGetField(name).getOrThrow(new IllegalArgumentException("No field with name '" + name + " in class " + getName()));
     }
 
+    /**
+     * @return The field with the given name, if this class has such a field, otherwise {@link Optional#absent()}.
+     */
     @PublicAPI(usage = ACCESS)
     public Optional<JavaField> tryGetField(String name) {
         for (JavaField field : fields) {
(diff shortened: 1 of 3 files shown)


archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
```diff
@@ -789,21 +818,58 @@ public JavaMethod getMethod(String name, String... parameters) {
         return allMethods.get();
     }
 
+    /**
+     * @return The constructor with zero parameters (the default constructor).
```
hankem, line 822: I'd drop `(the default constructor)` here to avoid confusion with the implicit constructor from [JLS §8.8.9](https://docs.oracle.com/javase/specs/jls/se14/html/jls-8.html#jls-8.8.9) that is automatically added in absence of explicit constructor declarations – which, for inner classes, actually has a parameter.

hankem: Same for `tryGetConstructor()`.

rweisleder: Ah, you're right, I forgot the inner classes. In this case the "default constructor"-note would be confusing.

codecholeric:
Cool, thanks a lot :smile: I've resorted the methods a little, not because I thought the order was bad, but because I think it should be the same as for methods. Other than that I added a little test for the `absent()` version for all members, because I saw that no test actually ever tested this case (I don't consider the danger high, but the cost for the test is also low :wink:)

Issue #397, opened 2020-07-13
NPE from getDirectDependenciesFromSelf() called on a JavaClass stub


hankem:
I've run into an issue with transitive dependencies not being imported completely.

Here's a simple example (which, admittedly, looks strange on its own 😃) to illustrate it:
```java
new ClassFileImporter().importClasses(java.lang.Object.class)
        .get(java.lang.Object.class)
        .getMethod("finalize").getThrowsClause().getTypes().get(0)  // java.lang.Throwable
        .getMethod("getStackTrace").getRawReturnType().getComponentType()  // java.lang.StackTraceElement
        .getDirectDependenciesFromSelf();
```
Since 85506e2fcae12c2838e49700ef07635c58faa054, I'm getting a
```
java.lang.NullPointerException
	at com.tngtech.archunit.core.domain.JavaClass.getDirectDependenciesFromSelf(JavaClass.java:894)
```
The line number is for ArchUnit 0.14.1, but [`getDirectDependenciesFromSelf`](https://github.com/TNG/ArchUnit/blob/v0.14.1/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java#L892-L895) is currently a one-liner anyways; the issue is that `javaClassDependencies` is not initialized, i.e. that `JavaClass.completeFrom(ImportContext)` was not called. But here, I don't know what's expected for stubs.

**Should I be able to `getDirectDependenciesFromSelf()` from a stub `JavaClass`?** (Anyways, the NPE isn't a nice result for sure.)

The hypothesis that the error might be because `Throwable.getStackTrace()` returns `StackTraceElement[]` (which seems to be imported correctly), whereas I'm asking for dependencies from `StackTraceElement`, is supported by the following simpler example:
```java
new ClassFileImporter().importClasses(DependsOnArray.class)
        .get(DependsOnArray.class)
        .getField("array").getRawType().getComponentType()  // Element
        .getDirectDependenciesFromSelf();
```
with
```java
static class DependsOnArray {
    Element[] array;
}

static class Element {
}
```

I'm still puzzled though why the same issue does not show up in:
```java
new ClassFileImporter().importClasses(java.lang.Throwable.class)
        .get(java.lang.Throwable.class)
        .getMethod("getStackTrace").getRawReturnType().getComponentType()  // java.lang.StackTraceElement
        .getDirectDependenciesFromSelf();
```

codecholeric:
[...]

Issue #253, opened 2019-10-23
Support for checking transitive dependencies


lukfas:
ArchUnit supports checking for dependencies between classes using com.tngtech.archunit.lang.conditions.AnyDependencyCondition via com.tngtech.archunit.lang.conditions.ArchConditions.dependOnClassesThat(DescribedPredicate<? super JavaClass>). 

However, this Condition does not cover indirect dependencies. 

Checking for immediate dependencies ist sufficient in projects where there exists a consistent architectural guideline that specifies for every package / class the allowed dependencies.  

When refactoring a large legacy project without such detailed guidelines, one might still pose independence constraints on two (groups of) classes, but defining guidelines for all indirectly referenced classes would be too expensive. In this case, one has to check for the absence of transitive dependencies.

As a Proof of Concept, I have implemented a subclass TransitiveDependenceCondition of ArchCondition<JavaClass>. It checks for transitive dependencies from the current class to some target class satisfying a target-condition, without passing through other classes matching some ignore-condition. 

When used on about 20.000 classes, a check for transitive independence takes much less than the import of all classes. Hence the transitive dependency check itself is not really expensive. 
Although this POC is working in our project, it is still preliminary and not as general as it could be.

lukfas:
Here is the POC: [transitiveDependencies.zip] (https://github.com/TNG/ArchUnit/files/3762706/transitiveDependencies.zip)

codecholeric:
I think this should be solved by #401, otherwise feel free to reopen!

PR #401, opened 2020-07-15
add method JavaClass.getTransitiveDependenciesFromSelf

hankem:
Instead of the initial `getTransitiveDependenciesFromSelfWithinPackage(String packageNamePrefix)`, I now suggest the simpler API:

```java
    /**
     * Returns the transitive closure of all dependencies originating from this class, i.e. its direct dependencies
     * and the dependencies from all imported target classes.
     * @return all transitive dependencies (including direct dependencies) from this class
     * @see #getDirectDependenciesFromSelf()
     */
    @PublicAPI(usage = ACCESS)
    public Set<Dependency> getTransitiveDependenciesFromSelf() {
        return JavaClassTransitiveDependencies.findTransitiveDependenciesFrom(this);
    }
```

Signed-off-by: Manfred Hanke <Manfred.Hanke@tngtech.com>

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
@@ -147,6 +147,7 @@ public class JavaClass implements JavaType, HasName.AndFullName, HasAnnotations<
     });
     private JavaClassDependencies javaClassDependencies = new JavaClassDependencies(this);  // just for stubs; will be overwritten for imported classes
     private ReverseDependencies reverseDependencies = ReverseDependencies.EMPTY;  // just for stubs; will be overwritten for imported classes
+    private boolean fullyImported = false;
 
     JavaClass(JavaClassBuilder builder) {
         source = checkNotNull(builder.getSource());
@@ -1150,6 +1151,17 @@ public class JavaClass implements JavaType, HasName.AndFullName, HasAnnotations<
     }
 
     /**
+     * @return Whether this class has been fully imported, including all dependencies.<br>
+     *         Classes that are only transitively imported are not necessarily fully imported.<br><br>
+     *         Suppose you only import a class {@code Foo} that calls a method of class {@code Bar}.
+     *         Then {@code Bar} is, as a dependency of the fully imported class {@code Foo}, only transitively imported.
+     */
+    @PublicAPI(usage = ACCESS)
+    public boolean isFullyImported() {
+        return fullyImported;
+    }
+
+    /**
      * @param clazz An arbitrary type
      * @return true, if this {@link JavaClass} represents the same class as the supplied {@link Class}, otherwise false
      */
@@ -1301,6 +1313,7 @@ public class JavaClass implements JavaType, HasName.AndFullName, HasAnnotations<
             codeUnit.completeFrom(context);
         }
         javaClassDependencies = new JavaClassDependencies(this);
+        fullyImported = true;
         return javaClassDependencies;
     }
 
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
@@ -145,8 +145,8 @@ public class JavaClass implements JavaType, HasName.AndFullName, HasAnnotations<
                     .build();
         }
     });
-    private JavaClassDependencies javaClassDependencies;
-    private ReverseDependencies reverseDependencies;
+    private JavaClassDependencies javaClassDependencies = new JavaClassDependencies(this);  // just for stubs; will be overwritten for imported classes
+    private ReverseDependencies reverseDependencies = ReverseDependencies.EMPTY;  // just for stubs; will be overwritten for imported classes
 
     JavaClass(JavaClassBuilder builder) {
         source = checkNotNull(builder.getSource());
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
@@ -1032,6 +1032,17 @@ public class JavaClass implements JavaType, HasName.AndFullName, HasAnnotations<
     }
 
     /**
+     * Returns the transitive closure of all dependencies originating from this class, i.e. its direct dependencies
+     * and the dependencies from all imported target classes.
+     * @return all transitive dependencies (including direct dependencies) from this class
+     * @see #getDirectDependenciesFromSelf()
+     */
+    @PublicAPI(usage = ACCESS)
+    public Set<Dependency> getTransitiveDependenciesFromSelf() {
+        return JavaClassTransitiveDependencies.findTransitiveDependenciesFrom(this);
+    }
+
+    /**
      * Like {@link #getDirectDependenciesFromSelf()}, but instead returns all dependencies where this class
      * is target.
      *
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClassTransitiveDependencies.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClassTransitiveDependencies.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2014-2020 TNG Technology Consulting GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.tngtech.archunit.core.domain;
+
+import com.google.common.collect.ImmutableSet;
+
+import java.util.HashSet;
+import java.util.Set;
+
+class JavaClassTransitiveDependencies {
+    private JavaClassTransitiveDependencies() {
+    }
+
+    static Set<Dependency> findTransitiveDependenciesFrom(JavaClass javaClass) {
+        ImmutableSet.Builder<Dependency> transitiveDependencies = ImmutableSet.builder();
+        Set<JavaClass> analyzedClasses = new HashSet<>();  // to avoid infinite recursion for cyclic dependencies
+        addTransitiveDependenciesFrom(javaClass, transitiveDependencies, analyzedClasses);
+        return transitiveDependencies.build();
+    }
+
+    private static void addTransitiveDependenciesFrom(JavaClass javaClass, ImmutableSet.Builder<Dependency> transitiveDependencies, Set<JavaClass> analyzedClasses) {
+        analyzedClasses.add(javaClass);  // currently being analyzed
+        Set<JavaClass> targetClassesToRecurse = new HashSet<>();
+        for (Dependency dependency : javaClass.getDirectDependenciesFromSelf()) {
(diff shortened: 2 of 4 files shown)


archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClassTransitiveDependencies.java
```diff
@@ -0,0 +1,56 @@
...
+            Predicate<JavaClass> shouldRecurse,
+            Set<Dependency> transitiveDependencies,
+            Set<JavaClass> analyzedClasses) {
+
+        analyzedClasses.add(javaClass);  // currently being analyzed
+        Set<JavaClass> targetClassesToRecurse = new HashSet<>();
+        for (Dependency dependency : javaClass.getDirectDependenciesFromSelf()) {
+            transitiveDependencies.add(dependency);
+            JavaClass targetClass = dependency.getTargetClass();
+            JavaClass targetClassToRecurse = targetClass.isArray() ? targetClass.getComponentType() : targetClass;
```
hankem, line 45: Oops, my initial test did not cover the special handling for arrays. I'll update it.

hankem: Hm, doesn't github update PRs anymore when pushing to the source branch? I'm confused...

hankem: Okay, it apparently just took some time...

hankem, line 45: This needs to be adjusted for multi-dimensional arrays.

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTest.java
```diff
@@ -624,6 +624,31 @@ public void direct_dependencies_to_self_finds_correct_set_of_origin_types() {
...
+    @Test
+    public void stubs_have_empty_dependencies_and_are_not_imported() {
+        class Element {
+        }
+        class DependsOnArray {
+            Element[] array;
+        }
+        JavaClass directlyImportedClass = new ClassFileImporter().importClasses(DependsOnArray.class).get(DependsOnArray.class);
+        JavaClass additionallyImportedClass = directlyImportedClass.getField("array").getRawType();  // Element[]
+        JavaClass stub = additionallyImportedClass.getComponentType();  // Element
```
codecholeric, line 636: Does this mean that component types are not resolved automatically, even if `resolveMissingDependenciesFromClasspath` is set to `true`? Because if yes, then I think this should actually be fixed! (and maybe this test then written with deactivated resolution from classpath to easier get a stub class...)

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
```diff
@@ -960,6 +962,24 @@ public JavaConstructor getConstructor(String... parameters) {
...
+    /**
+     * Returns the transitive closure of all dependencies originating from this class, i.e. its direct dependencies
+     * and the dependencies from all target classes that are within the given package (including sub-packages).
+     * Note that all target classes in this package must be imported.
+     * @param packageNamePrefix start of the package name of classes whose transitive dependencies should be considered
+     * @return all transitive dependencies (including direct dependencies) from this class
+     * @see #getDirectDependenciesFromSelf()
+     */
+    @PublicAPI(usage = ACCESS)
+    public Set<Dependency> getTransitiveDependenciesFromSelfWithinPackage(final String packageNamePrefix) {
```
codecholeric, line 974: To me `WithinPackage` still seems like a rather arbitrary restriction :wink: What about two public methods

```
Set<Dependency> getTransitiveDependenciesFromSelf();
Set<Dependency> getTransitiveDependenciesFromSelf(DescribedPredicate<JavaClass> incudePredicate);
```
?

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTest.java
```diff
@@ -624,6 +624,31 @@ public void direct_dependencies_to_self_finds_correct_set_of_origin_types() {
...
+        assertThat(stub.getFieldsWithTypeOfSelf()).isEmpty();
+        assertThat(stub.getMethodsWithParameterTypeOfSelf()).isEmpty();
+        assertThat(stub.getMethodsWithReturnTypeOfSelf()).isEmpty();
+        assertThat(stub.getMethodThrowsDeclarationsWithTypeOfSelf()).isEmpty();
+        assertThat(stub.getConstructorsWithParameterTypeOfSelf()).isEmpty();
+        assertThat(stub.getConstructorsWithThrowsDeclarationTypeOfSelf()).isEmpty();
+        assertThat(stub.getAnnotationsWithTypeOfSelf()).isEmpty();
+        assertThat(directlyImportedClass.isImported()).isTrue();
+        assertThat(additionallyImportedClass.isImported()).isTrue();
+        assertThat(stub.isImported()).isFalse();
```
codecholeric, line 649: Maybe make this the first assertion? Would set the tone for me when I read it, i.e. "ah, it is not imported -> ah, now all the dependencies are empty" :wink:

archunit/src/test/java/com/tngtech/archunit/testutil/assertion/DependenciesAssertion.java
```diff
@@ -74,6 +75,19 @@ public DependenciesAssertion containOnly(Class<?> expectedOrigin, Class<?> expec
         return this;
     }
 
+    public DependenciesAssertion containOnlyTargetClasses(Class<?>... targetClasses) {
+        assertThat(actual)
+                .extracting(new Extractor<Dependency, Class<?>>() {
+                    @Override
+                    public Class<?> extract(Dependency dependency) {
+                        return dependency.getTargetClass().reflect();
```
codecholeric, line 83: Maybe it would be better to use `assertThat(javaClass).matches(reflectionClass)`? Because `reflect()` only goes for the name of the `JavaClass` anyway to create the `Class<?>`, no matter if the other attributes of the `JavaClass` matches. If you only want to go for the name (e.g. to not distinguish stubs and imported classes) I think it would be clearer to really only compare the names here then (because that makes it clear that only the names are relevant).

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTransitiveDependenciesTest.java
```diff
@@ -0,0 +1,148 @@
...
+                        .contain(d, e)
+                            .contain(e, Object.class)
+                            .contain(e, a)
+                            .contain(e, String.class)
+                .containOnlyTargetClasses(a, b, cArray, d, e, Object.class, Integer.class, String.class);
+
+        assertThatDependencies(classes.get(d).getTransitiveDependenciesFromSelfWithinPackage(PACKAGE))
+                .contain(d, Object.class)
+                .contain(d, e)
+                    .contain(e, Object.class)
```
codecholeric, line 135: I would imagine that one shortcut "format" would destroy all this nice indent? :wink: If yes, I think there are some "disable formatter" comments for IntelliJ. Just because I know that I repeatedly format everything like a robot, so it would be really annoying to me to always watch out for this place :wink:

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClassTransitiveDependencies.java
codecholeric: Yeah, I feel that `while(clazz.isArray()) { clazz = clazz.getComponentType() }` is a repeating task for me, but I don't know how it would fit elegantly into the domain API :disappointed:

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
```diff
@@ -1063,6 +1083,14 @@ public JavaConstructor getConstructor(String... parameters) {
         return javaClassDependencies.getAnnotationsWithTypeOfClass();
     }
 
+    /**
+     * @return Whether this class has actually been imported (or is just a stub).
+     */
+    @PublicAPI(usage = ACCESS)
+    public boolean isImported() {
```
codecholeric, line 1090: This will be a nice addition I think :smiley: And it will finally make it possible to identify non-stubs in a safe way.

codecholeric:
Also compare #253?

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTest.java
hankem: I'll check again on `resolveMissingDependenciesFromClasspath`, but I believe that my issue #397 was related to that.

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
hankem: To keep it simple & stupid, I'm actually inclined to drop the version with the `includePredicate` again.
I had used the package-restriction initially only to work around #397, but as this could be solved quite easily, the filtering is IMO not needed anymore.

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClassTransitiveDependencies.java
hankem: `JavaClass` could probably offer a method `getComponentBaseType()`?

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTest.java
hankem: Ah `resolveMissingDependenciesFromClasspath=true` was the default anyways. So no, component types don't seem to get resolved automatically.

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
hankem: I've updated the PR accordingly.

archunit/src/test/java/com/tngtech/archunit/testutil/assertion/DependenciesAssertion.java
hankem: When collecting all dependencies, `containOnlyTargetClasses` doesn't make sense anymore, because the result might depend on JRE implementation details.
I've therefore removed this function again.

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTransitiveDependenciesTest.java
hankem: I've now added a `// @formatter:off`, but it's up to you whether your IDE respects that or not. 😉

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClassTransitiveDependencies.java
hankem: PR for `getComponentBaseType()`: #441

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTest.java
codecholeric: I think this meanwhile is fixed, i.e. component types will be resolved transitively automatically

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTransitiveDependenciesTest.java
codecholeric: I like the formatting and my IDE respects it :wink:

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTest.java
hankem: Right. I had confused stubs with classes that are not fully imported.

PR #402, opened 2020-07-20
Removed unused java imports

dariuszzbyrad:
Removed a few unused java imports.


codecholeric:
Thanks a lot for keeping the house clean :smiley: I wasn't sure if unused imports would really slip through with only a couple of devs involved, but obviously they do. So I have also added the Gradle Spotless plugin to make sure the build stays free of unused imports in the future :+1:

PR #410, opened 2020-08-13
clarify javadocs that used `iff`

kamilszymanski:


hankem:
I think `iff` was meant as a short notation for [if and only if](https://en.wikipedia.org/wiki/If_and_only_if).

codecholeric:
Yes, it is, but maybe I'm too much from a math background :joy: But in general, should we not use this shortcut if it is confusing? But then I would write "if and only if" in the long form, cause that's what's meant...

hankem:
Wouldn't a simple `return whether $X` instead of `return true, iff $X` be good enough?
(To me, this sounds like refactoring the code expression `X ? true : false`.)

kamilszymanski:
`iff` meaning is not dead obvious - my first thought was `it's a typo`.
Due to the above I'd prefer more verbose `if and only if` if a simple `if` would not be clear enough.

I've updated the PR trying to follow local conventions.
Those conventions may catch your eye since there are at least a few conventions being used across the ArchUnit' JavaDocs.

codecholeric:
Hey, thanks a lot for adjusting it :smiley: Yes, I know, it's not always perfectly consistent :wink: I took the liberty to fix up the couple of places you touched now to use `{@code true}` consistently, so it's a least consistent within the respective class. And I think I prefer to have the code version in these cases :thinking: 
Anyway, thanks again! You might have saved others from being confused by the `iff` :wink: I'll try to refrain from using mathematically inspired expressions in the future...

PR #416, opened 2020-08-29
Gradle 6.6.1

sullis:
https://docs.gradle.org/6.6.1/release-notes.html


sullis:
WDYT? @codecholeric

codecholeric:
Thanks a lot :smiley: Always good to stay current with Gradle so we can avoid a painful upgrade like 4.x -> 6.x the last time :joy:

PR #417, opened 2020-09-02
Consider classes referenced in instanceof

t-h-e:
Resolves: #371

moved changes to a feature branch as before issues arose that the pull request got closed when rebasing.

original PR can be found here: https://github.com/TNG/ArchUnit/pull/383/
all comments have been addressed and marked as resolved

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/InstanceofCheck.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/InstanceofCheck.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2014-2020 TNG Technology Consulting GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.tngtech.archunit.core.domain;
+
+import com.tngtech.archunit.PublicAPI;
+import com.tngtech.archunit.core.domain.properties.HasOwner;
+import com.tngtech.archunit.core.domain.properties.HasType;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
+
+public final class InstanceofCheck implements HasType, HasOwner<JavaCodeUnit> {
+
+    private final JavaCodeUnit owner;
+    private final JavaClass target;
+    private final int lineNumber;
+
+    private InstanceofCheck(JavaCodeUnit owner, JavaClass target, int lineNumber) {
+        this.owner = checkNotNull(owner);
+        this.target = checkNotNull(target);
+        this.lineNumber = lineNumber;
+    }
+
+    @Override
+    @PublicAPI(usage = ACCESS)
+    public JavaClass getRawType() {
+        return target;
+    }
+
+    @Override
+    @PublicAPI(usage = ACCESS)
+    public JavaType getType() {
+        return target;
+    }
+
+    @Override
+    @PublicAPI(usage = ACCESS)
+    public JavaCodeUnit getOwner() {
+        return owner;
+    }
+
+    @PublicAPI(usage = ACCESS)
+    public int getLineNumber() {
+        return lineNumber;
+    }
+
+    static InstanceofCheck from(JavaCodeUnit owner, JavaClass target, int lineNumber) {
+        return new InstanceofCheck(owner, target, lineNumber);
+    }
+}
--- a/archunit/src/test/java/com/tngtech/archunit/core/domain/DependencyTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/core/domain/DependencyTest.java
@@ -6,6 +6,8 @@ import java.lang.annotation.RetentionPolicy;
 
 import com.google.common.base.MoreObjects;
 import com.tngtech.archunit.base.DescribedPredicate;
+import com.tngtech.archunit.core.domain.testobjects.ClassWithDependencyOnInstanceofCheck;
+import com.tngtech.archunit.core.domain.testobjects.ClassWithDependencyOnInstanceofCheck.InstanceOfCheckTarget;
 import com.tngtech.archunit.testutil.Assertions;
 import com.tngtech.archunit.testutil.assertion.DependencyAssertion;
 import com.tngtech.java.junit.dataprovider.DataProvider;
@@ -25,6 +27,8 @@ import static com.tngtech.archunit.core.domain.TestUtils.importClassesWithContex
 import static com.tngtech.archunit.core.domain.TestUtils.simulateCall;
 import static com.tngtech.archunit.testutil.Assertions.assertThat;
 import static com.tngtech.archunit.testutil.Assertions.assertThatType;
+import static com.tngtech.java.junit.dataprovider.DataProviders.$;
+import static com.tngtech.java.junit.dataprovider.DataProviders.$$;
 import static com.tngtech.java.junit.dataprovider.DataProviders.testForEach;
 
 @RunWith(DataProviderRunner.class)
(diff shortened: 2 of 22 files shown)


archunit/src/main/java/com/tngtech/archunit/core/importer/TargetAtLine.java
```diff
@@ -0,0 +1,42 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.tngtech.archunit.core.importer;
+
+import com.tngtech.archunit.core.domain.JavaType;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+class TargetAtLine {
```
codecholeric, line 22: This class looks very similar to `RawInstanceofCheck` except that the type could not be resolved yet. If you look at `RawAccessRecord` and similar classes, then your `TargetAtLine` would actually be `RawInstanceofCheck` in this terminology and your `RawInstanceofCheck` would actually be `ProcessedInstanceofCheck` (because it has been enriched with the correct info from the `ImportedClasses`).
What about actually calling this class `RawInstanceofCheck` and give it a `process(ImportedClasses classes)` method to resolve the `JavaClass` for `target`? The result could just be an nested class `Processed` (which would then naturally reside in the `importer` package and thus solve my comment above). When traversing the border from the `importer` package to the `domain` package we could then just pass `JavaClass target` and `int lineNumber` as separate parameters and the `domain` package would not be influenced by `RawInstanceofCheck` and `RawInstancecofCheck.Processed`. What do you think?

I.e.

```
class RawInstanceofCheck {
    private final JavaType target;
    private final int lineNumber;

    // ...

    Processed process(ImportedClasses classes) {
        // convert JavaType -> JavaClass via classes.get(..)
    }

    // ...

    static class Processed {
        private final JavaClass target;
        private final int lineNumber;
    }
}
```

archunit/src/test/java/com/tngtech/archunit/core/domain/DependencyTest.java
```diff
@@ -71,6 +71,31 @@ public void Dependency_from_throws_declaration() {
                 .contains("Method <" + origin.getFullName() + "> throws type <" + IOException.class.getName() + ">");
     }
 
+    @Test
+    public void Dependency_from_instanceof_check_in_method() {
```
codecholeric, line 75: I am now missing a test for an `instanceof` check in a constructor, since you have meanwhile removed `InstanceofCheckTest` (which I think is good, because as you noticed there are other tests that cover the same aspects). Or have I missed it somehow?
Also you could consider making this a `@DataProvider` test, where one `JavaConstructor`, one `JavaMethod` and one `JavaStaticInitializer` are passed in from which you expect the same behavior (i.e. each of them doing the same `instanceof` check so the assertion can be uniform). (might need to parameterize the line number as well though :wink:)

archunit/src/test/java/com/tngtech/archunit/core/domain/DependencyTest.java
```diff
@@ -71,6 +71,31 @@ public void Dependency_from_throws_declaration() {
...
+        assertThat(dependency.getOriginClass()).matches(ClassWithDependencyOnInstanceofCheck.class);
+        assertThat(dependency.getTargetClass()).matches(SomeMemberType.class);
+        assertThat(dependency.getDescription()).as("description")
+                .contains("Method <" + origin.getFullName() + "> checks instanceof <" + SomeMemberType.class.getName() + ">" + " in (DependencyTest.java:2"); // do not check exact location as line will move if file changes
+        assertThat(dependency.getSourceCodeLocation()).isEqualTo(origin.getSourceCodeLocation());
+    }
+
+    @Test
+    public void Dependency_from_instanceof_check_in_static_initializer() {
+        JavaClasses javaClasses = importClassesWithContext(ClassWithDependencyOnInstanceofCheck.class, SomeMemberType.class);
```
codecholeric, line 91: Shouldn't this test actually use `ClassWithDependencyInStaticInitializerOnInstanceofCheck`? Somehow that gives me the feeling that this test was not written TDD :stuck_out_tongue_closed_eyes: Also why has the test a different format than the test before? Shouldn't the assertions be uniform? (see `@DataProvider`)

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTest.java
```diff
@@ -439,7 +439,11 @@ public void direct_dependencies_from_self_by_member_declarations() {
                 .areAtLeast(2, parameterTypeDependency()
                         .from(AhavingMembersOfTypeB.class)
                         .to(B.class)
-                        .inLineNumber(0));
+                        .inLineNumber(0))
+                .areAtLeastOne(methodChecksInstanceOfDependency()
```
codecholeric, line 443: I like it that you added it here :+1: :smiley: 
I wonder if we could test `>1` dependency in this place, since this assertion would fit in well here, and is missing otherwise?

codecholeric (changes requested):
Thanks a lot for looking into it and improving it so nicely :smiley: I think it's cool that you can now see the line number and that it works for static initializers as well :+1: 
I added some points that I think we should address before merging this into `master` but featurewise it seems to be working well :smiley: 
I added an `example`, since you said you had troubles with it. Unfortunately it's not completely trivial, I agree, because from a bigger distance the complexity to match all relevant details in a robust way is also a lot bigger. And the examples are written in a way that they also influence each other (somehow I thought having one unified example project would make it more "realistic", but I'm not completely sure anymore :wink:)
I also added one little refactoring commit, before your changes start, because I thought it is kind of weird to use `SourceCodeLocation` for `Dependency` for the description, but then not use the same source to derive the line number but hard-code 0.
From my point of view once all points are fixed we should squash all your commits together into one, because I think it is one unit. I think when I look at it in half a year I'm only interested in the changes that went together to introduce the feature `InstanceofCheck`, but I'm not interested that line number was added later or that there was an exception added to the arch tests and later removed again :wink:

archunit/src/main/java/com/tngtech/archunit/core/domain/InstanceofCheck.java
```diff
@@ -0,0 +1,84 @@
...
+    @PublicAPI(usage = ACCESS)
+    public int getLineNumber() {
+        return lineNumber;
+    }
+
+    static InstanceofCheck from(JavaCodeUnit owner, RawInstanceofCheck target) {
+        return new InstanceofCheck(owner, target);
+    }
+
+    @PublicAPI(usage = ACCESS)
```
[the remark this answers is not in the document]
t-h-e: remove RawInstanceofCheck here

archunit/src/main/java/com/tngtech/archunit/core/importer/TargetAtLine.java
t-h-e: Good point. Thanks. 😃 
I removed to old `RawInstanceofCheck` and renamed `TargetAtLine` to `RawInstanceofCheck`. The inner class `Processed` does not seem to be necessary anymore.

archunit/src/test/java/com/tngtech/archunit/core/domain/DependencyTest.java
t-h-e: Done. The `@DataProvider` is really nice 😃 
I had to extract the line number per regex from the `SourceCodeLocation` as there is no other way to get it from a `JavaCodeUnit`.

archunit/src/test/java/com/tngtech/archunit/core/domain/DependencyTest.java
t-h-e: integrated this test in the previous one 😃

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTest.java
t-h-e: thx
What do you mean with "test `>1` dependency"?

codecholeric: I mean more than one dependency. Because it seems like everywhere we're only checking if it works with one single dependency? In other words, if I would refactor it and only keep the last occurrence of an `instanceof` dependency I would expect all tests to pass, because we never check for more than one `instanceof` dependency anywhere. Or did I miss a place where this is already tested?

t-h-e: ah, now I get it.
In contrast to the other dependency checks, the instanceof dependency contains a line number, which is also checked by the test. So to have a test that checks that there are at least 2 instanceof check dependencies, you would need to have the same dependency in the same line for the same classes. This seems weird and also because the dependency would be identical according to the `hashcode` and `equals` method, the dependency will only appear once in the `Set` returned from `getDirectDependenciesFromSelf`.
Another option would be to add a method `ignoringLineNumber` in `DependencyConditionCreation` to check dependencies independent of the line number.
I think due to checking the line number, the test is fine.

codecholeric: Ah, yeah, you're right of course :joy: But you could have two checks of `.areAtLeastOne(methodChecksInstanceOfDependency()` including a different line number, right? That would also mean there is more than one `instanceof` dependency found...

t-h-e: Done 😄

codecholeric (approved):
Thanks a lot :smile: Super nice now :+1: 
I have adjusted the unit test a little, because it was a little to much magic for me to extract the expected line number programmatically (with some comment then cause it's obviously not trivial how to derive it :wink:). I usually try to keep the expectation directly specified, because otherwise you almost need a test for the test again... But to keep the test somewhat stable I put the class input into a separate file, otherwise this test will fail whenever any line is added or removed from the test (which I expect was the original intention to derive the expected line number programmatically).
Anyway, gonna merge it now, thanks again for making ArchUnit better!!

t-h-e:
Thanks for all your help.
Yes the magic with the line numbers wasn't really pretty. Moving the class to its own file is definitely a nicer solution 😃

PR #426, opened 2020-09-25
Add workflow for Update Gradle Wrapper Action

cristiangreco:
Hey there 👋, first of all thanks for your work on ArchUnit!

I've got a suggested change: would you be willing to use [this GitHub Action](https://github.com/gradle-update/update-gradle-wrapper-action) to automatically keep Gradle Wrapper updated to latest release?

**What does "Update Gradle Wrapper Action" do?** It can be configured to run at scheduled intervals (e.g. daily or weekly) and will check whether the Wrapper script in the repo is up-to-date to the latest Gradle release: in case a new Gradle version is available, it will create a PR to update the Wrapper. And that's it!

**Why is that a good thing?** Well, first of all it alleviates the chore of manually updating the Wrapper, as you got a task that keeps track of new Gradle releases for you! More importantly, it boosts security around the Wrapper update and usage processes: this actions verifies that the `gradle-wrapper.jar` file has not been tampered with (uses checksum comparison), and it sets the `distributionSha256Sum` property so that the new Gradle binary itself will be verified locally upon download.

**Where can I find more about?** The [README](https://github.com/gradle-update/update-gradle-wrapper-action/blob/master/README.md) contains quite detailed information!

In this PR I propose adding a new workflow which runs the action every day at midnight (but feel free to adjust the frequency as you prefer). Given ArchUnit is already up-to-date to Gradle 6.6.1 the action won't basically do anything until a new version is out (great work keeping it updated so regularly!).

The action is under active development, you can have a look at the [list of inputs](https://github.com/gradle-update/update-gradle-wrapper-action#action-inputs) currently supported. There's new features coming up soon and if you'd like to request any particular change just let me know!

I'd love to see the action used by ArchUnit as such a popular project would definitely help me spread the word, but I also genuinely hope you can find this useful. Would love your feedback! ❤️

### DIFF
--- a/.github/workflows/update-gradle-wrapper.yml
+++ b/.github/workflows/update-gradle-wrapper.yml
@@ -1,0 +1,17 @@
+name: Update Gradle Wrapper
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  update-gradle-wrapper:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Update Gradle Wrapper
+        uses: gradle-update/update-gradle-wrapper-action@v1
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}


codecholeric:
Thanks a lot for the initiative! :smiley: This looks like a really useful addition. To be honest we haven't always been that up to date with our Gradle version, so having an automatic push "here is a new version -> and you can just merge it, cause everything is green" would definitely be nice (and of course also "it's red, be aware that you need to do some work" :wink:)
I just have some questions before going forward:

* on the docs I saw `Note that, when using the GITHUB_TOKEN, Pull Requests created by the Update Gradle Wrapper action cannot trigger any workflow run in your repository`. So do I understand it correctly that if I merge your PR as it is, our CI workflows wouldn't be run on future update-gradle-wrapper PRs? Because that's something I definitely would want to immediately see that all build steps with the updated Wrapper would pass
* why does your action need the `GITHUB_TOKEN` at all, while things like Dependabot or Renovate can do without?
* I would imagine the action would need as it's only permission to create PRs in this repo? How would I configure access to allow this and only this? I feel that using a PAT I could only enable a generic read and write on the repo with all permissions I myself have? That doesn't sound like the right thing to do :thinking:

cristiangreco:
Hey @codecholeric thanks for taking the time to review the PR and the action! 🙌

Trying to answer your questions:

> on the docs I saw Note that, when using the GITHUB_TOKEN, Pull Requests created by the Update Gradle Wrapper action cannot trigger any workflow run in your repository. So do I understand it correctly that if I merge your PR as it is, our CI workflows wouldn't be run on future update-gradle-wrapper PRs? Because that's something I definitely would want to immediately see that all build steps with the updated Wrapper would pass

So update-gradle-wrapper will create a PR for you to review and merge, and yeah would be great to run CI workflows as usual in such PRs right? Unfortunately, if you use GitHub actions for CI checks, these are not automatically triggered: this is a limitation imposed by GitHub to prevent creating loops in your workflows. 

A workaround recommended by GitHub itself is to use a PAT, but this has the downsides you mention (see later for a complete asnwer). Another possible alternative is to **manually close and immediately re-open** the PR: it's a bit of a shame to have to do a manual steps in an automatic workflow but 🤷‍♂️ this is the easiest solution. 

> why does your action need the GITHUB_TOKEN at all, while things like Dependabot or Renovate can do without?

Dependabot and Renovate are Apps rather than Actions: apps get their access token the moment you authorise them, while Actions need to use a dedicated token at every run. `GITHUB_TOKEN` is a special secret that GitHub installs automatically for use with Actions, you can read more [here](https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow).

> I would imagine the action would need as it's only permission to create PRs in this repo? How would I configure access to allow this and only this? I feel that using a PAT I could only enable a generic read and write on the repo with all permissions I myself have? That doesn't sound like the right thing to do 🤔

The PAT to be used instead of `GITHUB_TOKEN` will need `repo` scope, like this:

pat

The security concern around using a PAT is that it cannot be scoped to a specific repository. That is not a big problem per-se, since you would just need to generate and immediately store it as an [encrypted secret](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets).

While this is the solution [recommended by GitHub](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token), I understand your concerns. As mentioned above, one easy alternative is to just close/re-open the PR manually with your account. I'm testing another approach: using Deploy Keys (which only have access to a particular repo) instead of PAT, but this is still WIP: if you want to wait until this is implemented just hold off using the action, otherwise I'd recommend testing one of the two approaches described above (either manual intervention or creating a PAT).

codecholeric:
Hey, thanks for elaborating that for me :smiley: I think I understand now. It's okay for me to go with the manual step of closing and reopening the PR for now, the benefit still outweighs this for sure :wink: 
The only thing is, that I tried to bring your PR up to date with `master` (i.e. rebase and force push, because we always strive to keep a linear history), but I was not allowed to push to your branch :wink: 
Would if be okay for you to allow me to force push, so I can update the repo?

cristiangreco:
@codecholeric thanks for your reply! I quickly force-pushed myself, so now the PR should be up-to-date with your recent changes.

codecholeric:
Hey, since the branch was outdated again and I want to avoid too many communication cycles, I've recreated the PR freshly based on master. Thanks again for the initiative!! :smiley: I'll close this in favor of #486.

PR #427, opened 2020-09-25
Run task archunit:jdk9Test during multi-jdk-matrix-build

rweisleder:
The default CI build `gradlew build` didn't execute the task `jdk9Test` and therefore the tests located in src/jdk9test/java are not executed regularly.

### DIFF
--- a/build-steps/testing/testing.gradle
+++ b/build-steps/testing/testing.gradle
@@ -55,7 +55,6 @@ ext {
         testTask.enabled = false
 
         def taskMinimumJdkVersion = testTask.name.replaceAll(/jdk(\d+)Test/, '$1').with { it ==~ /\d+/ ? JavaVersion.toVersion(it) : proj.targetCompatibility }
-        def additionalJdksToTest = testJdks().findAll { it.javaVersion >= taskMinimumJdkVersion }
 
         def findJavaExecutable = { jdkPath ->
             def javaExecutableUnix = new File("${jdkPath}", 'bin/java')
@@ -65,13 +64,14 @@ ext {
             result
         }
 
-        additionalJdksToTest.each { jdk ->
+        testJdks().each { jdk ->
             def additionalTestTask = proj.tasks.create(name: "${testTask.name}${jdk.suffix}", type: Test) {
                 executable = findJavaExecutable(jdk.jdkPath)
 
                 testClassesDirs = testTask.testClassesDirs
                 classpath = testTask.classpath
             }
+            additionalTestTask.enabled = jdk.javaVersion >= taskMinimumJdkVersion
 
             testTask.dependsOn(additionalTestTask)
         }
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,7 +72,7 @@ jobs:
         env:
           JAVA_HOME: ${{ env.build_jdk_path }}
         with:
-          arguments: testJre${{ matrix.test_java_version }} -PallTests -PmultiJdkTest -Pjava${{ matrix.test_java_version }}Home="${{ env.test_jdk_path }}"
+          arguments: testJre${{ matrix.test_java_version }} jdk9TestJre${{ matrix.test_java_version }} -PallTests -PmultiJdkTest -Pjava${{ matrix.test_java_version }}Home="${{ env.test_jdk_path }}"
 
   integration-test:
     strategy:


codecholeric:
Thanks a lot!! I have noticed parts of that a little while ago already, but unfortunately I made it part of my ongoing PR so it hasn't been merged yet. I'll merge my old #429 and then rebase your PR on top, because I haven't looked into the CI part so far!

codecholeric:
So what is the state about the line endings? I thought your change looked quite reasonable to specify it in the `.gitattributes`, what was the problem with GitHub's Windows runner? On CI it then failed on SpotBugs?

build-steps/testing/testing.gradle
```diff
@@ -65,13 +64,14 @@ ext {
             result
         }
 
-        additionalJdksToTest.each { jdk ->
+        testJdks().each { jdk ->
```
codecholeric, line 67: I guess you did this change to not fail when invoking `testJdk9Jre7/8`? If yes, it would be super nice to add it to the commit message :smiley: (I mean, why was this necessary and what did it solve)

codecholeric (changes requested):

rweisleder:
> So what is the state about the line endings? I thought your change looked quite reasonable to specify it in the `.gitattributes`, what was the problem with GitHub's Windows runner? On CI it then failed on SpotBugs?

On my machine, the current master fails on `gradlew build` with:
```
Execution failed for task ':archunit:spotlessJavaCheck'.
> The following files had format violations:
      archunit\src\jdk9main\java\com\tngtech\archunit\core\domain\Java9DomainPlugin.java
          @@ -1,36 +1,36 @@
          -/*\n
          - * Copyright 2014-2020 TNG Technology Consulting GmbH\n
          - *\n
...
          +/*\r\n
          + * Copyright 2014-2020 TNG Technology Consulting GmbH\r\n
          + *\r\n
...
      ... (24 more lines that didn't fit)
  Violations also present in 662 other files.
  Run 'gradlew.bat :archunit:spotlessApply' to fix these violations.
```
Spotless wants to change the line endings of all files in my working copy from \n to \r\n.

I tried the change with the `.gitattributes` file. It solved the problem locally.
But then the CI build fails on Windows runners, see https://github.com/TNG/ArchUnit/runs/1166974355
```
Execution failed for task ':archunit-junit4:spotlessJavaCheck'.
> The following files had format violations:
      archunit-junit\junit4\src\main\java\com\tngtech\archunit\junit\AnalyzeClasses.java
          @@ -1,18 +1,18 @@
          -/*\r\n
          - * Copyright 2014-2020 TNG Technology Consulting GmbH\r\n
          - *\r\n
...
          +/*\n
          + * Copyright 2014-2020 TNG Technology Consulting GmbH\n
          + *\n
...
```
Spotless wants to change the line endings from \r\n to \n.

I'm not quite sure why the working copy on the Windows runner has CRLF line endings even if the `.gitattributes` file is present with
```
*.java text eol=lf
```
Locally, I have `core.autocrlf=false`. Maybe the Windows runner has `core.autocrlf=true` and this overrides the `.gitattributes` file? There is also an open issue for this topic: https://github.com/actions/checkout/issues/135

Also, there seems to be no option for Spotless to simply ignore different line endings.

codecholeric:
Hmm, that's too bad. I think your guess is right, from the issue you linked I understood that `autocrlf=true` by default and on the other hand that overrides `.gitattributes` -> https://git-scm.com/docs/gitattributes#_text
Just don't know how to best go forward here to have a good experience for everyone, cause we also shouldn't break it for people using `autocrlf=true` :thinking:

rweisleder:
It looks like we can't resolve this quickly. I extracted this as new issue: #437. If it only affects me, then I can live with it. 😅

codecholeric:
Event though I think it's a little weird to have a task called `jdk9TestJre8` :joy: I guess it's the easiest to keep the CI scripts simple (which I feel I already had enough trouble with :wink: )
I mean, it would be possible to check in the CI workflow if the version is >= 9 and only then set an `env.additionalTestTask = 'jdk9TestJrex'` which could then be added to the command via 

```
testJre${{ matrix.test_java_version }} ${{ env.additionalTestTask }} -PallTests
```

But I feel the complexity there is worse and it's harder to test. So I think we should go with simply having a disabled task :+1:

codecholeric (approved):

codecholeric:
Thanks a lot again :smiley: Hope your other issue can be fixed somehow!

PR #428, opened 2020-09-25
Add component type of arrays to JavaClass dependencies #257

### LINKED ISSUE #257: Add component type to JavaClass dependencies
Now that #244 has been merged, we should add the component type of an array as a dependency to `JavaClass`.
For example

```
class Example {
    Dependency[] dependencies;
}
```

At the moment this would count `Example` to have a dependency on `Dependency[]`, but it would not yet report a dependency on `Dependency`, even though that is obviously present as well.

Note that in reality this is not a huge issue, because if no other dependency on `Dependency` arises anywhere, it pretty much means that we only have the array, but never actually do anything with the elements of type `Dependency`, thus the dependency is a very lose one.

Nevertheless, now that we have easy access to the component type, it should be well possible to extend the dependencies to cover all component types of all array type depedencies.

wengertj:
..damnit - wrong issue in the branch name 🤣 

## Summary

This PR partly resolves #257 by adding the component type of arrays to the dependencies *from self* in `JavaClass`. The description of the `Dependency` currently reads as follows:
```
 Field <com.tngtech.archunit.example.layers.service.ServiceType.$VALUES> has type <com.tngtech.archunit.example.layers.service.ServiceType> (component type) in (ServiceType.java:0)
```

## Open Points/Questions
* Does the description format work for you?
* Did I miss anything important?
* Are my tests thorough enough?
* ⚠️  **Help wanted** ⚠️  : For some reason the `ExamplesIntegrationTest#FrozenRulesTest` fails with weird (encoding?) issues. Running each of `FrozenRulesTest` individually works perfectly fine 😕 

Thank you very much for looking at this and have a nice weekend 😃


wengertj:
Update: I've rebased my branch and resolved the conflicts introduced by 56feedca93236738056c455fffa1f8cad2dd329f and 703f2f2a1912d0415a7207269bb5e1e0a4072770.

Still can't figure out how to fix `ExamplesIntegrationTest#FrozenRulesTest` I'm afraid...

wengertj:
Careful ping to @codecholeric 🙈 

If you could give me a little hint with regards to the failing `ExamplesIntegrationTest#FrozenRulesTest` I'd be thrilled 😀 

But no worries 😉

hankem:
`ExamplesIntegrationTest#FrozenRulesTest` verifies that 
* `FrozenRulesTest.no_classes_should_depend_on_service` reports violations including
  ```
  Method <com.tngtech.archunit.example.layers.controller.SomeController.doSthController()> calls method <com.tngtech.archunit.example.layers.service.ServiceViolatingDaoRules.doSthService()> in (SomeController.java:11)
  Method <com.tngtech.archunit.example.layers.controller.SomeController.doSthWithSecuredService()> calls method <com.tngtech.archunit.example.layers.service.ServiceViolatingLayerRules.properlySecured()> in (SomeController.java:15)
  ```
* `FrozenRulesTest.no_classes_should_use_the_EntityManager` reports the violations
  ```
  Method <com.tngtech.archunit.example.layers.service.ServiceViolatingDaoRules.illegallyUseEntityManager()> calls method <com.tngtech.archunit.example.layers.service.ServiceViolatingDaoRules$MyEntityManager.persist(java.lang.Object)> in (ServiceViolatingDaoRules.java:27)
  Method <com.tngtech.archunit.example.layers.service.ServiceViolatingDaoRules.illegallyUseEntityManager()> calls method <javax.persistence.EntityManager.persist(java.lang.Object)> in (ServiceViolatingDaoRules.java:26)
  ```
You had, however, added these violations to the `ViolationStore` in `archunit-example/example-plain/src/test/resources/frozen`, which is why they are not reported anymore, causing `ExamplesIntegrationTest#FrozenRulesTest` to fail.

For your PR, it would have been sufficient to freeze the new violations
```
Field <com.tngtech.archunit.example.layers.service.ServiceType.$VALUES> has type <com.tngtech.archunit.example.layers.service.ServiceType> (component type) in (ServiceType.java:0)
Method <com.tngtech.archunit.example.layers.service.ServiceType.values()> has return type <com.tngtech.archunit.example.layers.service.ServiceType> (component type) in (ServiceType.java:0)
```

hankem:
I've added one commit that did the trick.
Feel free to squash it into yours using `get fetch && git rebase -i --autosquash origin/master`.

codecholeric:
@wengertj sorry, lost a little bit track here :disappointed: Thanks a lot @hankem for jumping in :+1: :smiley: 
Guess now I just need to find the time to review it :wink: I'll try to get around to it at the weekend!!

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTest.java
```diff
@@ -1081,7 +1125,17 @@ Step3 to(Class<?> target, String targetName) {
 
                 Step3(Class<?> target) {
                     this.target = target;
-                    targetDescription = target.getSimpleName();
+                    targetDescription = createTargetDescription(target);
+                }
+
+                // NOTE: we need to distinguish here otherwise the description
+                // will produce an illegal regex pattern an cause a PatternSyntaxException
```
hankem, line 1132:
```suggestion
                // will produce an illegal regex pattern and cause a PatternSyntaxException
```

archunit/src/test/java/com/tngtech/archunit/core/domain/DependencyTest.java
```diff
@@ -68,7 +104,7 @@ public void Dependency_from_throws_declaration() {
                 .get(ClassWithDependencyOnThrowable.class).getMethod("method");
         ThrowsDeclaration<JavaMethod> throwsDeclaration = getOnlyElement(origin.getThrowsClause());
 
-        Dependency dependency = Dependency.tryCreateFromThrowsDeclaration(throwsDeclaration).get();
+        Dependency dependency = Dependency.tryCreateFromThrowsDeclaration(throwsDeclaration).iterator().next();
```
hankem, line 107: I'd probably use _`getOnlyElement`_ from `com.google.common.collect.Iterables`, wihch also ensures that there is only one element.

codecholeric (changes requested):
Thanks a lot for your patience! I've finally managed to look into it. I have added two little review commits that you can squash onto yours, if you feel they make sense!
Also I'm a little unsure about one particular case, namely

```
void what_about_these_cases() {
    // I'm not sure about this case. Should this cause a dependency on `File`?
    File[] files = new File[0];
    // Likewise this here, it's a call of `File[].clone()`, but should it still also be a dependency on `File`?
    files.clone();
}
```

I'm genuinely not completely sure how those should be treated :thinking: I slightly tend to also create a dependency to the component type there, what do you think? But it's not covered at the moment somehow, as far as I could tell? :thinking: 

Also I'm not super sure about the description (the brackets `(component type)`, since this feels inconsistent to how we normally do it. I was wondering anyway, if we need the original description in this case at all :thinking: 
Wouldn't it be okay to simply use something like

```
Field <some.Foo.bar> depends on component type <java.lang.String> in (Foo.java:0)
```

And we always use the same description for all component types, no matter, if it's a field or a method parameter? (or a method call target, if we add this? Because if the description would be `Method <some.Foo.bar()> calls method <other.Baz.quux()> (component type)`, that would feel even weirder to me :thinking: )

wengertj:
Thanks a lot for your Feedback 👍 

I agree with you on all Points and will address them. 
- [ ] Change the description format
- [ ] Consider dependencies from access
- [ ] Consider new Array creation

I'm  not  entirely sure about the last one though - that might be more work: As far as I can understand it a call to `new Object[]` is a different bytecode instruction than calling a constructor: [ANEWARRAY](https://asm.ow2.io/javadoc/org/objectweb/asm/Opcodes.html#ANEWARRAY) which does not seem to be handled in `JavaClassProcessor` 🤔 
But I am _way_ out of my element here so I might be very wrong.

codecholeric:
Ah, I think you're right, I forgot that `new SomeType[]` is a separate bytecode instruction that we don't handle at the moment :joy: No worries then, I think calls of methods are good enough. If somebody should ever complain, we can still add it, I think I know where to hook into the ASM visitor API...

Anyway, I have squashed it all together into one commit, the only thing I've adjusted is for it to work recursively. Because I realized at the moment we would have the same problem again for `Foo[][]`, then people would complain that there is no dependency on `Foo` :wink: 

So now the final behavior is dependency on `Foo[][][]` -> component type dependency on `Foo[][]`, `Foo[]` and `Foo`. I think that's consistent behavior.

So thanks again for your contribution, I really appreciate it!! :smile:

codecholeric (approved):

codecholeric:
The only thing I can already see popping up is confused people why `SomeEnum.values()` now also has component type dependencies even though you don't even see the method in the source code :wink:

wengertj:
Thanks a lot for merging and all the valuable feedback - I certainly learned a lot 😃

PR #438, opened 2020-09-27
Run tests against Java 15 during CI

rweisleder:
ArchUnit itself is compatible with Java 15. I only had to fix a single integration test to make the build executable with Java 15. But damn what you can learn about the JDK from a single failing test... 😅


archunit-example/example-junit4/src/test/resources/frozen/a81a2b54-5a18-4145-b544-7a580aba0425
```diff
@@ -20,4 +20,5 @@ Method <com.tngtech.archunit.example.layers.service.ComplexServiceAnnotation.ser
 Method <com.tngtech.archunit.example.layers.service.ComplexServiceAnnotation.simpleServiceAnnotation()> has return type <com.tngtech.archunit.example.layers.service.SimpleServiceAnnotation> in (ComplexServiceAnnotation.java:0)
 Method <com.tngtech.archunit.example.layers.service.ServiceType.values()> calls method <[Lcom.tngtech.archunit.example.layers.service.ServiceType;.clone()> in (ServiceType.java:3)
 Method <com.tngtech.archunit.example.layers.service.ServiceType.values()> has return type <[Lcom.tngtech.archunit.example.layers.service.ServiceType;> in (ServiceType.java:0)
+Method <com.tngtech.archunit.example.layers.service.ServiceType.$values()> has return type <[Lcom.tngtech.archunit.example.layers.service.ServiceType;> in (ServiceType.java:0)
```
hankem, line 23: When the tests are executed with a Java version <15, ArchUnit removes this line again.

rweisleder: Good catch! This I see two options here:
1. Set `freeze.store.default.allowStoreUpdate=false` if this helps.
2. Make the expected test failures in https://github.com/TNG/ArchUnit/blob/a86e9a5259023697fccba165bb10e308ea5ee159/archunit-integration-test/src/test/java/com/tngtech/archunit/integration/ExamplesIntegrationTest.java#L618-L626 depending on the JDK.
3. Add a comment to not remove this line or commit the removal. If I understand correctly, this would only happen during local development.

What would you recommend?

codecholeric: Hmm, I think 1. is unfortunately no solution, because that is more meant to prevent mistakes to keep the store up to date. E.g. you expect the current violations are correctly frozen in the store, but then in your CI environment suddenly there are some obsolete violations. Thus that throws an exception which would not be nice for this case.

Nr 3. I don't like so much either, because people might still overlook the comment :thinking: But if everything else takes a lot of effort it would still be okay for me. 2. sounds a little complex (wouldn't it be sth. like adding an extra line if JDK >= 15?).

What we could also do is to reconfigure the store location to use the one in the `build` folder. I.e. the resources plugin copies the frozen rules anyway to `build/resources/test` (could be found safely in the classpath by `Class.getResource("/frozen/stored.rules")`). Changes to files under `build` are ignored from a VCS point of view anyway. Wouldn't elaborate the idea very well, but the example could keep it as it is, while the integration test could reconfigure it. I would think that is a fair trade off. What do you think?

codecholeric:
Thanks for your support!! :smiley: I've patched up the freezing test fix with a version that just copies the store to a temp folder before running the test. I think that is still the cleanest given the circumstances (of course to be completely clean we would have to create one store per JDK where there are differences, but I don't really want to maintain that :wink:). I'll merge it now to finally get on with the PR about records (sorry that took me so long, too many parallel things :disappointed:)

PR #447, opened 2020-10-04
ow2 asm 9.0

sullis:
https://asm.ow2.io/versions.html

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ ext {
     googleRelocationPackage = "${thirdPartyRelocationPackage}.com.google"
 
     dependency = [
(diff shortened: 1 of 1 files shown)


sullis:
Replaced by https://github.com/TNG/ArchUnit/pull/466

codecholeric:
Yeah, thanks for the initiative, but I guess this time it was not such smooth sailing to upgrade :disappointed: The reason is that ASM this time has published some Gradle meta-data that claims it needs Java >= 8 (even though ASM in fact is compatible to older versions). I've opened an issue about it here: https://gitlab.ow2.org/asm/asm/-/issues/317920
Anyway, for the moment I found a workaround by tweaking the repo config.
Thanks for the PR anyway though! This way I got a push that there is finally a new non-beta major version :+1:

PR #453, opened 2020-10-19
Rules shared between tests have ambiguous names

### LINKED ISSUE #452: Rules shared between tests have ambiguous names
As mentioned in my other post #450, we have lots of "projects" that share a "platform", and we are restricting access to the "platform".

The "platform" provides some rules annotated as `@ArchTest`, e.g. "`only_use_allowed_platform_classes`" in a class "`platform.ApiUsageRules`". Each "project" declares a JUnit 4 test "`projectX.ApiUsageTest`", annotated to run the `ArchUnitRunner`, which tests the rules declared in `ApiUsageRules` via `ArchRules`.

Several "projects" are built together with Gradle on CircleCI. When one of the rules fails, the failing test is called "`only_use_allowed_platform_classes - platform.ApiUsageRules`", which the location where the rule was defined. That is of course the same for all "projects".

Most of the time, I could deduce from the violations which "project" is affected, but not always. From the build test output I could not see which test failed either.

Do you have suggestions how I can more easily find the actual test that failed?
Is this a problem due to our build setup, i.e. would I find the actual test class in the output or test artifacts of a simple/standard Gradle/Maven build?

I thought about adding the actual test class name ("`projectX.ApiUsageTest`") to the test name.

Kind regards,
Christian

csemrau:
Resolves #452


csemrau:
Superseded by #1279

PR #463, opened 2020-11-03
Add Optional.getOrThrow(Supplier) and replace deprecated method

### LINKED ISSUE #391: Avoid using Optional.getOrThrow(RuntimeException) in our code
Currently, we have some usages of `Optional.getOrThrow(RuntimeException)` in our code like this:
https://github.com/TNG/ArchUnit/blob/38f6748c4e5d5758cfb3c6509aefb616cf83e475/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java#L677-L679

The usage of this method requires to create a new `Throwable` object and fill the stack trace, whether the exception is needed or not. See also [The hidden performance costs of instantiating Throwables](http://normanmaurer.me/blog/2013/11/09/The-hidden-performance-costs-of-instantiating-Throwables/).

It may not have a big impact, but I think we should avoid the usage of this method in our code.

idosal:
The current implementation always forces the instantiation of an exception, which is unnecessary and expensive in most cases. Now there is a method `T getOrThrow(Supplier<? extends RuntimeException> exceptionSupplier)` which accepts a supplier that will only be invoked, if the exception really needs to be thrown (i.e. the `Optional` is absent even though it is assumed to be present).
Since we do not want Guava in the public API there is now a custom `Supplier` interface to be used as independent public API.

Resolves: #391

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/base/Optional.java
+++ b/archunit/src/main/java/com/tngtech/archunit/base/Optional.java
@@ -42,7 +42,6 @@ public abstract class Optional<T> {
     }
 
     @PublicAPI(usage = ACCESS)
-    @SuppressWarnings("unchecked")
     public static <T> Optional<T> absent() {
         return Absent.getInstance();
     }
@@ -53,8 +52,15 @@ public abstract class Optional<T> {
     @PublicAPI(usage = ACCESS)
     public abstract T get();
 
+    /**
+     * @deprecated Use {@link #getOrThrow(Supplier)} instead (this version always instantiates the exception, no matter if needed)
+     */
+    @Deprecated
     @PublicAPI(usage = ACCESS)
     public abstract T getOrThrow(RuntimeException e);
+
+    @PublicAPI(usage = ACCESS)
+    public abstract T getOrThrow(Supplier<? extends RuntimeException> exceptionSupplier);
 
     @PublicAPI(usage = ACCESS)
     public abstract <U> Optional<U> transform(Function<? super T, U> function);
@@ -115,6 +121,11 @@ public abstract class Optional<T> {
         }
 
         @Override
+        public T getOrThrow(Supplier<? extends RuntimeException> exceptionSupplier) {
+            throw exceptionSupplier.get();
+        }
+
+        @Override
         public <U> Optional<U> transform(Function<? super T, U> function) {
             return absent();
         }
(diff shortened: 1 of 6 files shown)


idosal:
Seeing as the `Supplier` here is Guava, should I just keep `getOrThrow`'s current public signature and wrap the `Throwable` with a `supplier` internally?

codecholeric:
Thanks a lot for your support!! :smiley: Yeah, as you noted putting Guava in public API is not something we want to do :thinking: 
I have already introduced a couple of custom copies of standard Java 8 classes, like `Optional`, `Predicate` or `Function`.
So if we want to make this a public API we need to also add our own `com.tngtech.archunit.base.Supplier`.
I think the days of these classes are numbered, because sooner or later we will switch to Java 8 and then finally standardize a lot of these concepts. But for now I still think it might be worth it to have an immediate benefit.
So what do you think? Wanna replace the usage of the Guava supplier with a custom one?

idosal:
> Thanks a lot for your support!! 😃 Yeah, as you noted putting Guava in public API is not something we want to do 🤔
> I have already introduced a couple of custom copies of standard Java 8 classes, like `Optional`, `Predicate` or `Function`.
> So if we want to make this a public API we need to also add our own `com.tngtech.archunit.base.Supplier`.
> I think the days of these classes are numbered, because sooner or later we will switch to Java 8 and then finally standardize a lot of these concepts. But for now I still think it might be worth it to have an immediate benefit.
> So what do you think? Wanna replace the usage of the Guava supplier with a custom one?

Sure. :)

codecholeric:
Hey thanks again :smiley: I have added some small review commits. If they make sense I would suggest we squash it all together into one commit ("make `Optional.getOrThrow(..)` accept supplier instead of actual exception" or something like that) and then merge it!

idosal:
@codecholeric Thanks for that. :) I'm sorry for the delay, I just had a lot on my plate lately. I'll do my best to introduce a new `Supplier` in place of Guava and merge as soon as possible.

idosal:
@codecholeric Added the `Supplier`. This is good for Squash and Merge as far as I'm concerned. :)

codecholeric:
Thanks a lot for helping to make ArchUnit better!! :smiley: I've decided to remove the internal usages of the API, as long as ArchUnit uses Java 7, because it just does not make any sense from my point of view as long as we have to use anonymous classes instead of lambdas. But it is a good API to have for all users that are on Java >= 8 (which is almost all :wink:)

Issue #478, opened 2020-11-26
[website] add link to c# version 


gernotstarke:
I get questions like "is there a c# version?" or "is there a c++ version"? quite often - but the archunit.org website does currently not answer these...

Shall we put that in the user guide or on the home page? Will create a PR if we decide where to put the info :-)

PR #480, opened 2020-11-26
propose fix for #478 (add link to C# version)

gernotstarke:


gernotstarke:
I simply added the link to the C# version on the homepage...

codecholeric:
Thanks a lot for the idea! I also think this is a good addition, since I get these questions, too :wink: 
I just wonder, if it's good to write it in the introductory sentence ("checking the architecture of your Java or C# code"), because ArchUnitNET is really like a different world and it's not described on the website or in the user guide at all. 
So I wonder if it wouldn't be better to make it clear in a separate sentence, like "There also exists a port for .NET/C#, which you can find [here]", linking to ArchUnitNET's site (i.e. https://archunitnet.readthedocs.io/en/latest/ ). 
What do you think? Maybe we even wouldn't need the GitHub link then, because the ArchUnitNET docs link to the GitHub repo if anyone is interested to contribute or open issues :thinking:

gernotstarke:
yes, sounds better. Will close THIS PR and open another...

PR #482, opened 2020-12-01
add simple use cases page to ArchUnit website

codecholeric:
The use case examples featuring the PlantUML diagrams are quite hidden within the user guide, even though they show a core motivation in which cases to use ArchUnit.
To make them more prominent I have included a simple page on the website, which is auto-generated from the respective Asciidoc in the user guide. It is quite hacky, but given that this part does not change that often it might be good enough.
The way it works at the moment is to split on headings of type '===', assume the first line is then the heading of the use case. Then split the section on source code snippet markers (`[plantuml,`/`[source,`), assume that the first part after `----` is the code snippet and assume there is always one PlantUML snippet followed by one Java source code snippet.
In the end the snippets are joined and the PlantUML is converted to an URL for the PlantUML server via simple HEX encoding (GitHub IMHO does not support PlantUML in markdown, but via PlantUML server we can just include the encoded PlantUML snippet as a SVG image URL (compare https://plantuml.com/text-encoding). We still need to check in the use case markdown, but at least when the user guide is recreated and the use cases change then the page will be updated and will be shown as changed file.


gernotstarke:
good enough for me :-)
we don't need to care for maintainability here... let's get this live!

codecholeric:
Done :tada: Thanks for the initiative @gernotstarke :smiley: 
I'm gonna release within the next couple of days, then I'll update the website to the newest master state on that occasion.

gernotstarke:
Bowing to your producitivy :-)

PR #488, opened 2020-12-06
slf4j 1.7.30

sullis:
slf4j 1.7.30 was released in December 2019

https://github.com/qos-ch/slf4j/releases


codecholeric:
Thanks for the update!! :+1:

PR #494, opened 2020-12-26
ci: add JDK 15

sullis:


rweisleder:
Same as #438?

codecholeric:
I'll close this in favor of #438, since I also think it's a duplicate.

PR #495, opened 2020-12-28
Fix typo

perlun:


perlun:
Ping @codecholeric

codecholeric:
Thanks a lot :smiley: I would have never thought someone really reads the imprint :joy: But thanks for not simply ignoring it, but fixing it!!

perlun:
> Thanks a lot smiley I would have never thought someone really reads the imprint joy But thanks for not simply ignoring it, but fixing it!!

Yeah, I have a tendency to be overly careful with meaningless details at times. But sometimes it's useful. :joy: :+1:

PR #501, opened 2021-01-05
Support to target only Test classes

pstanoev:
Sometimes it is useful to define rules about the tests them selves.
Like each test class should not be public, it should have name ending with "Test" etc.
This IncludeOption allows to target only test classes.

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
@@ -15,13 +15,16 @@
  */
 package com.tngtech.archunit.core.importer;
 
-import java.util.Set;
 import java.util.regex.Pattern;
 
-import com.google.common.collect.ImmutableSet;
+import com.google.common.base.Predicate;
 import com.tngtech.archunit.PublicAPI;
 
+import static com.google.common.base.Predicates.not;
+import static com.google.common.base.Predicates.or;
 import static com.tngtech.archunit.PublicAPI.Usage.INHERITANCE;
+import static com.tngtech.archunit.core.importer.ImportOption.Predefined.NO_TEST_LOCATION;
+import static com.tngtech.archunit.core.importer.ImportOption.Predefined.TEST_LOCATION;
 
 /**
  * Will be evaluated for every class location, to determine if the class should be imported.<br><br>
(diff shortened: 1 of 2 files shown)


archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
```diff
@@ -74,21 +74,23 @@ public boolean includes(Location location) {
...
+        }
+    }
+    
+    /**
+     * Import option to define rules only for Tests.
+     * See {@link DoNotIncludeTests} for limitations if tests identification.
+     */
+    final class OnlyTests extends LocationPatternImportOption {
+       
+        private static final Set<Pattern> INCLUDED_PATTERN = ImmutableSet.of(MAVEN_PATTERN, GRADLE_PATTERN, INTELLIJ_PATTERN);
```
hankem, line 90: This constant is unused, isn't it?

pstanoev: 8b715662339d41d8964862bf82f759ca927771bb

codecholeric:
Thanks a lot for your contribution :smiley: I agree that this makes sense!

Sorry that it took me so long to review, my pipeline is always too full :disappointed:

I had some review remarks which I actually put into a commit. 
One main thing is that I would not want the abstract class `LocationPatternImportOption` to be public (i.e. reachable from the outside). Also I'm not a big fan of sharing logic through inheritance. I.e. we never use this class in a polymorphic way, then that smells to me that we actually just wanted to share some logic. And for that I've had bad experiences with inheritance :wink: (on this small scale it's probably irrelevant, but composition over inheritance in these cases is more consistent to the rest of ArchUnit)
The other thing was, that booleans easily confuse me :joy: So passing a boolean `exclude` and then return `exclude ? false : true` or vice versa makes my brain cramp up too much. So I've tried to replace that by a `Predicate` based approach which I think is easier to read (i.e. `TEST_LOCATION = or(MAVEN_TEST, GRADLE_TEST, ...)` and simply inverting it via `not(..)` for the other `ImportOption`).

So feel free to check out my review commit and if it looks okay in the state that it is now, then please squash together all our commits into your first commit so I can merge it as one commit :smiley:

codecholeric:
Hey, I just squashed everything for you and rebased on master. I hope that's okay! It's just that I'm gonna release today and I was thinking that you'd be happy to have your contribution already in this release instead of having to wait for the next one :wink:

codecholeric:
Okay, it's merged and ready to ship :+1: Thanks a lot for your contribution! :smiley:

Issue #508, opened 2021-01-12
Use of System.lineSeparator() prevents development on heterogenous systems


gopf:
We have just started using ArchUnit in a small way, and have existing violations that we need to freeze.

Storing the currently frozen rules in the project along with the source seems to be the right thing to do. Getting that to work is a little bit tricky as I couldn't find a `freeze.store.default.path` setting for this use case that works both locally and in the automated build, but I have a tolerable workaround (copy with variable expansion from Ant, use variable setting in the local run target to override).

Unfortunately we have some developers that use Windows and some that use Macs, and the stored violations can't be shared between these two groups as the line endings vary. This is a blocker, and means that we can't really use ArchUnit.

hankem:
This might be a duplicate of #458?

gopf:
It's definitely related, but #458 relates more to the rule itself as key. Our rules are simple one-liners so the key isn't a problem, rather it's the violations themselves (which seem to be often multiline descriptions).

codecholeric:
Hmm, there is definitely some escaping / replacing of Windows linebreaks in place there, but maybe it's not enough :thinking: Do you have by any chance a minimal sample of a problematic scenario that works on Windows, but does not on Linux? (or vice versa, but since I'm using Linux myself it would be easier to reproduce, if I have a sample that works on Windows, but not on Linux :wink:)
I'm just a little puzzled, because there is an ArchUnit integration test, that has a checked in violation store file, and that works on Linux and Windows :thinking: 
You can compare [FrozenRulesTest](https://github.com/TNG/ArchUnit/blob/master/archunit-example/example-plain/src/test/java/com/tngtech/archunit/exampletest/FrozenRulesTest.java) / [the respective violation store](https://github.com/TNG/ArchUnit/tree/master/archunit-example/example-plain/src/test/resources/frozen) which works as a freeze input

* [on Ubuntu](https://github.com/TNG/ArchUnit/runs/1674208161?check_suite_focus=true#step:6:7650)
* [on MacOS](https://github.com/TNG/ArchUnit/runs/1674208214?check_suite_focus=true#step:6:8960)
* [on Windows](https://github.com/TNG/ArchUnit/runs/1674208292?check_suite_focus=true#step:6:7547)

[...]

Issue #458, opened 2020-10-23
Freezing rules issues on windows/non-windows environments


stefanroeck:
The default violation storage uses the textual description of violating rules as key. The description of LayeredArchitecture rules however use linebreaks in their text to separate the layer description which is nice on the console but leads to issues when running the same rules on windows and linux environments, as both rule violations are not considered as equal anymore (\r vs. \r\n).

A quick and dirty fix could be to replace any kind of linebreak inside of the FreezingRule, but one could also rethink if using the rule description is a good key or something more stable should be used instead.

stefanroeck:
Looking into the source code, `com.tngtech.archunit.library.freeze.ViolationLineMatcher` seems to tackle a similar problem, so this could also be a place to "fix" this. Or more robust solution could allow to provide an explicite rule key, similar to Sonar's `squid:` ids and use that to reference rules.

codecholeric:
Thanks for raising the issue! Yes, I can remember I had to tackle this in `ViolationLineMatcher` :wink: And it would certainly not be too hard to fix this for the rule text as well :thinking: 
As to your idea to use something like `squid`, I'm not sure I understand how to apply this. Wouldn't that require the rules to be from a fixed limited set? I.e. known in advance? Or in other words, if ArchUnit rules are completely free to design (both in what predicates/conditions they apply and the description they contain), how would there be deterministic IDs that are more robust than the rule description itself?
If there is a fixed set of provided rules it's easy, but ArchUnit essentially is just a rule toolkit, so until now I had no better idea how to identify a rule as by the mere description (including all the drawbacks).

stefanroeck:
> If there is a fixed set of provided rules it's easy, but ArchUnit essentially is just a rule toolkit, so until now I had no better idea how to identify a rule as by the mere description (including all the drawbacks).

I agree, this is tricky. The id somehow needs to be calculated for each violation by considering all related criteria. If there's a source-code location, this could be used, if there's a cycle of classes or so, maybe create a hash of all related class names or so? Technically, `ConditionEvent` could be the place where this id needs is attached to, would you agree?

codecholeric:
Sorry, I got sucked off this by some other issues :disappointed: I'll try to fix the concrete pain point together with #508 now, i.e. ignore line breaks in the key. 

[...]

PR #516, opened 2021-01-23
fix line ending problems in FreezingArchRule

codecholeric:
This addresses two problems:
1) if the rule description itself had different line separators than it was originally frozen with, `TextBasedViolationStore` would report `contains(rule) == false` and thus the result would be stored as new and the rule checking would succeed
2) if the single lines of the violation descriptions had a different line break than they were originally frozen with, these violations would all be removed as "solved" and the same violations with the different line break would be reported as new.

I think unfortunately the API is a little strange, i.e. that `event.getDescriptionLines()` would return lines that contain a line break (since what is the use to talk about a list of lines if the lines can contain line breaks and are thus multiple lines in themselves). But unfortunately this is not so easy to change, since the counting of violations is based on this at the moment. For example, that one long cycle report `cycle detected ... \n dependencies of ... \n ...` is counted as a single "description line". To change this we would have to adjust this mechanism and then optimally make it illegal to return strings with line breaks from `event.getDescriptionLines()`. But extending `ConditionEvent` and thus implementing this method is public API, so this might be a breaking change again. Altogether I decided to just fix this within `FreezingArchRule` for now and see, if there will be further problems with this in other areas.

To solve 1) I have adjusted `TextBasedViolationStore` to clean up line breaks from property names on save and load. I decided to add it to the file reading direction as well to a) also fix it for stores that have already saved Windows line breaks and b) support manual tinkering adding Windows line breaks in the future. Otherwise just storing them exclusively with `\n` as line separator would have probably been good enough. It would be better to solve 1) in a store independent way, so other store implementations would not have to deal with this problem, but unfortunately I could not find any way to do this, as long as the API is `contains(ArchRule)` and `store(ArchRule, ...)`. We could maybe always have converted these rules via `contains(rule.as(ensureUnixLineBreaks(rule.getDescription())))`, but that also felt a little shady, and might be surprising in a custom store implementation to get `rule.as(..)` instead of the rule directly.

Problem 2) I have addressed within `FreezingArchRule` to make it independent of the concrete store implementation. I decided to put adapters on both "ends" of the rule, i.e. on the `EvaluationResult` producing the violations "lines" on the one side, and the violation store where we read violations from on the other side. That way comparisons should only happen between Unix line separators and violations should always be stored with Unix line breaks within the store.

Resolves: #458
Resolves: #508

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/library/freeze/FreezingArchRule.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/freeze/FreezingArchRule.java
@@ -19,6 +19,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Properties;
 import java.util.Set;
 
 import com.tngtech.archunit.ArchConfiguration;
@@ -27,6 +28,7 @@ import com.tngtech.archunit.base.Predicate;
 import com.tngtech.archunit.core.domain.JavaClasses;
 import com.tngtech.archunit.lang.ArchRule;
 import com.tngtech.archunit.lang.EvaluationResult;
+import com.tngtech.archunit.lang.Priority;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


hankem (approved):
LGTM!

I've played a bit with violation stores on Windows and Linux and think that it works as intended.

Issue #515, opened 2021-01-23
Dependency do not support constant 


GeeJoe:
I tried to scan a class's dependency using `JavaClass.directDependenciesToSelf` but I found the constant in the class which is using by another class was not in the dependency list.

What should I do?

hankem:
I believe that the statement from https://github.com/TNG/ArchUnit/issues/298#issuecomment-583745288 is unfortunately still true:
> there is at the moment no workaround with ArchUnit to detect a dependency that is only observable through the constant pool.

Somebody needs to find some time to implement #309.

codecholeric:
I just played around a little and I think it will work to inspect `ldc` instructions with the current ASM version :smiley: I'll see if I can find some time the next days to implement this!

GeeJoe:
> I just played around a little and I think it will work to inspect `ldc` instructions with the current ASM version 😃 I'll see if I can find some time the next days to implement this!

Did you resolve this issue? I find the problem still exits in com.tngtech.archunit:archunit:0.17.0

codecholeric:
Are you sure? Can you show me an example? AFAIS something like this should now be detected, right?

```
class One {
    private Class<?> causesDependency = Two.class;
}

class Two {
}
```

What is your concrete case?

GeeJoe:
Like
```java 
class One {
    public static final String TAG = "Hello";
}

class Two {
     private String tag = One.TAG;
}
```
If I try to get `getDirectDependenciesFromSlef()` on class Two，the result won't contain class One

hankem:
There is some confusion: Since [ArchUnit 0.16.0](https://github.com/TNG/ArchUnit/releases/tag/v0.16.0), `JavaClass`/`JavaCodeUnit` know their references to other _class objects_ (like `Two.class` in the example above), but _primitive constants_ (like your `String TAG`) have a different signature in the bytecode (see also #298).

<details><summary>The bytecode for your example</summary>

```
class Two
  minor version: 0
  major version: 55
  flags: (0x0020) ACC_SUPER
  this_class: #5                          // Two
  super_class: #6                         // java/lang/Object
  interfaces: 0, fields: 1, methods: 1, attributes: 1
Constant pool:
   #1 = Methodref          #6.#15         // java/lang/Object."<init>":()V
   #2 = Class              #16            // One
   #3 = String             #17            // Hello
   #4 = Fieldref           #5.#18         // Two.tag:Ljava/lang/String;
   #5 = Class              #19            // Two
[...]

Issue #474, opened 2020-11-23
Cycles not detected


simasch:
`slices().matching("ch.jtaf.(*)").should().beFreeOfCycles().check(classes);` 

does not detect cycle between ch.jtaf.ui and ch.jtaf.security packages.

I assume that the problem is that the cycle comes from a subpackage of ch.jtaf.ui (see attchached image).

![Screenshot 2020-11-23 161306](https://user-images.githubusercontent.com/593352/99978921-d4074780-2da6-11eb-8717-4516fb02dc33.png)

codecholeric:
Ah, I think you need the pattern `ch.jtaf.(*)..` instead of `ch.jtaf.(*)`. The expression in total is still a package identifier (compare `com.tngtech.archunit.base.PackageMatcher`), so `ch.jtaf.(*)` only matches classes that are exactly in those packages, adding the `..` will add all sub-packages as well.
Does that help you?

simasch:
Hi Peter,

Unfortunately the test passes even with this expression:

`slices().matching("ch.jtaf.(*)..").should().beFreeOfCycles().check(classes);`

codecholeric:
Do you know how exactly the cycle is formed? I.e. which class from `ch.jtaf.ui` is depending on `ch,jtaf.security` and vice versa? ArchUnit still has some missing features that I'm trying to fix at the moment with respect to generic type parameters and method references. So maybe it's one of those causing the cycle? BTW. the most recent version 0.15.0 detects a couple of more occurrences (e.g. generic type parameters and array component types)

simasch:
These lines of code that produces the cycle:

SecurityContext
```
 public static boolean isAccessGranted(Class<?> securedClass) {
        final boolean publicView = LoginView.class.equals(securedClass) || DashboardView.class.equals(securedClass)
            || RouteNotFoundError.class.equals(securedClass);
```

ConfigureUIServiceInitListener
```
 private void beforeEnter(BeforeEnterEvent event) {
        final boolean accessGranted = SecurityContext.isAccessGranted(event.getNavigationTarget());
        if (!accessGranted) {
            if (SecurityContext.isUserLoggedIn()) {
                event.rerouteToError(AccessDeniedException.class);
            } else {
                event.rerouteTo(LoginView.class);
            }
        }
    }


```![image](https://user-images.githubusercontent.com/593352/102694536-fd8a8600-4221-11eb-8676-f65f27add9c6.png)

[...]

Issue #309, opened 2020-02-08
Import constant pool of classes


codecholeric:
If it is feasible with reasonable effort, we could take the constant pool of classes into account when importing classes. This would make it possible to detect dependencies which occur just through the constant pool of a class (compare #298)

Note that ASM does not support a direct access to the constant pool, thus I don't know out of the box how to achieve this in the best way. Maybe it would be possible to observe all the `ldc` instructions of all classes through ASM. Or we would need a different approach to read the constant pool of classes.

Issue #446, opened 2020-10-03
Violations not detected by ArchUnit


enriquemolinari:
Trying to understand why ArchUnit does not detect the cases below... 
(1) I guess this one is because the compiler remove that unused variable at all and nothing is translated to bycode. 
(3) Casting is just for the compiler nothing goes into bytecode.
(2) Why is not detected? can this be fixed?

Thanks!!!

```
package app.ui;
import app.business.AServiceImpl;

public class View {
 public void aMethod( ) {
  AServiceImpl a; // (1)
  Class a2 = AServiceImpl.class; // (2)
  System.out.println(a2.getName());
  if (AServiceImpl.class.equals(this.getClass())) { // (2)
   ....
  }
  Object o = new Object();
  AServiceImpl b = (AServiceImpl)o; // (3)
 }
}
	
 @Test
 public void arch() {
  JavaClasses jc = new ClassFileImporter().importPackages("app");
  noClasses().that()
   .resideInAnyPackage("app.ui..")
   .should()
   .accessClassesThat()
   .resideInAPackage("app.business")
   .check(jc);
 }
```

codecholeric:
ArchUnit does currently not handle all the information the bytecode would offer, e.g. the constant pool (compare #131).
Local variables without use like (1) are not detected at the moment. If you call a method or access a field on this concrete local variable, it will be detected.
Generic properties of `Class<?>` are currently also not detected, the only way to find this in the bytecode would again be the constant pool. Because all the accesses / dependencies are to generic `Class<?>` in the bytecode.
Casts like (3) could be detected (they are related closely to e.g. #417) but it hasn't been implemented yet :disappointed: 
Does that answer your questions?

enriquemolinari:
Many thanks for your reply. ArchUnit is an excellent tool, hope can be improved further.

codecholeric:
I'm gonna close this for now, since it's a little undirected. (2) has been implemented meanwhile, about (1) I'm not sure, because the compiler might also remove it if it's unused and and unused local variable will also never cause a real maintenance problem. For (3) I have created #710

PR #518, opened 2021-01-27
Add referenced class objects to `JavaClass`

codecholeric:
So far ArchUnit has not been able to detect the pure usage of class objects as dependencies. E.g. the following example would not have detected a dependency on `Evil`, since besides the reference to the class object no further dependency on `Evil` (like a field access or method call) has occurred.

```
class Example {
  final Map<Class<?>, Object> association = Map.of(Evil.class, anything);
}
```

With this PR `JavaClass` now knows its `referencedClassObjects`, including the respective line number. Furthermore class objects are now parts of the `dependencies{From/To}Self` of a `JavaClass`.

Resolves: #309 
Issue: #446 
Resolves: #474 
Resolves: #515


codecholeric:
Thanks a lot for the review @kaebiscs ! I agree that `ReferencedClassObject` is a way better name from a user's perspective :smile: I've tried to adjust all the places to this new terminology. Can you check again real quick that everything looks good to you?

kaebiscs (approved):
Review changes look good to me.

PR #522, opened 2021-02-06
gradle 6.8.2

sullis:
Command:  ./gradlew wrapper --gradle-version=6.8.2


codecholeric:
Thanks a lot, but we now have an automated process for upgrading Gradle :smiley: -> #531
Unfortunately the action was broken, so I had to fix it first to kick in again :disappointed: I hope it will be more robust now...

Issue #499, opened 2021-01-04
JavaAnnotation always populates properties with default values since 0.15.0


heri333:
Some libraries check whether properties are set *explicitly* on annotations. An example is the mapstruct [Mapper](https://github.com/mapstruct/mapstruct/blob/master/core/src/main/java/org/mapstruct/Mapper.java) annotation where the `unmappedTargetPolicy` does only override the `config` when set *explicitly*.

In ArchUnit **0.14.1** it could be checked whether the property was specified by the developer like `((JavaAnnotation)item).get("unmappedTargetPolicy").isPresent()`.

In ArchUnit **0.15.0** `isPresent()` is always `true`, since the property is initialized with the default value specified in the annotation.

Since the library does check whether the property was set, we need a way to also check this via ArchUnit to ensure the desired configuration. Since this worked in past versions, this looks like a regression to me. Or is there any new way to check this with ArchUnit?

hankem:
I can confirm that the behaviour changed from ArchUnit 0.14.1 to 0.15.0.
I personally agree with you that `isPresent()` should indicate whether the annotation's property is defined or not.

But irrespective of the desired behavior, I believe that the underlying issue has actually been there before: Even with ArchUnit 0.14.1, the following example, which explicitly imports `@Mapper`, fails to recognize that  `InvalidMapper` defines no `unmappedTargetPolicy`:
```java
import com.tngtech.archunit.junit.AnalyzeClasses;
import com.tngtech.archunit.junit.ArchTest;
import com.tngtech.archunit.junit.ArchUnitRunner;
import com.tngtech.archunit.lang.ArchRule;
import org.junit.runner.RunWith;
import org.mapstruct.Mapper;
import org.mapstruct.ReportingPolicy;

import static com.tngtech.archunit.base.DescribedPredicate.describe;
import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.classes;

@RunWith(ArchUnitRunner.class)
@AnalyzeClasses(packagesOf = {Issue499.class, Mapper.class})
public class Issue499 {
    @ArchTest
    static final ArchRule mapper_should_define_unmappedTargetPolicy =
        classes()
            .that().areAnnotatedWith(Mapper.class)
            .should().beAnnotatedWith(describe("an annotation with unmappedTargetPolicy property",
                annotation -> annotation.get("unmappedTargetPolicy").isPresent()
            ));

    @Mapper(unmappedTargetPolicy = ReportingPolicy.WARN)
[...]

PR #536, opened 2021-02-20
add methods to check for explicitly declared annotation properties

codecholeric:
We want to support the use case to check, if an annotation property has been explicitly declared as opposed to a default value. This is important for rules that want to check that certain values have always been specified instead of relying on the default. For example a mapstruct Mapper annotation where the `unmappedTargetPolicy` only overrides the config when set explicitly, but not if the default value is used. If the requirement is only that the property is specified, but the set of possible values is not known beforehand, then so far it was somewhat tedious to test this (compare to the default value, etc.).
The new API now allows to query if some property has been declared explicitly, and offers the two standard ways to `tryGet` the value (if it's assumed to possibly be absent) or "force" `get` the value (if it is known to be explicitly declared for sure).

Resolves: #499

### DIFF
--- a/archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAnnotationsTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAnnotationsTest.java
@@ -98,6 +98,30 @@ public class ClassFileImporterAnnotationsTest {
     }
 
     @Test
+    public void distinguishes_between_explicitly_set_values_and_default_values() {
+        @SuppressWarnings("DefaultAnnotationParam")
+        @SimpleAnnotationWithDefaultValues(
+                // this value is explicitly set to a different value than the default
+                defaultValue2 = "overwritten",
+                // this value is explicitly set, but actually the same as the default
+                defaultValue3 = "defaultValue3",
+                noDefault = "mustBeSet")
+        class AnnotatedClass {
+        }
+
+        JavaAnnotation<?> annotation = new ClassFileImporter().importClasses(AnnotatedClass.class)
+                .get(AnnotatedClass.class).getAnnotationOfType(SimpleAnnotationWithDefaultValues.class.getName());
+
+        assertThatAnnotation(annotation)
+                .hasExplicitlyDeclaredStringProperty("defaultValue2", "overwritten")
+                .hasExplicitlyDeclaredStringProperty("defaultValue3", "defaultValue3")
+                .hasExplicitlyDeclaredStringProperty("noDefault", "mustBeSet")
+
+                .hasNoExplicitlyDeclaredProperty("defaultValue1")
+                .hasStringProperty("defaultValue1", "defaultValue1");
+    }
+
+    @Test
     public void imports_class_with_one_annotation_correctly() {
         JavaClass clazz = new ClassFileImporter().importPackagesOf(ClassWithOneAnnotation.class)
                 .get(ClassWithOneAnnotation.class);
@@ -552,6 +576,17 @@ public class ClassFileImporterAnnotationsTest {
     @SuppressWarnings({"unchecked", "unused"})
     private static <T> T getAnnotationDefaultValue(JavaClass javaClass, String methodName, Class<T> valueType) {
         return (T) javaClass.getMethod(methodName).getDefaultValue().get();
+    }
+
+    @SuppressWarnings("unused")
+    private @interface SimpleAnnotationWithDefaultValues {
+        String defaultValue1() default "defaultValue1";
+
+        String defaultValue2() default "defaultValue2";
+
+        String defaultValue3() default "defaultValue3";
+
+        String noDefault();
     }
 
     @SuppressWarnings("unused")
--- a/archunit/src/test/java/com/tngtech/archunit/testutil/assertion/JavaAnnotationAssertion.java
+++ b/archunit/src/test/java/com/tngtech/archunit/testutil/assertion/JavaAnnotationAssertion.java
@@ -18,6 +18,7 @@ import com.tngtech.archunit.core.domain.JavaAnnotation;
 import com.tngtech.archunit.core.domain.JavaClass;
 import com.tngtech.archunit.core.domain.JavaEnumConstant;
 import org.assertj.core.api.AbstractObjectAssert;
+import org.assertj.core.api.ThrowableAssert;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.tngtech.archunit.testutil.Assertions.assertThat;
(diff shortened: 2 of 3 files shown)


archunit/src/test/java/com/tngtech/archunit/testutil/assertion/JavaAnnotationAssertion.java
```diff
@@ -58,6 +60,39 @@ public JavaAnnotationAssertion hasAnnotationProperty(String propertyName, Annota
...
+                .isEqualTo(expectedValue);
+        assertThat(actual.tryGetExplicitlyDeclaredProperty(propertyName))
+                .as("String @%s.%s()" + context(), actual.getRawType().getSimpleName(), propertyName)
+                .contains(actualValue);
+        return this;
+    }
+
+    public JavaAnnotationAssertion hasNoExplicitlyDeclaredProperty(final String propertyName) {
+        assertThat(actual.hasExplicitlyDeclaredProperty(propertyName))
+                .as("String @%s.%s()" + context() + " has explicitly declared property '%s'", propertyName)
```
hankem, line 79: When this assertion fails, it throws `java.util.MissingFormatArgumentException: Format specifier '%s'`.

I wouldn't worry about a performance impact of eagerly evaluating a few more String concatenations, and would simply always use 
```java
private String annotationPropertyDescription(String typeName, String propertyName) {
    return String.format("%s @%s.%s()", typeName, actual.getRawType().getSimpleName(), propertyName) + context();
}
```
You can find a suggestion on branch `add-API-to-retrieve-explicitly-set-annotation-values.review`.

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAnnotationsTest.java
```diff
@@ -97,6 +97,21 @@ public void imports_annotation_defaults() {
                 .as("default of clazzWithDefault()").matchExactly(Serializable.class, List.class);
     }
 
+    @Test
+    public void distinguishes_between_explicitly_set_values_and_default_values() {
+        @SimpleAnnotationWithDefaultValue(hasDefaultButOverwritten = "hasDefaultButOverwritten", hasNoDefault = "hasNoDefault")
```
hankem, line 102: I'd replace `"hasDefaultButOverwritten"` with `"overwritten"`, and `"hasNoDefault"` with `"value"`.

In addition, I'd love to test one more property with default, for which an explcit value is given which however coincides with the default.

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAnnotationsTest.java
```diff
@@ -554,6 +569,14 @@ public void parameters_of_meta_annotations_are_transitively_imported(HasAnnotati
         return (T) javaClass.getMethod(methodName).getDefaultValue().get();
     }
 
+    private @interface SimpleAnnotationWithDefaultValue {
+        String fallsBackToDefault() default "expected";
+
+        String hasDefaultButOverwritten() default "toOverwrite";
```
hankem, line 575: I think I'd find it less confusing if the property names of the annotation wouldn't contain semantics of the test that is defined 470 lines away. (Unfortunately, the annotation cannot be a local class – but it could be defined right next to the test... 😜)

So I'd just use:
```java
String hasDefault1() default "default value 1";
String hasDefault2() default "default value 2";
String hasDefault2() default "default value 3";
```
instead of `fallsBackToDefault` and `hasDefaultButOverwritten`.

archunit/src/test/java/com/tngtech/archunit/testutil/assertion/JavaAnnotationAssertion.java
```diff
@@ -58,6 +60,39 @@ public JavaAnnotationAssertion hasAnnotationProperty(String propertyName, Annota
         return this;
     }
 
+    public JavaAnnotationAssertion hasExplicitlyDeclaredStringProperty(String propertyName, String expectedValue) {
+        assertThat(actual.hasExplicitlyDeclaredProperty(propertyName))
+                .as("String @%s.%s()" + context() + " has no explicitly declared value", actual.getRawType().getSimpleName(), propertyName)
```
hankem, line 65: This boolean is about whether the property _does_ have an explicitly declared value, isn't it?

hankem: I've updated my commit on [`add-API-to-retrieve-explicitly-set-annotation-values.review`](https://github.com/TNG/ArchUnit/commits/add-API-to-retrieve-explicitly-set-annotation-values.review). What I forgot to say before: Feel free to squash it, if you like it. 😉

archunit/src/test/java/com/tngtech/archunit/testutil/assertion/JavaAnnotationAssertion.java
codecholeric: Good point, thanks a lot!

codecholeric: (it's a weird mix of concat + format anyway :joy:)

codecholeric: (it's a weird mix of concat + format anyway :joy:)

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAnnotationsTest.java
codecholeric: Fair point :smiley:

codecholeric: I kind of like `hasNoDefault` though, because it makes it clearer what the point is here? :thinking:

codecholeric: I mean, for "overwritten" it already implies there is a default :wink: But "value" does not really communicate that there is no default to me :thinking:

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAnnotationsTest.java
codecholeric: Yeah, I know, I wish you could define it as local class :disappointed: I still like it to communicate what the test means :thinking: Maybe I can find a way that this part is only in the test.

archunit/src/test/java/com/tngtech/archunit/testutil/assertion/JavaAnnotationAssertion.java
codecholeric: Yeah, I should not do some last minute description copy and paste at 2am :disappointed: Thanks a lot for cleaning it up!

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAnnotationsTest.java
codecholeric: As mentioned below, I have tried to put the specifics closer to the test. Do you think it's okay now?

Issue #469, opened 2020-11-12
Inheritance and Interface checks String vs Class parameter


nixfuerdiecharts:
Frist of all I am an ArchUnit newbie. So I am not shure if this is a real issue or I am misunderstanding the api.

I have implemented a simple Code structure, where a Car class inherits from a Vehicle class. I am trying to test that inheritance with ArchUnit.
When I implement the test with a String parameter, everything works fine, the test goes trough.
```java
@ArchTest
    private final ArchRule vehicle_extentions_should_be_car = classes().that().implement("Vehicle")
            .should().haveSimpleName("Car");
```
But when I pass through the Vehicle as a class, it throws Exceptions
```java
@ArchTest
    private final ArchRule vehicle_extentions_should_be_car = classes().that().implement(Vehicle.class)
            .should().haveSimpleName("Car");
```
```java
java.lang.reflect.InvocationTargetException
com.tngtech.archunit.junit.ArchTestInitializationException: java.lang.reflect.InvocationTargetException
	at com.tngtech.archunit.junit.ArchTestInitializationException$1.apply(ArchTestInitializationException.java:27)
	at com.tngtech.archunit.junit.ArchTestInitializationException$1.apply(ArchTestInitializationException.java:24)
	at com.tngtech.archunit.junit.ReflectionUtils.getValueOrThrowException(ReflectionUtils.java:102)
	at com.tngtech.archunit.junit.ArchTestExecution.getValue(ArchTestExecution.java:53)
	at com.tngtech.archunit.junit.ArchRuleExecution.evaluateOn(ArchRuleExecution.java:39)
	at com.tngtech.archunit.junit.ArchUnitRunner.runChild(ArchUnitRunner.java:154)
	at com.tngtech.archunit.junit.ArchUnitRunner.runChild(ArchUnitRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at com.tngtech.archunit.junit.ArchUnitRunner$1.evaluate(ArchUnitRunner.java:88)
[...]

PR #537, opened 2021-02-20
improve Javadoc for 'implement' vs 'assignable'

codecholeric:
In the past there has been repeatedly confusion about the methods `..that().implement(..)` / `.should().implement(..)`. Users do not always realize that 'implements' really only refers to the Java syntax, i.e. `A implements B` is only valid, if `A` is a class and `B` is an interface. Otherwise it can only be `extends`, which is not meant here. A while ago I had added some rejection for `implement(clazz)` to throw an exception, if `clazz` is not an interface, because this predicate can never match anything. Unfortunately this is still not clear enough. I think in many cases people actually want `assignableTo(..)`, which is very close to the naming of the Java Reflection API, but still not widely spread as it seems. To make it easier to find `assignableTo(..)` for users that simply want to test for some inheritance relation I have extended the Javadoc to explain the `implements` vs `assignableTo` case in more detail and also added a hint "do you maybe want to use assignableTo(..)" to the exception of `implement(type)`. I hope this will provide more guidance which method to use in which case and makes it easier to find what is needed.

Resolves: #469

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
@@ -1637,50 +1637,124 @@ public class JavaClass implements JavaType, HasName.AndFullName, HasAnnotations<
             return new SimpleNameEndingWithPredicate(suffix);
         }
 
+        /**
+         * @param type the type to check for assignability
+         * @return a {@link DescribedPredicate} that returns {@code true}, if the respective {@link JavaClass}
+         *         is assignable to the supplied {@code type}. I.e. the type represented by the tested {@link JavaClass}
+         *         could be casted to the supplied {@code type}.<br>
+         *         This is the opposite of {@link #assignableFrom(Class)}:
+         *         some class {@code A} is assignable to a class {@code B} if and only if {@code B} is assignable from {@code A}.
+         *
+         * @see #assignableTo(String)
+         * @see #assignableTo(DescribedPredicate)
+         * @see #assignableFrom(Class)
+         */
         @PublicAPI(usage = ACCESS)
         public static DescribedPredicate<JavaClass> assignableTo(final Class<?> type) {
             return assignableTo(type.getName());
         }
 
+        /**
+         * @param type the type to check for assignability
+         * @return a {@link DescribedPredicate} that returns {@code true}, if the respective {@link JavaClass}
+         *         is assignable from the supplied {@code type}. I.e. the supplied {@code type}
+         *         could be casted to the type represented by the tested {@link JavaClass}.<br>
+         *         This is the opposite of {@link #assignableTo(Class)}:
+         *         some class {@code B} is assignable from a class {@code A} if and only if {@code A} is assignable to {@code B}.
+         *
+         * @see #assignableFrom(String)
+         * @see #assignableFrom(DescribedPredicate)
+         * @see #assignableTo(Class)
+         */
         @PublicAPI(usage = ACCESS)
         public static DescribedPredicate<JavaClass> assignableFrom(final Class<?> type) {
             return assignableFrom(type.getName());
         }
 
+        /**
+         * Same as {@link #assignableTo(Class)} but takes a fully qualified class name as an argument instead of a class object.
+         */
         @PublicAPI(usage = ACCESS)
         public static DescribedPredicate<JavaClass> assignableTo(final String typeName) {
             return assignableTo(GET_NAME.is(equalTo(typeName)).as(typeName));
         }
 
+        /**
+         * Same as {@link #assignableFrom(Class)} but takes a fully qualified class name as an argument instead of a class object.
+         */
         @PublicAPI(usage = ACCESS)
         public static DescribedPredicate<JavaClass> assignableFrom(final String typeName) {
             return assignableFrom(GET_NAME.is(equalTo(typeName)).as(typeName));
         }
 
+        /**
+         * Same as {@link #assignableTo(Class)}, but returns {@code true} whenever the tested {@link JavaClass}
+         * is assignable to a class that matches the supplied predicate.<br>
+         * This is the opposite of {@link #assignableFrom(DescribedPredicate)}:
+         * some class {@code A} is assignable to a class {@code B} if and only if {@code B} is assignable from {@code A}.
+         *
+         * @see #assignableTo(Class)
+         * @see #assignableTo(String)
+         * @see #assignableFrom(DescribedPredicate)
+         */
         @PublicAPI(usage = ACCESS)
         public static DescribedPredicate<JavaClass> assignableTo(final DescribedPredicate<? super JavaClass> predicate) {
             return new AssignableToPredicate(predicate);
         }
 
+        /**
+         * Same as {@link #assignableFrom(Class)}, but returns {@code true} whenever the tested {@link JavaClass}
+         * is assignable from a class that matches the supplied predicate.<br>
+         * This is the opposite of {@link #assignableTo(DescribedPredicate)}:
+         * some class {@code B} is assignable from a class {@code A} if and only if {@code A} is assignable to {@code A}.
+         *
+         * @see #assignableFrom(Class)
+         * @see #assignableFrom(String)
+         * @see #assignableTo(DescribedPredicate)
+         */
         @PublicAPI(usage = ACCESS)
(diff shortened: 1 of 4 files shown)


archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
```diff
@@ -1637,50 +1637,119 @@ public boolean apply(JavaClass input) {
...
         public static DescribedPredicate<JavaClass> assignableFrom(final DescribedPredicate<? super JavaClass> predicate) {
             return new AssignableFromPredicate(predicate);
         }
 
+        /**
+         * @param type the interface type to check for
+         * @return a {@link DescribedPredicate} that returns {@code true}, if the respective {@link JavaClass} implements the supplied
+         *         interface type. I.e. the supplied class must be an interface and the {@link JavaClass} must be a class for this
+         *         to ever be true (it resembles delarations like `class A implements B`, which only works for a class A and
+         *         an interface B)
```
hankem, line 1721:
```suggestion
         * @return a {@link DescribedPredicate} that returns {@code true} if the tested {@link JavaClass} implements the supplied
         *         interface {@code type}. I.e. the supplied {@code type} must be an interface and the tested {@link JavaClass} must be a class
         *         (it resembles delarations like {@code class A implements B}, which only works for a class {@code A} and an interface {@code B}).
```

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
```diff
@@ -1637,50 +1637,119 @@ public boolean apply(JavaClass input) {
...
+        /**
+         * Same as {@link #implement(Class)} but takes a fully qualified class name as an argument instead of a class object.
+         */
         @PublicAPI(usage = ACCESS)
         public static DescribedPredicate<JavaClass> implement(final String typeName) {
             return implement(GET_NAME.is(equalTo(typeName)).as(typeName));
         }
 
+        /**
+         * Same as {@link #implement(Class)} but returns {@code true} whenever the respective {@link JavaClass} implements
```
hankem, line 1746:
```suggestion
         * Same as {@link #implement(Class)} but returns {@code true} whenever the tested {@link JavaClass} implements
```

archunit/src/main/java/com/tngtech/archunit/lang/conditions/ArchConditions.java
```diff
@@ -775,91 +775,145 @@ public static AllDependenciesCondition onlyHaveDependenciesWhere(DescribedPredic
         return not(ArchConditions.<HAS_ANNOTATIONS>beMetaAnnotatedWith(predicate));
     }
 
+    /**
+     * @return A condition matching classes analogously to {@link JavaClass.Predicates#implement(Class)}
```
hankem, line 779: I wouldn't start with a capital `A` here (and in the similar cases below).
```suggestion
     * @return a condition matching classes analogously to {@link JavaClass.Predicates#implement(Class)}
```

archunit/src/main/java/com/tngtech/archunit/lang/syntax/elements/ClassesShould.java
```diff
@@ -411,16 +417,23 @@
      *
      * @param type An interface imported classes should NOT implement
      * @return A syntax element that can either be used as working rule, or to continue specifying a more complex rule
+     *
+     * @see ArchConditions#implement(Class)
```
hankem, line 421: Shouldn't we consequently link for `ArchConditions#notImplement` from the `notImplement` methods (here and below)?
```suggestion
     * @see ArchConditions#notImplement(Class)
```
(Same for `notBeAssignable`{`To`,`From`}.)

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
```diff
@@ -1637,50 +1637,119 @@ public boolean apply(JavaClass input) {
...
             return new AssignableFromPredicate(predicate);
         }
 
+        /**
+         * @param type the interface type to check for
+         * @return a {@link DescribedPredicate} that returns {@code true}, if the respective {@link JavaClass} implements the supplied
+         *         interface type. I.e. the supplied class must be an interface and the {@link JavaClass} must be a class for this
+         *         to ever be true (it resembles delarations like `class A implements B`, which only works for a class A and
+         *         an interface B)
+         *
```
hankem, line 1722: Should we even document
```suggestion
         * @throws InvalidSyntaxUsageException if {@code type} is not an interface
```
?

archunit/src/main/java/com/tngtech/archunit/lang/syntax/elements/ClassesThat.java
```diff
@@ -391,6 +394,9 @@
     /**
      * Matches classes that implement a certain interface with the given type name. This is equivalent to
      * {@link #implement(Class)}, but does not depend on having a certain type on the classpath.
+     * Note that this only matches non-interface {@link JavaClass classes} that implement the supplied interface {@code typeName}
+     * (compare {@link JavaClass.Predicates#implement(Class)}.
```
hankem, line 398:
```suggestion
     * (compare {@link JavaClass.Predicates#implement(String)}.
```

archunit/src/main/java/com/tngtech/archunit/lang/syntax/elements/ClassesThat.java
```diff
@@ -413,6 +419,9 @@
      * Matches classes that implement a certain interface matching the given predicate. For example, a call with
      * {@link Predicates#name(String)} would be equivalent to
      * {@link #implement(String)}, but the approach is a lot more generic.
+     * Note that this only matches non-interface {@link JavaClass classes} that implement an interface matching the {@code predicate}
+     * (compare {@link JavaClass.Predicates#implement(Class)}.
```
hankem, line 423:
```suggestion
     * (compare {@link JavaClass.Predicates#implement(DescribedPredicate)}.
```

hankem (approved):
Thanks for these clarifyign JavaDocs, I'm sure they will be very helpful for users! 💙 

I've left a few comments, which are however just ideas that came to my mind; they're not necessarily better than your initial versions. Feel free to pick the alternative you like better at the end.

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
```diff
@@ -1637,50 +1637,119 @@ public boolean apply(JavaClass input) {
             return new SimpleNameEndingWithPredicate(suffix);
         }
 
+        /**
+         * @param type the type to check for assignability
+         * @return a {@link DescribedPredicate} that returns {@code true}, if the respective {@link JavaClass}
+         *         is assignable to the supplied {@code type}. I.e. the type represented by the {@link JavaClass}
+         *         could be casted to the type represented by the supplied {@link Class}.<br>
+         *         This is the opposite of {@link #assignableFrom(Class)}
```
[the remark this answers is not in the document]
codecholeric: Hmm, yeah, that's why I picked "opposite" vs "negation" :joy: But yeah, might be misleading :thinking:

archunit/src/main/java/com/tngtech/archunit/lang/syntax/elements/ClassesThat.java
codecholeric: Hmm, this was on purpose, because only `implement(Class)` has the explanation :thinking: If you jump to the `String` version you'll just find another link to the `Class` version :man_shrugging:

archunit/src/main/java/com/tngtech/archunit/lang/syntax/elements/ClassesShould.java
codecholeric: I thought there was only the positive case in `ArchConditions`, sure :+1:

archunit/src/main/java/com/tngtech/archunit/lang/syntax/elements/ClassesThat.java
codecholeric: As described above, I left these, because it's really meant as a "compare", i.e. see the explanation of the semantics there, not see the equivalent method there. I hope it's still clear, I just feel that jumping to a method that simply says "this is the same as $otherMethod, see there" is also somewhat frustrating :wink:

Issue #525, opened 2021-02-11
ArchRules doesn't consider static ArchRule fields


OLibutzki:
As far as I can see `com.tngtech.archunit.junit.ArchRules` scans the given `definitionLocation` for methods and fields which are annotated with `com.tngtech.archunit.junit.ArchTest`.

I would like to include a class which defines fields of type `com.tngtech.archunit.lang.ArchRule` like `com.tngtech.archunit.library.GeneralCodingRules`. 

```
@AnalyzeClasses
public class CodingRuleTest {

	@ArchTest
	static ArchRules generalRules = ArchRules.in(GeneralCodingRules.class);
	
}
```

hankem:
Could you elaborate what is not considered in your opinion?

Including rules via `ArchRules.in` should actually work, cf. [`RuleSetsTest`](https://github.com/TNG/ArchUnit-Examples/blob/master/example-junit5/src/test/java/com/tngtech/archunit/exampletest/junit5/RuleSetsTest.java) and [`RuleLibraryTest`](https://github.com/TNG/ArchUnit-Examples/blob/master/example-junit5/src/test/java/com/tngtech/archunit/exampletest/junit5/RuleLibraryTest.java) of the [ArchUnit-Examples](https://github.com/TNG/ArchUnit-Examples).

OLibutzki:
The fields/methods of the class referenced by `ArchRules.in` have to be annotated with `@ArchTest`.

In my opinion it should be sufficient to scan the fields/methods pretty much the arch-unit-maven-plugin does.

The `@AnalyzeClasses` annotation is provided by the test class which includes the other test classes... Therefore there is no need to use archunit-junit API in the classes you include by using `ArchRules.in`.

hankem:
Ah, I see: [`GeneralCodingRules`](https://github.com/TNG/ArchUnit/blob/v0.16.0/archunit/src/main/java/com/tngtech/archunit/library/GeneralCodingRules.java) does not have `@ArchTest` annotations, but you'd still want to include its `ArchRule`s.

Thanks for the clarification! 💙

codecholeric:
Hmm, I'm not completely sure how this would go :thinking: Because `ArchRules.in(..)` is also a little specific to `@ArchTest`. For example, it will also evaluate methods like

```
@ArchTest
static void my_rule(JavaClasses classes) {
    // evaluate something on classes
}
```

[...]

PR #538, opened 2021-02-21
introduce `ArchTests` instead of `ArchRules`

codecholeric:
The term `ArchRules` is misleading, because it suggests that the class collects instances of `ArchRule`, while in fact it looks for members annotated with `@ArchTest`. This becomes particular clear, if you look at method declarations like `@ArchTest static void rule_method(JavaClasses classes) {..}`. I have thus introduced `ArchTests.in(someClass)` as a new version that makes this clearer (including some Javadoc this time) and kept the old `ArchRules` as a deprecated version for now to be backwards compatible.
I have duplicated quite some code (like whole test classes) and not unified much of the additional complexity by having `ArchTests` as well as `ArchRules`, since this is only temporary and we can kick it out in the near future.

Resolves: #525

### DIFF
--- a/archunit-junit/src/api/java/com/tngtech/archunit/junit/ArchTests.java
+++ b/archunit-junit/src/api/java/com/tngtech/archunit/junit/ArchTests.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2018 TNG Technology Consulting GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.tngtech.archunit.junit;
+
+import com.tngtech.archunit.PublicAPI;
+
+import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
+
+/**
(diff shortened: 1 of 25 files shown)


archunit-junit/src/api/java/com/tngtech/archunit/junit/ArchTests.java
```diff
@@ -0,0 +1,78 @@
...
+ * limitations under the License.
+ */
+package com.tngtech.archunit.junit;
+
+import com.tngtech.archunit.PublicAPI;
+
+import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
+
+/**
+ * Includes all {@code @ArchTest} members of another class into this ArchUnit test. For example
```
hankem, line 23: Maybe 
```suggestion
 * Includes all {@code @ArchTest} annotated members of another class into this ArchUnit test. For example
```
?

docs/userguide/009_JUnit_Support.adoc
```diff
@@ -186,14 +186,14 @@ public class PersistenceRules {
...
-    public static final ArchRules persistenceRules = ArchRules.in(PersistenceRules.class);
+    static final ArchTests persistenceRules = ArchTests.in(PersistenceRules.class);
 
 }
 ----
 
-The runner will evaluate all rules within `ServiceRules` and `PersistenceRules` against
-the classes declared at `ArchitectureTest`. This also allows an easy reuse of a rule library
-in different projects or modules.
+The runner will include all `@ArchTest` members within `ServiceRules` and `PersistenceRules` and evaluate
```
hankem, line 197:
```suggestion
The runner will include all `@ArchTest` annotated members within `ServiceRules` and `PersistenceRules` and evaluate
```

hankem (approved):
👍

archunit-junit/src/api/java/com/tngtech/archunit/junit/ArchTests.java
codecholeric: Accepted and merged, thanks :smiley:

Issue #209, opened 2019-08-01
Record Method Calls to the same Class


waschmittel:
I want to create an ArchRule to make sure that methods annotated with Spring's @Async annotation are not called from the same class.

(These methods should not be called from the same class because Spring needs to create a proxy to make @Async work.)

So I made this rule:

```java
    @ArchTest
    public static final ArchRule asyncMethodsShouldNotBeCalledFromTheSameInstance = noClasses()
            .should()
            .callMethodWhere(DescribedPredicate.describe("no async from same class", input ->
                    {
                        return input.getTarget().isAnnotatedWith(Async.class) &&
                                input.getOriginOwner().equals(input.getTargetOwner());
                    }
            ))
            .as("@Async Methods should never be called from the same class because no proxy can be used then.");
```

Unfortunately, this does not work because apparently method calls to the same class do not seem to be collected. Is there any way I can configure this?

codecholeric:
You are right, once upon a time people complained, that those self accesses would be utterly useless, because you're *always* interested in accesses to *other* classes :wink: So I changed the behavior to please the masses.
I'm really sorry, but I think at the moment you need to add a little boilerplate, e.g. like this:

```
noClasses().should(callAsyncMethodsFromTheSameClass())
    .because("it circumvents Spring's proxy mechanism");

private ArchCondition<JavaClass> callAsyncMethodsFromTheSameClass() {
    return new ArchCondition<JavaClass>("call @Async Methods from the same class") {
        @Override
        public void check(JavaClass javaClass, ConditionEvents events) {
            if (!javaClass.isAnnotatedWith(Async.class)) {
                return;
            }

            for (JavaMethodCall call : javaClass.getMethodCallsFromSelf()) {
                boolean originIsTarget = call.getOriginOwner().equals(call.getTargetOwner());
                events.add(new SimpleConditionEvent(call, originIsTarget, call.getDescription()));
            }
        }
    };
}
```

codecholeric:
This reminds me that `ArchCondition` also needs some lambda'ification factory method :wink:

honnel:
Thanks for your attention and quick reply!

We had to slightly customize the workaround for `@Async`annotated methods:

```java
   @ArchTest
    public static final ArchRule asyncMethodsShouldNotBeCalledFromTheSameInstance = noClasses()
            .should(callAsyncMethodsFromTheSameClass())
            .as("@Async Methods should never be called from the same class because no proxy can be used then.");

[...]

PR #539, opened 2021-02-21
Add library rule to check proxy bypasses

codecholeric:
In frameworks like Spring there often are annotations that cause the framework to proxy specific classes and add some AOP behavior. For example `@Transactional` in Spring denoting a method to create/participate in a transaction on enter and commit it on leaving the method. This only works though, if all callers really call this method on the proxy and not the plain class itself (as long as it isn't weaved into the bytecode). It is a common error to call one of these decorated methods from within another method in the same class, thus bypassing the proxy.
I've added a new library class `ProxyRules` to encapsulate some rules that help with ensuring correct calls to proxies. At the moment it offers rules / conditions to check for self method calls to methods annotated with a certain annotation or more generally matching some predicate.

Issue: #209

### DIFF
--- a/archunit/src/test/java/com/tngtech/archunit/library/ProxyRulesTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/library/ProxyRulesTest.java
@@ -1,0 +1,82 @@
+package com.tngtech.archunit.library;
+
+import com.tngtech.archunit.core.importer.ClassFileImporter;
+import com.tngtech.archunit.lang.ArchRule;
+import com.tngtech.java.junit.dataprovider.DataProvider;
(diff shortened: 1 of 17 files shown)


archunit/src/test/java/com/tngtech/archunit/library/ProxyRulesTest.java
```diff
@@ -0,0 +1,77 @@
...
+            }
+        }
+        class SomeClass {
+            void evil() {
+                self();
+            }
+
+            void okay() {
+                new OtherClass().proxied();
+            }
```
hankem, line 57:
```suggestion
                nonProxied();
            }
            
            void nonProxied() {
            }
```

Issue #527, opened 2021-02-12
Repetitive annotation conjunction


knittl:
ArchUnit is a wonderful library, thanks! I found one little nuisance though with regard to annotation-based rules. Consider the following Spring controller:

```
@RestController
public class MyController {
  @GetMapping("{id}/name")
  public String getName(@PathVariable final String id) {
    return "Name of " + id;
  }

  @RequestMapping(value = "{name}/id", method = GET)
  public String getId(@PathVariable final String name) {
    return "Id of " + name;
  }
}
```

Let's say I want to have a rule that verifies that all my controller methods return `String`s (not a real requirement, just an example). Simple:

```
@ArchTest
static final ArchRule must_return_strings = methods()
        .that().areAnnotatedWith(RequestMapping.class)
        .or().areMetaAnnotatedWith(RequestMapping.class)
        .should().notHaveRawReturnType(String.class);
```

This works and does what I expect it to do. But it feels repetitive having to specify the annotation twice: once as a direct annotation and a second time as a meta annotation. I find myself very often in this situation, especially when writing rules for a Spring application, since Spring mostly does not distinguish between annotations and meta annotations.

Changing this to:

```
@ArchTest
static final ArchRule must_return_strings = methods()
        .that().areMetaAnnotatedWith(RequestMapping.class)
        .should().notHaveRawReturnType(String.class);
```

no longer fails for the `getId` method if its signature is in violation of the rule. Am I missing something from the docs? Ideally, I'd like to express the rule `.that().areAnnotatedOrMetaAnnotatedWith(MyAnnotation.class)` without having to repeat myself.

Are there common cases where we need to filter annotation targets by meta annotation _only_? From my point of view, when filtering for meta annotations, you want to include direct annotations most of the time.

Unfortunately, I cannot come up with a new name which would combine the two conjunctions. Thus my proposal would be a breaking change:

* `areMetaAnnotatedWith()` should (by default) include meta annotated elements _and_ directly annotated elements (**this potentially changes behavior of existing rules**)
* Introduce a new conjunction `areOnlyMetaAnnotatedWith()` which excludes directly annotated elements and only matches elements whose annotations itself are meta-annotated with the specified class.

I'd be happy to work on a PR, but before starting, I wanted to double-check if no such possibility already exists. If such a method already exists, please point me to the docs and close this issue :)

codecholeric:
[...]

PR #540, opened 2021-02-21
Make meta annotated match direct annotations as well

codecholeric:
So far the most common use case for checking rules for meta-annotations seems to be annotation composition, where users want to test that certain classes/members are either directly annotated with some annotation, or with some annotation that is meta-annotated with the respective annotation. Thus it makes sense to cover this use case directly with the `metaAnnotated(..)` syntax. For example

```
@interface A{}

@A
@interface B{}

class Foo {
    @A Object directlyAnnotated;
    @B Object indirectlyAnnotated;
}
```

Previously only `indirectlyAnnotated` would have counted as meta-annotated with `@A`. Now we count both those methods as meta-annotated with `@A`. The old behavior of really being annotated only through the hierarchy, but not directly can still be expressed as `metaAnnotatedWith(A.class).and(not(annotatedWith(A.class)))`, which seems to be good enough still to cover the old behavior. Besides that the only use case I could imagine is to enforce the use of a custom composite annotation, which could also be achieved by simply forbidding the original annotation altogether.

Resolves: #527

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/properties/CanBeAnnotated.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/properties/CanBeAnnotated.java
@@ -175,8 +175,12 @@ public interface CanBeAnnotated {
                 return false;
             }
 
+            if (predicate.apply(annotation)) {
+                return true;
+            }
+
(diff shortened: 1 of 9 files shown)


hankem:
LGTM!

hankem (approved):
LGTM!

PR #544, opened 2021-02-28
Dependency upgrades

hankem:
This PR upgades some dependencies that are not automatically detected by @dependabot due to the structure of the build script.


codecholeric:
Thanks a lot :smiley: I took the liberty to squash the Guava upgrade, since we discussed it and agreed that all tests green is good enough :wink: In the end it's relocated, so there shouldn't be any problem. Looking at all the comments I really think we should drop Java 7 support in the near future :thinking:

Issue #545, opened 2021-03-03
Feature Request: Classes that contain annotated methods


nils-christian:
Hello,

First of all I would like to thank you for providing us with ArchUnit. It is a very helpful tool during our development.

Now about my feature request. We have some use cases where we want to find classes that contain methods annotated with a specific annotation. Currently it seems like this is not possible. Assume we want to find classes that contain at least one method annotated with _EventHandler_ and want to make sure that those classes are annotated with _ProcessingGroup_. This would lead to following code:

    public static final ArchRule EVENT_HANDLERS_SHOULD_BE_EXPLICITLY_ASSIGNED_TO_A_PROCESSING_GROUP = classes( )
		.that( containEventHandlers( ) )
		.should( ).beAnnotatedWith( ProcessingGroup.class )
		.because( "Event handlers should be explicitly assigned to a processing group." );

    private static DescribedPredicate<JavaClass> containEventHandlers( ) {
	  return new DescribedPredicate<>( "contain event handlers" ) {

		@Override
		public boolean apply( final JavaClass input ) {
			return input.getAllMethods( ).stream( ).anyMatch( m -> m.isAnnotatedWith( EventHandler.class ) );
		}
	};

    }

I would like to use something like the following

    public static final ArchRule EVENT_HANDLERS_SHOULD_BE_EXPLICITLY_ASSIGNED_TO_A_PROCESSING_GROUP = classes( )
		.that( ).containMethodsAnnotatedWith( EventHandler.class )
		.should( ).beAnnotatedWith( ProcessingGroup.class )
		.because( "Event handlers should be explicitly assigned to a processing group." );

hankem:
Thanks for your feedback! Do I understand correctly that your statement
> it seems like this is not possible

refers to the fluent API only? (Because your predicate-based solution seems to be functional to me. 😉)

nils-christian:
Hi,

You are correct of course. It works like a charm, but I would prefer if we could express the rule with your fluent API.

codecholeric:
Thanks for the input :smiley: I'm always trying to balance what should go in the core API and what should just patched in by users via `Predicates` the way you did it.
I feel the `containMethodsAnnotatedWith(..)` extension would still be fair, since it's generic enough to bring benefit to various users :+1: 
[...]

PR #547, opened 2021-03-07
Issue: #545 Feature Request: Classes that contain annotated methods

nils-christian:
Signed-off-by: Nils Christian Ehmke <nils@rhocas.de>

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
@@ -1827,6 +1827,18 @@ public class JavaClass implements JavaType, HasName.AndFullName, HasAnnotations<
         public static DescribedPredicate<JavaClass> belongToAnyOf(Class<?>... classes) {
             return new BelongToAnyOfPredicate(classes);
         }
+        
+        /**
+         * A predicate to determine if a {@link JavaClass} contains one or more methods matching the supplied predicate.
+         *
+         * @param predicate The predicate to check against the {@link Class classes'} methods.
+         * @return A {@link DescribedPredicate} returning true, if and only if the tested {@link JavaClass} coontains at least
+         * one method matching the given predicate.
+         */
+        @PublicAPI(usage = ACCESS)
+        public static DescribedPredicate<JavaClass> containMethodsThatAre(DescribedPredicate<? super JavaMethod> predicate) {
+            return new ContainMethodsThatArePredicate(predicate);
+        }
 
         private static class BelongToAnyOfPredicate extends DescribedPredicate<JavaClass> {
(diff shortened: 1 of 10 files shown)


codecholeric:
Hey, thank you so much for the PR, I really appreciate the support :smiley: 
I only think I didn't completely look at this from all angles, when I said I consider it fair to add this method. Somehow I looked at the predicate method you added to `JavaClass`, namely

```
containMethodsThatAre(DescribedPredicate<? super JavaMethod> predicate)
```

and I immediately thought "that method would be super flexible and cover it all". Since I had a super bad conscience that I told you "go for containMethodsThatAreAnnotatedWith(..) in the API" and now do a 180 and say "well, actually I like that more generic method better" I decided to give up my resolution about my current priorities and put in some effort here as well to make up for it.

So what I did is I pushed your new method `containMethodsThatAre(..)` through to the API and replaced the more specific `containMethodsthatAreAnnotatedWith(..)` method by that. I just had the feeling, with this method you could easily and concisely write

```
import static com.tngtech.archunit.core.domain.properties.CanBeAnnotated.Predicates.annotatedWith;
import static com.tngtech.archunit.lang.conditions.ArchPredicates.are;
---
classes().that().containAnyMethodsThat(are(annotatedWith(Deprecated.class)))
```

But at the same time nobody is going to ask "I want to filter classes that have a member of type X, why can I only check for annotations?" in two weeks. And also this way you can easily plug in `metaAnnotatedWith(..)` for example.

What do you think about this way to go? If you like it, you could squash my commit on top of yours. I thought about keeping the `containMembersAnnotatedWith(..)` method as well, but then decided against it, because it is pretty specific. And there are a lot of other methods of the same type that would be missing.
On the other hand with the current approach of `containAnyMethodsThat(predicate)`, you could even simply extend the syntax for `MethodsThat containAnyMethodsThat()` where you can use all the `MethodsThat` predicate methods again in a fluent way :thinking: 

Anyway, let me know what you think, if you like it we can simply squash it.

One thing I noticed though, we usually don't have any "Issue X" or similar in the commit messages. I typical follow the standard pattern of `if applied this commit will ${first_line_of_commit_message}`. So I would simply suggest something like

```
add `containAny{Members,Fields,Methods,...}That(..)` to rule syntax
```

nils-christian:
Hi, 

I totally agree with your approach. A more generic version of the method seems like a good idea to me. I will open a new PR with the squashed commits. Thank you very much!

hankem:
Just FYI: You could also have force-pushed your branch. 😉

codecholeric:
@nils-christian oh yeah, sorry if I didn't make this explicit, you could have simply done some `git rebase -i`, squash my commit onto yours with a common message you think fits and then force push your branch :slightly_smiling_face:

PR #551, opened 2021-03-13
Add generic interfaces

codecholeric:
This next step to fully support generics within ArchUnit adds support for generic interfaces. In particular

* add `Set<JavaType> JavaClass.getInterfaces()`, which will return a set of `JavaParameterizedType` for a parameterized generic interface and the raw types otherwise
* add all type arguments of generic interfaces to `JavaClass.directDependencies{From/To}Self`

Example: ArchUnit would now detect `String` as a Dependency of `class Foo implements Set<List<? super String>>`

This PR contains a breaking change regarding `JavaClass.getInterfaces()`. The same method previously returned the raw interface types (now available through `JavaClass.getRawInterfaces()`). I was pondering about going some way through deprecation to get there, but figured that in the end that would also be tedious for users, if the API I introduce newly for generics will change in the future quickly again to become `JavaClass.getInterfaces()`. And I'm strongly convinced it should be `JavaClass.getInterfaces()` that returns the generic types, because that is consistent to other places like `JavaClass.getSuperclass()`.

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
@@ -88,15 +88,15 @@ public class JavaClass implements JavaType, HasName.AndFullName, HasAnnotations<
         }
     });
     private final Set<JavaClass> interfaces = new HashSet<>();
-    private final Supplier<Set<JavaClass>> allInterfaces = Suppliers.memoize(new Supplier<Set<JavaClass>>() {
+    private final Supplier<Set<JavaClass>> allRawInterfaces = Suppliers.memoize(new Supplier<Set<JavaClass>>() {
         @Override
         public Set<JavaClass> get() {
             ImmutableSet.Builder<JavaClass> result = ImmutableSet.builder();
             for (JavaClass i : interfaces) {
                 result.add(i);
-                result.addAll(i.getAllInterfaces());
+                result.addAll(i.getAllRawInterfaces());
             }
-            result.addAll(superclass.getAllInterfaces());
+            result.addAll(superclass.getAllRawInterfaces());
             return result.build();
         }
     });
@@ -708,13 +708,13 @@ public class JavaClass implements JavaType, HasName.AndFullName, HasAnnotations<
     }
 
     @PublicAPI(usage = ACCESS)
-    public Set<JavaClass> getInterfaces() {
+    public Set<JavaClass> getRawInterfaces() {
         return interfaces;
     }
 
     @PublicAPI(usage = ACCESS)
-    public Set<JavaClass> getAllInterfaces() {
-        return allInterfaces.get();
+    public Set<JavaClass> getAllRawInterfaces() {
+        return allRawInterfaces.get();
     }
 
     /**
@@ -730,7 +730,7 @@ public class JavaClass implements JavaType, HasName.AndFullName, HasAnnotations<
         return ImmutableSet.<JavaClass>builder()
                 .add(this)
                 .addAll(getAllRawSuperclasses())
-                .addAll(getAllInterfaces())
+                .addAll(getAllRawInterfaces())
                 .build();
     }
 
@@ -1191,7 +1191,7 @@ public class JavaClass implements JavaType, HasName.AndFullName, HasAnnotations<
     @PublicAPI(usage = ACCESS)
     public boolean isAssignableTo(DescribedPredicate<? super JavaClass> predicate) {
         List<JavaClass> possibleTargets = ImmutableList.<JavaClass>builder()
-                .addAll(getClassHierarchy()).addAll(getAllInterfaces()).build();
+                .addAll(getClassHierarchy()).addAll(getAllRawInterfaces()).build();
 
         return anyMatches(possibleTargets, predicate);
     }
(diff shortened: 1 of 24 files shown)


archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
```diff
@@ -87,14 +89,14 @@
             return result.build();
         }
     });
-    private final Set<JavaClass> interfaces = new HashSet<>();
-    private final Supplier<Set<JavaClass>> allInterfaces = Suppliers.memoize(new Supplier<Set<JavaClass>>() {
+    private Interfaces interfaces = Interfaces.EMPTY;
+    private final Supplier<Set<JavaClass>> allRawInterfaces = Suppliers.memoize(new Supplier<Set<JavaClass>>() {
```
hankem, line 93: Do we additionally need `Supplier<Set<JavaType>> allnterfaces`?

Would we detect that `Child` _is a_ `List<String>` in the following example?
```java
    interface Parent<T> extends List<T> {
    }

    interface Child extends Parent<String> {
    }
```

hankem (approved):

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
hankem: I realized that we don't have the corresponding feature for superclasses either. So this could, of course, be added with another PR, if we want to have it at all (but IMO it would be very useful to handle `JavaTypeVariable`s in the library already).

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
```diff
@@ -1342,7 +1358,7 @@ private Superclass(Optional<JavaType> type) {
         }
 
         Set<JavaClass> getAllInterfaces() {
-            return rawType.isPresent() ? rawType.get().getAllInterfaces() : Collections.<JavaClass>emptySet();
+            return rawType.isPresent() ? rawType.get().getAllRawInterfaces() : Collections.<JavaClass>emptySet();
         }
 
         Superclass withRawType(JavaClass newRawType) {
```
rubieHess, line 1364: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Class names should start with an upper case letter (NM_CLASS_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-class-names-should-start-with-an-upper-case-letter-nm-class-naming-convention).
Class names should be nouns, in mixed case with the first letter of each internal word capitalized. Try to keep your class names simple and descriptive. Use whole words-avoid acronyms and abbreviations (unless the abbreviation is much more widely used than the long form, such as URL or HTML).

codecholeric: Are you talking about the camel casing of `Superclass`? That was actually a longer thought process :wink: You can compare this to the Reflection API: `Class.getSuperclass()`, so this is where it is coming from (besides that there is also "Superman", "supermarket" and "superconductor", i.e. "super" is in general not considered a separate word for alle examples I could find in the English language)

codecholeric: If you're not talking about that, then please elaborate a little more precisely where exactly you think the problem is.

hankem: I wonder where some automatic pattern matching has detected `class withRawType`... 😝

rubieHess: My mistake. It is a method function instead of a class.

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterGenericInterfacesTest.java
```diff
@@ -0,0 +1,545 @@
...
+            }
+        }
+        throw new AssertionError(String.format("Class %s has no interface of raw type %s", javaClass.getName(), rawType.getName()));
+    }
+
+    private interface SomeInterface {
+    }
+
+    @SuppressWarnings("unused")
+    private interface InterfaceWithSimpleTypeParameter<T> {
```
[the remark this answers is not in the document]
codecholeric: Yeah, it's "simple" because it has only an unbound type parameter :joy: But you're right, the "simple" doesn't add much value, I'm gonna drop it :wink:

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
```diff
@@ -1342,7 +1358,7 @@ private Superclass(Optional<JavaType> type) {
         }
 
         Set<JavaClass> getAllInterfaces() {
```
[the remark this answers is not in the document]
codecholeric: Yes, makes sense :+1:

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
codecholeric: Yeah, I agree that that would be nice, but I'll leave it open for a later time :wink: (and then for generic interfaces as well as superclasses)
The Reflection API doesn't offer that either AFAIK. I see it as nice to have, because all the information to retrieve the actual type arguments of super classes are present. It's just a little tedious right now, since the mapping has to be performed manually (i.e. associate type parameters with actual type arguments, then carry that mapping over to the class hierarchy).
For the Reflection API there are utilities like `org.springframework.core.GenericTypeResolver` though, that do this work for you, while for ArchUnit there is no such easy way, so sooner or later we should probably do it. But it's not essential to get this here out in my view :man_shrugging:

PR #555, opened 2021-03-20
Improve documentation of rules library

rweisleder:
I noticed that the Library API now contains more classes besides `GeneralCodingRules`. Currently, those neither have Javadoc nor are mentioned in the user guide. For users it may be difficult to discover the already available rules.

This is my attempt the improve both aspects of the documentation.


codecholeric:
Cool, thanks a lot for all your effort :smiley: That's certainly helpful to understand the available library functions :+1:

PR #568, opened 2021-04-13
Make message format configurable

### LINKED ISSUE #343: Make message format of failing rules configurable
Related to #301 and #214

We should provide a way to adjust the way rules report their violations. There is already `com.tngtech.archunit.lang.ConfiguredMessageFormat` which at the moment has only one configured instance. We might be able to make this configurable by a custom message format via `archunit.properties`:

```
messageFormat=com.my.CustomMessageFormat
```

For this to work `ConfiguredMessageFormat` should be turned into an interface with some sort of factory. We can then ponder if for #301 we should add an additional structured element like `solutions` to the result of an evaluated rule.

:warning: We should definitely consider to pass on a richer context than just raw text lines. It should be simple to pass on the full `ConditionEvent`, which would then also be able to provide the `correspondingObject` (i.e. `JavaMethodCall`, `Dependency`, ...), compare #361

Farbauti89:
This PR makes the message format configurable and therefore should resolve #343.
I followed the instructions of @codecholeric from [a comment in the issue](https://github.com/TNG/ArchUnit/issues/343#issuecomment-748499457)

Please let me know if there are any issues :)

Resolves #343


codecholeric:
Thanks a lot for your support!! :smile: And sorry for the delay :disappointed: 
When I looked at the code again I really couldn't tell anymore, why I introduced this second method `formatRuleText(..)`. It feels the input parameters were way to specific to ever allow a reasonable different processing, and also kind of confusing to be a reasonable API for extension. I have thus put a little pre-commit to inline this method and remove it.
Also in the end I figured "message format" is a kind of bad name, cause it's so utterly generic. Could be everything from the schema of an SQS message to a simple text format. Thus I've renamed it in the very end to "failure display format", because I felt that tells more what it is :man_shrugging: (hope I'm not the only person in the world that sees it that way :joy:)

codecholeric:
I have also squashed some of your commits that I felt should just be fix-ups, I hope you don't mind! (like fix the formatting that was broken in an earlier commit).

codecholeric:
Does it look good to you in the state that it is now? Is "failure display format" okay for you as a name? Or is this my own tunnel vision to see this as more descriptive? :wink:

Farbauti89:
> Does it look good to you in the state that it is now? Is "failure display format" okay for you as a name? Or is this my own tunnel vision to see this as more descriptive? 😉

Yes, the code looks good 👍 
I think `FailureDisplayFormat` is a quite reasonable name. As you stated it is much more precise than the old one.

codecholeric:
Okay, that's great :smile: Then I'll merge it and thanks a lot again for your support!!

PR #576, opened 2021-04-18
setup-java v1 -> v2

sullis:


codecholeric:
Thanks for the initiative! But Dependabot already tried it and failed -> #563
I looked into it a little and I think v2 does not support JDK 7 anymore, or am I seeing this wrongly? Because if yes we need to wait until we can drop Java 7 support...

codecholeric:
Gonna close this for now, since there doesn't seem to be any solution as long as we support JDK 7

PR #579, opened 2021-04-21
support primitve array as generic type argument

hankem:
With this change, the supertypes of a class such as 
```java
class Child extends ArrayList<int[]> implements List<int[]> {
}
```
are properly recognized as `JavaParameterizedType`s.


codecholeric:
Thanks a lot :smile: I wish I would have checked this PR before I started working on the generic field types :disappointed:
Anyway, I took the liberty to split out the part about using the constants, since a) I had done that commit already on the other branch and b) if I nitpick this is not part of solving the primitive array issue :wink: 
Also I added the same tests to the generics test, because this actually was also broken for type variable bounds with types parameterized with primitive arrays (e.g. `class Foo<T extends List<int[]>>`).
So thanks again for fixing this and also for detecting it!! One future bug report solved :+1:

PR #580, opened 2021-04-22
Fix typo

OLibutzki:


rweisleder:
Thanks a lot for your contribution!

In order to get this PR merged, please sign off your commit according to the [DCO](https://github.com/TNG/ArchUnit/blob/main/DCO). See also the [Contributing Guide](https://github.com/TNG/ArchUnit/blob/main/CONTRIBUTING.md#commits).
You can achieve this with: `git commit --amend -s`

OLibutzki:
Ok, my idea to fix it using the GitHub UI might have been a little bit naive ;-)

codecholeric:
Thanks a lot and sorry that the process is so complicated :wink: We just need this to be legally safe, so it's an automated check that we constantly follow. But thanks for your effort to fix this and make ArchUnit better! :smiley:

PR #584, opened 2021-04-25
correct same package being imported in the user guide example

GiorgadzeLuka:
The example in section 9.1.2 is supposed to show that several different package can be imported with `@AnalyzeClasses`, but the same package was provided twice.

Signed-off-by: Luka Giorgadze <luka.giorgadze94@gmail.com>


codecholeric:
Thanks a lot for the fix! :smiley:

PR #598, opened 2021-05-11
Fix wrongly repackaged dependencies

codecholeric:
When we upgraded Guava to version `guava:30.1-android` the dependency unfortunately contained some additional transitive dependencies that we overlooked to repackage. I have excluded all irrelevant transitive dependencies. We have sufficient test coverage to find any problems that could arise from excluding these dependencies (since they are also excluded from the Gradle build, not only the release artifacts).
I have also added an additional release artifact check that ensures we don't repackage any unexpected 3rd party libs in the future.

### DIFF
--- a/build-steps/release/check-uploaded-artifacts.gradle
+++ b/build-steps/release/check-uploaded-artifacts.gradle
@@ -103,6 +103,18 @@ def checkJavadocExists = { String artifactId ->
     assert getUploadedFile(artifactId, 'jar', 'javadoc') != null
 }
 
+def checkArtifactContent = { Project project, JarFile jarFile ->
+    checkManifest(project.name, jarFile.manifest)
+    if (project.repackagesAsm) {
+        println "Artifact ${project.name} is configured to repackage 3rd party libs -> checking existence of 3rd party package..."
+        checkThirdParty(jarFile)
+    } else {
+        println "Artifact ${project.name} is configured to not repackage 3rd party libs -> checking absense of 3rd party package..."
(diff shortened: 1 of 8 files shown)


build-steps/release/check-uploaded-artifacts.gradle
```diff
@@ -82,6 +83,18 @@ def checkNoThirdParty = { JarFile jarFile ->
     assert jarFile.getEntry('com/tngtech/archunit/thirdparty') == null: 'There exists a third party folder'
 }
 
+def checkNoIllegalJarEntries = { JarFile jarFile ->
```
hankem, line 86: Ah, very nice! 👍
Could we hook that into the regular build? If I understand correctly, this would only fail _after_ a release has already been published, right?

hankem (approved):

build-steps/release/check-uploaded-artifacts.gradle
codecholeric: I thought about that, too, but then refrained from it, because I was wondering if we should make the regular build more complex by that. But I guess it's worth it, since it's way better to catch this error at merge time, instead of at release time :+1: 
So I've adjusted all the jar file checks to run on every build and they are quite fast AFAIS.
"After a release has been published" is not completely correct though, we would catch it at the staging release, i.e. before it is really published, but at the very last moment before we would actually publish it to Maven Central. But anyway, as I said, now it should be caught at the moment something goes wrong with the dependencies :wink:

hankem (approved):

PR #623, opened 2021-06-19
onionArchitecture() now keeps ignored dependencies when overriding description

### LINKED ISSUE #622: onionArchitecture loses ignored dependencies when overriding description
When adding a because() or as() to an onionArchitecture() rule, all specified ignored dependencies are lost – which causes the architecture check to fail:

```java
    ArchRule rule = onionArchitecture()
            .domainModels("...")
            .adapter("... a name...", "....")
            .ignoreDependency(MySource.class, MyTarget.class)
            .because("docs/adr/001-hex-arch.md");
```

PR with test case and fix will follow shortly.

thmuch:
When adding a because() or as() to an onionArchitecture() rule, all specified ignored dependencies were lost – which caused the architecture check to fail.

This happened because the overridden OnionArchitecture.as() method did not copy the ignoredDependencies list.

Fixes #622

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/library/Architectures.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/Architectures.java
@@ -479,11 +479,13 @@ public final class Architectures {
                 String[] domainServicePackageIdentifiers,
                 String[] applicationPackageIdentifiers,
                 Map<String, String[]> adapterPackageIdentifiers,
+                List<IgnoredDependency> ignoredDependencies,
                 Optional<String> overriddenDescription) {
             this.domainModelPackageIdentifiers = domainModelPackageIdentifiers;
             this.domainServicePackageIdentifiers = domainServicePackageIdentifiers;
             this.applicationPackageIdentifiers = applicationPackageIdentifiers;
             this.adapterPackageIdentifiers = adapterPackageIdentifiers;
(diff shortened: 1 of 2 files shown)


hankem:
Thanks for fixing the bug (which I had introduced with #320... 🙈)!

The failing [Checks](https://github.com/TNG/ArchUnit/pull/623/checks) however prevent your commit from being merged, as it has not been signed-off according to the [DCO](https://github.com/TNG/ArchUnit/blob/master/DCO), cf. [CONTRIBUTING.md](https://github.com/TNG/ArchUnit/blob/master/CONTRIBUTING.md). Could you please amend your commit and force push your branch as [described](https://github.com/TNG/ArchUnit/blob/master/DCO)?

codecholeric:
Awesome, that's my favorite issues that come with a fix, thanks a lot :smile: :pray: Looks very good to me! I hope it's okay, but I slimmed down the test a tiny bit, because I felt we don't need to combine 3 separate ignore's to reproduce the problem, a single ignore is already sufficient to cause the problem and then it's a little bit less to read the test (except if I missed something, but it failed without your change and passed with your change :man_shrugging:)

thmuch:
Sure, that's absolutely ok – looks fine. I just copied the test case for ignored dependencies and forgot to minimize the test setup... 😉

PR #624, opened 2021-06-20
Colored plantuml components

### LINKED ISSUE #258: PlantUmlArchCondition should support colored components
At the moment using syntax like

```
[MyComponent] <<..mycomponent..>> #cceeff
```

will lead to 

```
com.tngtech.archunit.library.plantuml.IllegalDiagramException: There is no Component with name or alias = 'MyComponent'. Components must be specified separately from dependencies.
```

We should make this possible, since it is valid syntax and can provide visual guidelines.

johthor:
Add support for parsing colored components in `PlantUmlArchCondition`.

Resolves #258

### DIFF
--- a/archunit-example/example-junit4/src/test/resources/com/tngtech/archunit/exampletest/junit4/shopping_example.puml
+++ b/archunit-example/example-junit4/src/test/resources/com/tngtech/archunit/exampletest/junit4/shopping_example.puml
@@ -8,9 +8,9 @@ skinparam component {
 
 [Address] <<..address..>>
 [Customer] <<..customer..>>
-[Order] <<..order..>>
-[Products] <<..product..>>
-[Product Catalog] <<..catalog..>> as catalog
+[Order] <<..order..>> #white/PowderBlue
+[Products] <<..product..>> #white/PowderBlue
+[Product Catalog] <<..catalog..>> as catalog #PowderBlue
 [Product Import] <<..importer..>> as import
 
 ' Could be some random comment
--- a/archunit-example/example-junit5/src/test/resources/com/tngtech/archunit/exampletest/junit5/shopping_example.puml
+++ b/archunit-example/example-junit5/src/test/resources/com/tngtech/archunit/exampletest/junit5/shopping_example.puml
@@ -8,9 +8,9 @@ skinparam component {
 
 [Address] <<..address..>>
 [Customer] <<..customer..>>
-[Order] <<..order..>>
-[Products] <<..product..>>
-[Product Catalog] <<..catalog..>> as catalog
+[Order] <<..order..>> #white/PowderBlue
+[Products] <<..product..>> #white/PowderBlue
+[Product Catalog] <<..catalog..>> as catalog #PowderBlue
 [Product Import] <<..importer..>> as import
 
 ' Could be some random comment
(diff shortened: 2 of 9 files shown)


codecholeric:
Wow, thanks for all the hard work, I really appreciate it :smiley: 
What I'm wondering about though is how exactly this part of the API is going to be used? Originally my idea was just "it should not crash if someone specifies a color and still be able to parse the abstract components" :sweat_smile:. Do you have any use case where you would actually use the colors for your ArchUnit rules? :thinking: (I'm just asking since now the colors are part of the model and can be queried, right? At which point/in which case would that actually happen then?)

johthor:
When implementing the color parsing I thought "Now that I have the color information it would be shame to let it go to waste". Therefore I extended the model with the color for someone else to find a good use for it and create the needed rules. :smile: 

But now after some consideration I can't think of any good use case where additional stereotypes wouldn't be the better or cleaner choice. Maybe the color would help with understanding but would not encode the important information itself (e.g. marking legacy code or core components). Therefore the color doesn't need to be queryable from the model.

I will remove the color from the model again to reduce unnecessary code and keep the model small. But I would still extend the `PLANTUML_COMPONENT_PATTERN` with a `COLOR_FORMAT` without transferring the color information into the model to document the color awareness.

codecholeric:
Yes, I know that impulse :joy: But I also agree that the cheapest code to maintain is no code, so if we can't think of any good use case for the model extension I would also vote to simplify it. Having a specific part of the regex pattern for color is good though I think, because in the past I was sometimes too lenient and the pattern caught more than I had expected :+1:

johthor:
I have removed the color information from the component model again and added a hopefully helpful comment.

But I'm not sure how to describe the color awareness best in documentation as it is no longer like the alias.
It is optional but can't be analyzed via conditions later.

archunit-example/example-junit4/src/test/resources/com/tngtech/archunit/exampletest/junit4/shopping_example.puml
```diff
@@ -8,9 +8,9 @@ skinparam component {
 
 [Address] <<..address..>>
 [Customer] <<..customer..>>
-[Order] <<..order..>>
-[Products] <<..product..>>
-[Product Catalog] <<..catalog..>> as catalog
+[Order] <<..order..>> #white/PowderBlue
```
codecholeric, line 11: I didn't even know that these crazy gradients were supported :joy: :+1:

codecholeric:
Looks good now, thanks a lot :smile: I've adjusted it a little (without first getting back to you, sorry, but I thought you want it in the next release), in particular

* I really don't want to support quoted aliases :wink: The spec unfortunately seems super fuzzy, I can find examples with quotes and without, I could find this issue that suggests it would not be allowed https://forum.plantuml.net/8696/alias-quote-bug, but obviously it works. But when you quote it, it displays the alias as component name, so all that is so much complexity and fuzziness... We only support a subset of diagram syntax anyway, so I just wrote into the user guide that aliases must not be quoted :man_shrugging: Let's see if someone will ever desperately want this :wink: 
* I slimmed down the pattern a little, because I don't think we need to name the group then, etc.

codecholeric:
BTW. I think what you wrote in the user guide is fine, and I think it's self-explanatory enough in the code as well. You can see it's no capturing group and it's evident it's not in the model, so I think this is all fine :+1: I mean the user guide only says "components may have an optional color", so that's true :man_shrugging: And color directives in dependencies are also ignored by the model...

johthor:
I'm all good with your adjustments. Thanks again for integrating my submission.

codecholeric:
Thank **you** for your contribution! :smiley:

PR #626, opened 2021-07-01
Evaluate DisplayNameGenerator.ReplaceUnderscores

thmuch:
I like to use JUnit 5's feature for automatic display name generation, especially the strategy for replacing underscores in test method names by spaces. This way, I can phrase the test names like a human readable sentence – without having to use ```@DisplayName``` all the time.

Unfortunately, this doesn't work with ArchUnit yet. On the one hand, ArchUnitTestDescriptor isn't derived (or does not use) the specialised Jupiter descriptor classes. But refactoring this might be too much of a change. And on the other hand, JUnit's DisplayNameGenerator API isn't really suited for ArchUnit's way of dealing with tests, especially the field based tests.

So I tried a really simple approach: Evaluating only the one DisplayNameGenerator strategy that makes sense most IMHO for ArchUnit (ReplaceUnderscores) with an implementation not (syntactically) derived from JUnit's implementation of the DisplayNameGenerator API.

With the changes in this PR it's possible to write

```java
@AnalyzeClasses(packages = "some.dummy.package")
@DisplayNameGeneration(DisplayNameGenerator.ReplaceUnderscores.class)
public class SomeTestClass { ... }
```

for field and method based ArchUnit tests.

Feel free to reject this PR if it is too little benefit for too much change. Or if you think this is crude use of a public API. But I tried to be as little invasive as possible, while still solving my main use case.

### DIFF
--- a/archunit-junit/junit5/engine/src/test/java/com/tngtech/archunit/junit/ArchUnitTestEngineTest.java
+++ b/archunit-junit/junit5/engine/src/test/java/com/tngtech/archunit/junit/ArchUnitTestEngineTest.java
@@ -12,6 +12,7 @@ import java.util.Set;
 import java.util.stream.Stream;
 
 import com.google.common.collect.ImmutableSet;
+import com.tngtech.archunit.ArchConfiguration;
 import com.tngtech.archunit.core.domain.JavaClasses;
 import com.tngtech.archunit.junit.ArchUnitTestEngine.SharedCache;
 import com.tngtech.archunit.junit.testexamples.ClassWithPrivateTests;
@@ -47,6 +48,7 @@ import com.tngtech.archunit.junit.testexamples.subtwo.SimpleRules;
 import com.tngtech.archunit.junit.testexamples.wrong.WrongRuleMethodNotStatic;
 import com.tngtech.archunit.junit.testexamples.wrong.WrongRuleMethodWrongParameters;
 import com.tngtech.archunit.junit.testutil.MockitoExtension;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
@@ -749,13 +751,6 @@ class ArchUnitTestEngineTest {
                     .collect(onlyElement());
         }
 
-        private TestDescriptor getOnlyTest(TestDescriptor descriptor) {
-            TestDescriptor testClass = getOnlyElement(descriptor.getChildren());
-            TestDescriptor ruleDescriptor = getOnlyElement(testClass.getChildren());
-            assertThat(ruleDescriptor.getType()).isEqualTo(TEST);
-            return ruleDescriptor;
-        }
-
         private Stream<TestDescriptor> getArchRulesDescriptorsOfOnlyChild(TestDescriptor descriptor) {
             TestDescriptor testClass = getOnlyElement(descriptor.getChildren());
             Set<? extends TestDescriptor> archRulesDescriptors = testClass.getChildren();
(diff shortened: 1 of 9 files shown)


thmuch:
Hm... thinking about it... Is it ok that we'd required all JUnit 5 users to use at least JUnit 5.4 (when DisplayNameGenerator became an experimental API) or JUnit 5.7 (when die API was declared stable)? Because now the Jupiter API is an implementation dependency in ArchUnit's JUnit 5 engine artifact.

thmuch:
Another option for implementing this feature would be to create synthetic Method objects for the test fields (and strip the JavaClasses param from test methods). With that approach, we probably could (re)use all existing DisplayNameGenerator implementations. But that would feel somewhat hacky, too.
Maybe we keep this option in mind if the simple implementation isn't good enough any more.

codecholeric:
Thanks a lot for the initiative! :smiley: You're right that the test method names themselves aren't super pretty. 
However, as you noted, the philosophy so far was to be completely independent of the Jupiter API and only depend on the JUnit platform :thinking: 
I would also like to keep it that way, because it is one less dependency to care about and more important the mental model of ArchUnit's tests just doesn't completely match the Jupiter model (which is why I went a separate road back then, instead of integrating it into the Jupiter engine as extension, etc.)
That being said we could still think about a custom implementation within the `ArchUnitTestEngine`, but it would need to be an ArchUnit API then as well I think, and completely independent of the Jupiter specifics :thinking: 
What do you think about it? We could also ponder if we shouldn't just always replace underscores by spaces within display names of the tests and not even make it customizable :thinking: Because I can't imagine anybody would complain about that, i.e. not seeing snake case but nicely formatted test display names. But you never know :man_shrugging:

thmuch:
+1 for keeping ArchUnit as independant from Jupiter as possible, especially if it was a deliberate decision (although you have different modules for JUnit 4 and 5 already; and users of ArchUnit and JUnit 5 probably will already have a dependency to the Jupiter API already)

If we make replacing underscores in test names an ArchUnit custom feature, maybe a property would be enough? Like ```displayName.replaceUnderscores```? I'd make ```false``` the default, though, as using this feature should be a deliberate decision IMHO.

thmuch:
Now underscores in test names are replaced by spaces if the property ```displayName.replaceUnderscores``` is set to ```true```.

If you think this is a good approach, I could try to integrate this kind of display name generation in archunit-junit/junit4. Where to place ```DisplayNameResolver``` best, in the "base" package of the "archunit" module? And make it ```@Internal``` then?

archunit-junit/junit5/engine/src/test/java/com/tngtech/archunit/junit/ArchUnitTestEngineTest.java
```diff
@@ -1122,6 +1116,61 @@ void with_reason() {
         }
     }
 
+    @Nested
+    class GeneratesDisplayName {
+        @Test
+        void for_field_by_replacing_underscores_with_blanks_if_property_is_set_to_true() {
+            EngineDiscoveryTestRequest discoveryRequest = new EngineDiscoveryTestRequest().withClass(SimpleRuleField.class);
+
+            ArchConfiguration.get().setProperty(DisplayNameResolver.DISPLAYNAME_REPLACE_UNDERSCORES_PROPERTY_NAME, "true");
```
codecholeric, line 1125: If we do this we should also have some

```
@AfterEach
void resetConfiguration() {
    ArchConfiguration.get().reset();
}
```

Otherwise other tests might start to behave surprisingly / flaky :wink:

archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/DisplayNameResolver.java
```diff
@@ -0,0 +1,40 @@
...
+
+    static String determineDisplayName(String elementName) {
+        if (replaceUnderscores()) {
+            return underscoresReplacedBySpaces(elementName);
+        } else {
+            return elementName;
+        }
+    }
+
+    private static String underscoresReplacedBySpaces(String elementName) {
```
codecholeric, line 32: Should we make the property also `replaceUnderscoresBySpaces`? I like this method name, because it says exactly what it does. The property name just says `replaceUnderscores`, but not with what :thinking:

archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/DisplayNameResolver.java
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.tngtech.archunit.junit;
+
+import com.tngtech.archunit.ArchConfiguration;
+
+final class DisplayNameResolver {
+
+    static final String DISPLAYNAME_REPLACE_UNDERSCORES_PROPERTY_NAME = "displayName.replaceUnderscores";
```
codecholeric, line 22: I would prefix the property name by `junit.` if it is for both JUnit versions and `junit5.` if it is only for JUnit 5. Because it does not influence the core of ArchUnit like most other properties.

codecholeric (commented):
I like this approach! :+1: Added some minor comments, what do you think about them? Also I think it would be nice if it would work consistently for JUnit 4 and 5, but up to you if you want to invest the effort for now :wink: 
If you would add the support also to JUnit 4, then at the moment those modules share a common package anyway, so `DisplayNameResolver` still wouldn't have to be public (this will be a problem sooner or later, since we're violating the Java 9 modules contract here, but that is to be solved at a different time and there is more to do then :wink:)
In other words there already is a common module also containing the package `com.tngtech.archunit.junit`.

archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/DisplayNameResolver.java
thmuch: Thanks for your feedback! All suggestions seem reasonable.
I'll amend the pull request next week (hopefully), including JUnit 4 support.

thmuch:
@codecholeric: This PR should be feature complete, finally. Unless you have some feedback, of course 🙃

Issue #440, opened 2020-09-29
Assert that all fields that are primitive type or collection of primitive type


GeeJoe:
I want to make sure that all fields of a Class should be primitive type or collection of primitive type:
```java
class MyClass {
 String a;
 int b;
 float c;
 int[] d;
 List<String>;
}
```
How can I achieve that?
I've tried the following code to make part of it works
```java
javaField.rawType.isPrimitive // check whether the field is primitive type
javaField.rawType.simpleName == "String" // check whether it's String
```
But how can I check whether the field is an Array of primitive type, such as `int[]` `byte[]`, or a List of primitive type, such as `List<String>`, etc?

In other words:
- How can I know the generic type inside a `List<T>`?
- How to describe an Array of a primitive type?

hankem:
* ArchiUnit 0.14.1 unfortunately doesn't support generic types yet (#115). 🤷
* For array types, you can easily call `getComponentType()` until you find the "base component type" (e.g. `int[][]` → `int[]` → `int`). Based on that, you can use something like:
```java
        fields()
            .that().areDeclaredIn(MyClass.class)
            .should().haveRawType(DescribedPredicate.describe("that is primitive or an array of primitives", type -> {
                while (type.isArray()) {
                    type = type.getComponentType();
                }
                return type.isPrimitive();
            }));
```

(`JavaClass`  could probably get a `getBaseComponentType()` convenience function at some point, as this is often needed when working with arrays.)

GeeJoe:
When will ArchiUnit support generic types？

hankem:
I'm sorry, I cannot promise a date. I just know that it is a really tough issue... 🤷
However, 💪 @codecholeric has recently mentioned in https://github.com/TNG/ArchUnit/issues/423#issuecomment-701025933 that
> [he] really want[s] to push for the Generics support with all the (very limited) capacity [he has] at the moment!

GeeJoe:
🤝

Issue #144, opened 2019-02-07
Layered architecture ignores violations in type parameters


stefanroeck:
Given:
- A layered architecture with 2 packages "presentation" and "shared"
- presentation might access shared but not vice versa
- class in shared references class in presentation as type parameter of java.util.list

Expected:
- Error because of layered architecture violation

Actual:
- No error. 

```
package com.archunitbug.presentation;
public class ExceptionDetails {}
```

```
package com.archunitbug.shared;
import java.util.List;
import com.archunitbug.presentation.ExceptionDetails;

public class ExceptionWithBackReference extends RuntimeException {
    private List<ExceptionDetails> details;
    public ExceptionWithBackReference(List<ExceptionDetails> details) {
        this.details = details;
    }
}

```
Test:
```
import static com.tngtech.archunit.library.Architectures.layeredArchitecture;
import org.junit.Test;
import com.tngtech.archunit.core.domain.JavaClasses;
import com.tngtech.archunit.core.importer.ClassFileImporter;
import com.tngtech.archunit.core.importer.ImportOption;
import com.tngtech.archunit.core.importer.ImportOptions;
import com.tngtech.archunit.library.Architectures.LayeredArchitecture;
public class ArchunitLayersException {
    @Test
    public void enforceAllNcrLayersTopDown() {
        JavaClasses classes = new ClassFileImporter(new ImportOptions().with(ImportOption.Predefined.DONT_INCLUDE_TESTS)).importPackages("com.archunitbug");
        LayeredArchitecture rule = layeredArchitecture() //
                .layer("presentation").definedBy("com.archunitbug.presentation") //
                .layer("shared").definedBy("com.archunitbug.shared") //
                .whereLayer("presentation").mayNotBeAccessedByAnyLayer()//
                .whereLayer("shared").mayOnlyBeAccessedByLayers("presentation");

        rule.check(classes);
    }
}
```

hankem:
I believe that ArchUnit (in its current version 0.9.3) does not yet handle type parameters, cf. #115.

If `ExceptionWithBackReference` had an `ExceptionDetails` field instead of `List<ExceptionDetails>`, your `LayeredArchitecture` test would fail as expected.

codecholeric:
Yes, this is correct, the current version cannot detect this, since ArchUnit only sees a raw reference to `java.util.List`. As soon as #115 is done, dependencies of a `JavaClass` can be extended with type parameters and this violation will then be detected.

PR #640, opened 2021-07-16
Add generic method parameter types

codecholeric:
As last step to fully support Generics within ArchUnit this adds support for generic method/constructor parameter types. In particular

* `JavaMethod.getParameterTypes()` now returns a `List` of `JavaParameterizedType` for parameterized generic method/constructor parameter types and the raw types otherwise
* all type arguments of generic method parameter types are added to `JavaClass.directDependencies{From/To}Self`

Example: ArchUnit would now detect `String` and `File` as `Dependency` of a method declaration like

```
class SomeClass {
    void someMethod(Set<List<? super String>> first, Map<?, ? extends File> second) {
    }
}
```

Note that analogously to the Java Reflection API `JavaConstructor.getParameterTypes()` and `JavaConstructor.getRawParameterTypes()` do not behave exactly the same for inner classes. While `JavaConstructor.getRawParameterTypes()` contains the enclosing class as first parameter type, `JavaConstructor.getParameterTypes()` does not contain it if the constructor has generic parameters. On the other hand it just returns the same list of raw parameter types if the constructor is non-generic. This might surprise some users, but I decided to stick to the same behavior as the Reflection API, because this has always been the strategy and the solution can never truly satisfy all assumptions a user might have.

Resolves: #115
Resolves: #144
Resolves: #440

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/AccessTarget.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/AccessTarget.java
@@ -318,6 +318,12 @@ public abstract class AccessTarget implements HasName.AndFullName, CanBeAnnotate
         }
 
         @Override
+        @SuppressWarnings({"unchecked", "rawtypes"}) // cast is okay, since this list can only be used in a covariant way (immutable)
+        public List<JavaType> getParameterTypes() {
+            return (List) parameters;
+        }
+
+        @Override
         @PublicAPI(usage = ACCESS)
         public List<JavaClass> getRawParameterTypes() {
             return parameters;
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/properties/HasParameterTypes.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/properties/HasParameterTypes.java
@@ -21,6 +21,8 @@ import com.google.common.collect.ImmutableList;
 import com.tngtech.archunit.PublicAPI;
 import com.tngtech.archunit.base.DescribedPredicate;
 import com.tngtech.archunit.core.domain.JavaClass;
+import com.tngtech.archunit.core.domain.JavaType;
+import com.tngtech.archunit.core.domain.JavaTypeVariable;
 
 import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
 import static com.tngtech.archunit.base.DescribedPredicate.equalTo;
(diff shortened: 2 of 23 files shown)


OLibutzki:
I'm highly interested in getting this PR to be merged. 🙂

Are there any more obstacles to overcome before this can be merged?

archunit/src/main/java/com/tngtech/archunit/core/domain/properties/HasParameterTypes.java
```diff
@@ -30,6 +31,9 @@
 
 public interface HasParameterTypes {
 
+    @PublicAPI(usage = ACCESS)
+    List<JavaType> getParameterTypes();
```
hankem, line 35: Should we add some JavaDoc to disambiguate `getParameterTypes()` from `getTypeParameters()` (which also exists in `JavaCodeUnit`)?

archunit/src/main/java/com/tngtech/archunit/core/domain/AccessTarget.java
```diff
@@ -317,6 +317,12 @@ private Functions() {
             this.returnType = builder.getReturnType();
         }
 
+        @Override
+        @SuppressWarnings({"unchecked", "rawtypes"}) // cast is okay, since this list can only be used in a covariant way (immutable)
+        public List<JavaType> getParameterTypes() {
+            return (List) parameters;
```
hankem, line 323: As far as I could see, we can replace the implementation with:
```
            throw new RuntimeException("No test breaks if we do this!");`
```
😜

hankem (approved):
I'm sorry that it took me so long to review this PR!

codecholeric:
> I'm sorry that it took me so long to review this PR!

No worries, I know that it's not in all phases of life super easy to spend a lot of time on OSS :wink: You're always giving great support to ArchUnit, so thanks a lot!

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterGenericCodeUnitParameterTypesTest.java
```diff
@@ -0,0 +1,1246 @@
...
+    }
+
+    @Test
+    @UseDataProvider
+    public void test_imports_generic_code_unit_parameter_type_with_array_type_argument(JavaCodeUnit codeUnit, Class<?> expectedParameterTypeErasure) {
+        JavaType genericMethodParameterType = codeUnit.getParameterTypes().get(0);
+
+        assertThatType(genericMethodParameterType).as("generic parameter type")
+                .hasErasure(expectedParameterTypeErasure)
+                .hasActualTypeArguments(String[].class);
```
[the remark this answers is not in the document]
codecholeric: Yeah, it's a scoping problem. But yes, maybe I went a little overboard there :joy: I defined three reusable types with one, two and three type params, that makes things simpler :+1:

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterGenericCodeUnitParameterTypesTest.java
```diff
@@ -0,0 +1,1246 @@
...
+        }
+
+        JavaConstructor constructor = new ClassFileImporter()
+                .importClasses(
+                        LocalClassThatWillHaveEnclosingClassAsFirstRawConstructorParameter.class,
+                        getClass(), GenericParameterType.class, String.class)
+                .get(LocalClassThatWillHaveEnclosingClassAsFirstRawConstructorParameter.class)
+                .getConstructor(getClass(), GenericParameterType.class);
+
+        // raw parameter types
```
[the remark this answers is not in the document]
codecholeric: Yeah, I guess that's true, don't know exactly why I thought these were necessary originally :wink: Maybe for "generic parameters", cause the method doesn't say "getGenericParameterTypes" and could theoretically return raw types, if the parameters are not generic :man_shrugging:

archunit/src/main/java/com/tngtech/archunit/core/domain/AccessTarget.java
codecholeric: True, it's not rocket science :wink: , but since this is a library and also for documentation it makes sense to still test this I guess.

archunit/src/main/java/com/tngtech/archunit/core/domain/properties/HasParameterTypes.java
codecholeric: Good idea, I added some Javadoc with examples! :+1:

PR #646, opened 2021-08-10
Enable CI release

NilsOliverLinden:
With this PR ArchUnit can be released via a GitHub Action.
In the release script release.sh all necessary Gradle commands are executed. This script is called by the CI pipeline.
Central change for the CI release process is the adaption of Gradle plugins to be able to obtain necessary credentials via Gradle properties. 
For releasing the artifacts, a new plugin is used that allows to reference the staging repository id in Sonatype.

### DIFF
--- a/build-steps/release/check-uploaded-artifacts.gradle
+++ b/build-steps/release/check-uploaded-artifacts.gradle
@@ -3,6 +3,9 @@ import java.time.LocalDateTime
 import java.util.jar.JarFile
 import java.util.jar.Manifest
 
+apply from: scriptRelativePath(this, 'vcs-utils.gradle')
+apply from: scriptRelativePath(this, 'archunit-examples-utils.gradle')
+
 File scriptRoot = currentScriptRootOf this
 def rootUrl =  {
     if (!project.hasProperty('tngRepoId')) {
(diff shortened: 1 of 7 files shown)


release/release.sh
```diff
@@ -0,0 +1,77 @@
...
+echo "Old version is $OLD_VERSION. Releasing version $NEW_VERSION..."
+
+echo Updating version in build.gradle, README.md and docs...
+sed -i -e s/version\ =.*/version\ =\ \'$NEW_VERSION\'/ build.gradle
+sed -i -e s/$OLD_VERSION/$NEW_VERSION/ README.md ./docs/_data/navigation.yml ./docs/_pages/getting-started.md
+
+echo Create release news...
+./gradlew createReleaseNews
+
+if [ -n "$(git status --porcelain)" ]; then
```
codecholeric, line 63: Shouldn't we fail, if this is not the case? :thinking: I mean, it would be very suspicious and if the branch would be already prepared the `createReleaseNews` task would fail...

release/release.sh
```diff
@@ -0,0 +1,77 @@
...
+    echo Commiting version change
```
codecholeric, line 64:
```suggestion
    echo Committing version change
```

release/release.sh
```diff
@@ -0,0 +1,77 @@
...
+    git add build.gradle README.md ./docs -- ':!docs/**.png'
```
codecholeric, line 65: I know the thing with the user guide images that are checked in is not optimal, but I would just commit those also to be safe. Sometimes they also change a pixel in size or something and the html changes as well :man_shrugging:

release/release.sh
```diff
@@ -0,0 +1,77 @@
...
+    git commit -m "Update version to $VERSION"
```
codecholeric, line 66: I think I usually call the commit "prepare release x", because it also creates the release news.

release/release.sh
```diff
@@ -0,0 +1,77 @@
...
+fi
+
+echo Building and testing...
+./gradlew clean publishToMavenLocal runMavenTest -PallTests
```
codecholeric, line 70: I'm not sure if this is really necessary. All builds before pass the extensive CI runs (be it PR or main) and the release script only does non-code changes. From my point of view, if it can run the Gradle publish task it should be fine :man_shrugging:

build-steps/release/check-uploaded-artifacts.gradle
```diff
@@ -4,12 +4,13 @@ import java.util.jar.JarFile
...
+    tngRepoId = ""
+}
+
 def rootUrl =  {
-    if (!project.hasProperty('tngRepoId')) {
-        throw new IllegalArgumentException(
-                'You must pass the repo id (see Sonatype Repository Manager -> comtngtech-${repoId}) as parameter, e.g. -P tngRepoId=1162')
-    }
-    "https://oss.sonatype.org/service/local/repositories/comtngtech-${tngRepoId}/content/com/tngtech/archunit"
+    "https://oss.sonatype.org/service/local/repositories/${tngRepoId}/content/com/tngtech/archunit"
```
codecholeric, line 13: How comes the `comtngtech-` is obsolete now? :thinking:

release/release.sh
NilsOliverLinden: You are right indeed.

build-steps/release/check-uploaded-artifacts.gradle
NilsOliverLinden: The string `comtngtech-` is part of the repo-id the plugin provides. Specifying it explicitly would double it and the URL would be wrong.

release/release.sh
NilsOliverLinden: Fine for me :)

release/release.sh
NilsOliverLinden: Okay :)

PR #672, opened 2021-09-17
GitHub Action to check whether ASM is up-to-date

hankem:
As dependabot does not recognize how dependencies are encoded in this gradle project and as we don't want to miss a new version of the ASM library, this PR adds a GitHub Action that regularly checks whether ASM is up-to-date and creates an issue or comment on that issue if it isn't.


hankem:
See hankem/ArchUnit-check-dependencies#1 for an "end-to-end test".

codecholeric:
Thanks a lot, this will be very useful :smile:

PR #677, opened 2021-09-27
setup-java v2

sullis:


hankem:
Note that @dependabot has also attempted to bump `actions/setup-java` with #660.
For both PRs, it seems that the Java version is not yet picked up correctly by the integration tests.

codecholeric:
Also note that you created the exact same PR as #576, which I back then closed and commented that Dependabot had already tried the update and that it doesn't work as long as we're still supporting Java 7 :wink: Which is still the case... We're likely gonna drop Java 7 support with version 1.0, then we can also do all these upgrades.
Gonna close this PR for now...

PR #688, opened 2021-10-09
fix `URISyntaxException` if `ClassLoader` returns unencoded URL chars

### LINKED ISSUE #683: Can't handle spaces in dependencies
We use ArchUnit 0.21.0 on an OSGi system. To load our java classes we just point to a rootdir, pretty much like this:

```java
private static JavaClasses getJavaClasses() {
	return new ClassFileImporter()
			.withImportOption(DO_NOT_INCLUDE_ARCHIVES)
			.importPath("rootPath");
}
```

OSGi being OSGi, the code above will only find the class files for our product. The dependencies reside in different directories such as `~/.m2/` or `$workspace/.metadata.plugins/`. When loading a dependency class the OSGi classloader will know to check such directories.

The stacktrace we're encountering is the following:

```
ClassResolverFromClasspath.tryGetUriOf(String) line: 62
ClassResolverFromClasspath.tryResolve(String) line: 49
ImportedClasses.getOrResolve(String) line: 57
ImportedClasses.ensurePresent(String) line: 69
ClassGraphCreator.ensureMemberTypesArePresent() line: 117
ClassGraphCreator.complete() line: 106
ClassFileProcessor.process(ClassFileSource) line: 63
ClassFileImporter.importLocations(Collection<Location>) line: 336
ClassFileImporter.importPaths(Collection<Path>) line: 158
ClassFileImporter.importPaths(String...) line: 133
ClassFileImporter.importPath(String) line: 112
BaseArchUnitTest.getJavaClasses() line: 56
```

Line 56 in `BaseArchUnitTest` being the `.importPath("rootPath");` in the code above.

For most of our `typeName`s in `ImportedClasses.getOrResolve(String)`, the javaClass != null and thus we have no need to resolve it. However, for dependencies such as OSGi's `Version.class` (org.osgi.framework.Version.class) we need to go look into the directories mentioned above.

This results in `resource` in `ClassResolverFromClasspath.tryGetUriOf(String)` being of the form `C:/Users/David Name/sigasi-development-project-master/ws/.metadata/.plugins/org.eclipse.pde.core/.bundle_pool/plugins/org.eclipse.osgi_3.16.300.v20210525-1715.jar!/org/osgi/framework/Version.class`. Note the space in the user name.

When we then try to do `resource.toURI()`, we get the error below.

Conclusion: The URL received from the classloader should first be encoded, and only then transformed into a URI

```
java.net.URISyntaxException: Illegal character in opaque part at index 24: jar:file:/C:/Users/David Name/sigasi-development-project-master/ws/.metadata/.plugins/org.eclipse.pde.core/.bundle_pool/plugins/org.eclipse.osgi_3.16.300.v20210525-1715.jar!/org/osgi/framework/Version.class
```

codecholeric:
It seems like some OSGI `ClassLoaders` would return resource URLs that don't have correctly encoded special characters (in particular spaces). This leads to `URISyntaxException` if `url.toURI()` is called on those `URLs`.
This PR makes the `URL` to `URI` conversion more resilient by falling back to encoding the `URL` if we run into this issue.

Resolves #683

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/importer/resolvers/ClassResolverFromClasspath.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/importer/resolvers/ClassResolverFromClasspath.java
@@ -57,9 +57,19 @@ public final class ClassResolverFromClasspath implements ClassResolver {
             return Optional.empty();
         }
         try {
-            return Optional.of(resource.toURI());
+            return Optional.of(toUri(resource));
         } catch (URISyntaxException e) {
             throw new ArchUnitException.LocationException(e);
+        }
+    }
+
+    private URI toUri(URL resource) throws URISyntaxException {
+        try {
+            return resource.toURI();
+        } catch (URISyntaxException e) {
+            // In case the ClassLoader returns a URL with unencoded characters (e.g. spaces), this will correctly encode them.
+            // This was added to solve https://github.com/TNG/ArchUnit/issues/683 where some OSGI ClassLoader would not encode spaces correctly.
+            return new URI(resource.getProtocol(), resource.getHost(), resource.getPath(), null);
         }
     }
 }
--- a/archunit/src/test/java/com/tngtech/archunit/core/importer/resolvers/ClassResolverFromClassPathTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/core/importer/resolvers/ClassResolverFromClassPathTest.java
@@ -1,20 +1,37 @@
 package com.tngtech.archunit.core.importer.resolvers;
 
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+
+import com.tngtech.archunit.base.Function;
 import com.tngtech.archunit.base.Optional;
 import com.tngtech.archunit.core.domain.JavaClass;
 import com.tngtech.archunit.core.importer.resolvers.ClassResolver.ClassUriImporter;
 import com.tngtech.archunit.testutil.TestUtils;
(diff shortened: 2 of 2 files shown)


archunit/src/test/java/com/tngtech/archunit/testutil/TestUtils.java
```diff
@@ -63,4 +67,21 @@ public int compare(HasName o1, HasName o2) {
             }
         });
     }
+
+    public static URI toUri(URL url) {
+        try {
+            return url.toURI();
+        } catch (URISyntaxException e) {
+            throw new RuntimeException(e);
```
kaebiscs, line 75: I think that introducing the fix here as well might avoid the same problem as the `OutsideOfClassPathRule` uses file system folders. Not completely sure if the caller from `ClassFileImporterTest` is using absolute or relative paths.

codecholeric: I think in this case it's not so bad, because the problem only occurs with an OSGI `ClassLoader` AFAIK. But test utils are only used to build ArchUnit itself. And if anybody tries to build ArchUnit using an OSGI `ClassLoader`... Well... :wink: I don't think it's a case we have to guard against.

hankem:
> Did you check the state before as well?

Yes: I got test failures when the directory contained spaces.

codecholeric:
Ah, okay! But that's a different problem :slightly_smiling_face: That relates to the first commit simply fixing the test infrastructure. The bugfix refers to production code, where the problem only occurs with the OSGI `ClassLoader`

hankem:
I know, but that part was unfortunately the only one I could reasonably test... 🤷‍♂️

PR #722, opened 2021-11-19
Resolve call targets like Reflection API

codecholeric:
Instead of having a complicated `JavaMethodCallTarget.resolve()` API that returns a `Set<JavaMethod>` we can follow the same logic as the Reflection API, which always offers one `Method` matching a certain signature. This should make the vast majority of cases a lot simpler, and for diamond scenarios we are following an established precedence logic. Thus, the argument why method x is chosen in a specific case is a lot simpler ("we just follow the established precedence of the Java Reflection API").

Note that this also fixes a bug in the field resolution logic. Earlier accesses of constants in interfaces of parent classes were not resolved correctly (i.e. if we find a field in an interface it should have precedence).

### DIFF
--- a/archunit/src/jdk15test/java/com/tngtech/archunit/core/importer/testexamples/methodresolution/DeterminesMethodAnalogouslyToReflectionApi.java
+++ b/archunit/src/jdk15test/java/com/tngtech/archunit/core/importer/testexamples/methodresolution/DeterminesMethodAnalogouslyToReflectionApi.java
@@ -1,0 +1,874 @@
+package com.tngtech.archunit.core.importer.testexamples.methodresolution;
+
+import java.io.Serializable;
+
+@SuppressWarnings({"unused", "ConstantConditions"})
+public class DeterminesMethodAnalogouslyToReflectionApi {
+
+    public static class EquivalentMethodsAreChosenDepthFirst {
+
+        public static class LeftLeftHasPrecedence {
+            interface LeftLeftGrandParent {
+                @ExpectedMethod
+                void target();
+            }
+
+            interface LeftRightGrandParent {
+                void target();
+            }
+
+            interface LeftParent extends LeftLeftGrandParent, LeftRightGrandParent {
+            }
+
+            interface RightLeftGrandParent {
+                void target();
+            }
+
+            interface RightRightGrandParent {
+                void target();
+            }
+
+            interface RightParent extends RightLeftGrandParent, RightRightGrandParent {
+            }
+
+            interface Child extends LeftParent, RightParent {
+            }
+
+            void scenario() {
+                Child child = null;
+                child.target();
+            }
+        }
+
+        public static class LeftRightHasPrecedenceOverRight {
+            interface LeftLeftGrandParent {
+            }
+
+            interface LeftRightGrandParent {
+                @ExpectedMethod
+                void target();
+            }
+
+            interface LeftParent extends LeftLeftGrandParent, LeftRightGrandParent {
+            }
+
+            interface RightLeftGrandParent {
+                void target();
+            }
+
+            interface RightRightGrandParent {
+                void target();
+            }
+
+            interface RightParent extends RightLeftGrandParent, RightRightGrandParent {
+            }
+
+            interface Child extends LeftParent, RightParent {
+            }
+
+            void scenario() {
+                Child child = null;
+                child.target();
+            }
+        }
+
+        public static class RightLeftHasPrecedenceOverRightRight {
+            interface LeftLeftGrandParent {
+            }
+
+            interface LeftRightGrandParent {
+            }
+
+            interface LeftParent extends LeftLeftGrandParent, LeftRightGrandParent {
+            }
+
+            interface RightLeftGrandParent {
+                @ExpectedMethod
+                void target();
+            }
+
+            interface RightRightGrandParent {
+                void target();
+            }
+
+            interface RightParent extends RightLeftGrandParent, RightRightGrandParent {
+            }
+
+            interface Child extends LeftParent, RightParent {
+            }
+
+            void scenario() {
+                Child child = null;
+                child.target();
+            }
+        }
+
+        public static class RightRightIsPickedIfThereIsNoAlternative {
+            interface LeftLeftGrandParent {
+            }
+
+            interface LeftRightGrandParent {
+            }
+
+            interface LeftParent extends LeftLeftGrandParent, LeftRightGrandParent {
+            }
+
+            interface RightLeftGrandParent {
+            }
+
+            interface RightRightGrandParent {
+                @ExpectedMethod
+                void target();
+            }
+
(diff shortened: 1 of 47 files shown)


archunit/src/jdk15test/java/com/tngtech/archunit/core/importer/testexamples/methodresolution/DeterminesMethodAnalogouslyToReflectionApi.java
```diff
@@ -0,0 +1,874 @@
...
+                Serializable target();
+            }
+
+            interface RightRightGrandParent {
+                Serializable target();
+            }
+
+            interface RightParent {
+                @ExpectedMethod
+                String target();
```
KorSin, line 584: The way you implemented the resolve process it makes sense that this is the expected method, IF we know the `child.target()` call returns a String, which I presume we know from the byte code. Is it on byte code level the same that the most specific return type is chosen?

archunit/src/main/java/com/tngtech/archunit/core/importer/AccessRecord.java
```diff
@@ -280,215 +274,131 @@ private static JavaCodeUnit getCaller(CodeUnit caller, ImportedClasses classes)
...
+                    : searchTargetMethodInHierarchy(targetOwner, matchingMethods);
+        }
 
-            public ClassHierarchyPath(JavaClassDescriptor childType, JavaClass parent) {
-                Optional<JavaClass> child = tryFindChildInHierarchy(childType, parent);
-                path = child.isPresent() ? createPath(parent, child.get()) : Collections.<JavaClass>emptyList();
+        private static Optional<JavaMethod> searchTargetMethodInHierarchy(JavaClass targetOwner, MatchingMethods matchingMethods) {
+            Optional<JavaClass> superclass = targetOwner.getRawSuperclass();
+            if (superclass.isPresent()) {
+                matchingMethods.addMatching(superclass.get().getMethods(), true);
```
KorSin, line 329: I get why we should not include static methods but why only for interfaces?

archunit/src/jdk15test/java/com/tngtech/archunit/core/importer/testexamples/methodresolution/DeterminesMethodAnalogouslyToReflectionApi.java
```diff
@@ -0,0 +1,874 @@
...
+
+            void scenario() {
+                Child child = null;
+                child.target();
+            }
+        }
+    }
+
+    public static class InterfaceOnParentHasPrecedenceOverInterfaceOnChild {
+
```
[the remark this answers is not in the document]
codecholeric: I can't tell you the original reason the devs of the Reflection API had (would be interesting though). All I can see is that this is how the algorithm does it. One argument could be that an interface higher up the hierarchy is implemented by more classes, so in a way it's the "more visible" method :man_shrugging: (I.e. in this case `Child.class.getMethod(x)` is equal to `Parent.class.getMethod(x)`). In any case we're sure the Reflection API resolves it the same way and it should be an extreme corner case anyway...

archunit/src/jdk15test/java/com/tngtech/archunit/core/importer/testexamples/methodresolution/DeterminesMethodAnalogouslyToReflectionApi.java
codecholeric: In the bytecode of `Child` you won't find any trace of `target()`. Because `Child` is abstract and doesn't override `target()`. However, the compiler will always forbid you to override a method with a less specific return type. So whoever will eventually override / implement this method, the declaration will have the more specific return type. From that point of view it makes sense to pick the method with the most specific return type in the hierarchy (and again it's also what the Reflection API does, so no uncertainty there :man_shrugging: )

archunit/src/main/java/com/tngtech/archunit/core/importer/AccessRecord.java
codecholeric: My guess is because for classes it's forbidden. I can see that the Reflection API only considers this case, but I can also see that it's impossible to define a static and a clashing instance method within the hierarchy of a class (feel free to prove me wrong, then I need to reconsider this :joy:)
I think the reason is that static methods on interfaces always needs to be called on the Interface. So you can't do

```
interface IFace {
  static void method() {}
}
class Clazz implements IFace {
  void call() {
    method(); // doesn't compile, only IFace.method() would...
  }
}
```

But for classes it's allowed to call them that way

```
class Parent {
  static void method() {}
}
class Child extends Parent {
  void call() {
    method(); // this is fine now
  }
  void method() {} // but this causes a compile error because of ambiguity
}
```

Don't ask me why decisions were made that way, but that's how it is :man_shrugging: My guess would be because interface default methods came later, so the behavior for interfaces was the only way to add this in a backwards compatible way :man_shrugging:

archunit/src/jdk15test/java/com/tngtech/archunit/core/importer/testexamples/methodresolution/DeterminesMethodAnalogouslyToReflectionApi.java
KorSin: Yes you are right, `Child` does not know about it. But the call of `scenario()` seems to know that `child.target()` is of return type String (I checked that in the byte code). Does this mean the compiler also resolves the more specific type?

archunit/src/jdk15test/java/com/tngtech/archunit/core/importer/testexamples/methodresolution/DeterminesMethodAnalogouslyToReflectionApi.java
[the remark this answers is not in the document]
KorSin: Alright, I just wanted to check if I understood it correctly.

archunit/src/main/java/com/tngtech/archunit/core/importer/AccessRecord.java
KorSin: Thanks for the example. That helped me to understand it.

archunit/src/jdk15test/java/com/tngtech/archunit/core/importer/testexamples/methodresolution/DeterminesMethodAnalogouslyToReflectionApi.java
codecholeric: Yes, otherwise covariant return type overrides wouldn't work I guess, right? Once it's covariantly overridden in the hierarchy the more specific return type is the return type of the method so it can be assigned/used safely wherever the more specific type is needed (under the hood a lot more uglyness is going on, like bridge-methods :joy:)

Issue #179, opened 2019-05-26
Make containNumberOfElements() available for all Java elements


Bananeweizen:
Currently the `containNumberOfElements` Conjunction is only available for classes via `com.tngtech.archunit.lang.syntax.elements.ClassesShould.containNumberOfElements(DescribedPredicate<Integer>)`. It should be available for every Java domain model type, e.g. fields, methods, constructors etc.

codecholeric:
Could you elaborate a little more what use case you have in mind? Are you missing it from the members API? I.e.

```
members()...should().containNumberOfElements(predicate)
```

Or are you thinking about members of classes when using the classes API? (`classes()...should()..`)

If you urgently need this, you can always write a custom `ArchCondition` for everything! Like

```
classes()...should(haveNumberOfFieldsGreaterThan(42))
```

or

```
members()...should(containNumberOfElements(greaterThan(1)))
```

codecholeric:
I'm closing this due to inactivity. Feel free to reopen if you still have the problem!

crehn:
I currently have the same problem. Due to the problem described in #178, I'd like to do the following:

```java
methods().that().areDeclaredInClassesThat().haveSimpleNameEndingWith("Foo")
    .should()...
    .andShould().containNumberOfElements(greaterThan(0));
```

codecholeric:
Ah, okay, I see :smiley: Okay, I'll reopen this. It shouldn't be hard to add it (since it's already implemented for classes), so I see if anyone wants to pitch in a PR. Otherwise I will implement it myself once I get around to it (which unfortunately might take a little while). 
Anyway, as mentioned, you can extend the API yourself to provide the feature, the easiest way would be to just copy

```
com.tngtech.archunit.lang.conditions.ArchConditions.NumberOfElementsCondition
```

and adjust it to members. Actually I think implementing that for members might be as easy as generifying this class a little and reusing it from `MembersShould`.

codecholeric:
[...]

PR #724, opened 2021-11-21
Draft: Build on Java 8, 11 & 17 using Gradle 7.3

timtebeek:
Hi! Hope it's ok to just open a pull request like this unprompted; I had wanted to look into #179 originally, but had some issues getting JDK 15 to play nice. Then I read in #565 that you had planned to drop Java 7 support in the near future, so I figured couldn't hurt to start a pull request with some of the required changes.

There's a few puzzling test failures that might need a second look:
```
ClassFileImporterAnnotationsTest. parameters_of_meta_annotations_are_transitively_imported[0: JavaClass{name='com.tngtech.archunit.core.importer.ClassFileImporterAnnotationsTest$MetaAnnotatedClass'}]
ClassFileImporterAnnotationsTest. parameters_of_meta_annotations_are_transitively_imported[1: JavaField{com.tngtech.archunit.core.importer.ClassFileImporterAnnotationsTest$ClassWithMetaAnnotatedField.metaAnnotatedField}]
ClassFileImporterAnnotationsTest. parameters_of_meta_annotations_are_transitively_imported[2: JavaMethod{com.tngtech.archunit.core.importer.ClassFileImporterAnnotationsTest$ClassWithMetaAnnotatedMethod.metaAnnotatedMethod()}]
ClassFileImporterAnnotationsTest. parameters_of_meta_annotations_are_transitively_imported[3: JavaConstructor{com.tngtech.archunit.core.importer.ClassFileImporterAnnotationsTest$ClassWithMetaAnnotatedConstructor.<init>()}]
ClassFileImporterAnnotationsTest. parameters_of_meta_annotations_are_transitively_imported[4: JavaParameter{owner='com.tngtech.archunit.core.importer.ClassFileImporterAnnotationsTest$ClassWithMetaAnnotatedConstructorParameter.<init>(java.lang.String)', index='0', type='java.lang.String'}]
ClassFileImporterAnnotationsTest. parameters_of_meta_annotations_are_transitively_imported[5: JavaParameter{owner='com.tngtech.archunit.core.importer.ClassFileImporterAnnotationsTest$ClassWithMetaAnnotatedMethodParameter.method(java.lang.String)', index='0', type='java.lang.String'}]
MethodCallChainTest. run_test_cases[6: CallChainTestCase{start: [FiveStepsInterfaceChildStep1, FiveStepsHierarchyImplementationStep1], numberOfInvocations: [1], expectedResult: [FiveStepsInterfaceChildStep2, FiveStepsHierarchyImplementationStep2]}]
MethodCallChainTest. run_test_cases[7: CallChainTestCase{start: [FiveStepsInterfaceChildStep1, FiveStepsHierarchyImplementationStep1], numberOfInvocations: [2], expectedResult: [FiveStepsInterfaceChildStep3, FiveStepsHierarchyImplementationStep3]}]
MethodCallChainTest. run_test_cases[9: CallChainTestCase{start: [FiveStepsInterfaceChildStep1, FiveStepsHierarchyImplementationStep1], numberOfInvocations: [4], expectedResult: [FiveStepsInterfaceChildStep5, FiveStepsHierarchyImplementationStep5]}]
```
Other than that I think this should cover most of the odd ends. Any feedback/help much appreciated!

Once Java 7 support is officially dropped one could look into replacing the Guava FluentIterable with Java 8 streams for instance, as that could help with code familiarity to other developers.

I doubted whether you'd want to rename the jdk9/jdk15 folders to for instance jdk11 & jdk17, to target stable JDK versions.

I recon I'll start a separate one for that #179, as this bit of yak shaving got out of hand. :)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ on:
   pull_request:
 
 env:
-  build_java_version: 15
+  build_java_version: 17
 
 jobs:
   build:
@@ -41,6 +41,8 @@ jobs:
           - 13
           - 14
           - 15
+          - 16
+          - 17
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
@@ -90,6 +92,8 @@ jobs:
           - 13
           - 14
           - 15
+          - 16
+          - 17
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,7 +33,6 @@ jobs:
           - macos-latest
           - windows-latest
         test_java_version:
-          - 7
           - 8
           - 9
           - 10
@@ -83,7 +82,6 @@ jobs:
           - macos-latest
           - windows-latest
         java_version:
-          - 7
           - 8
           - 9
           - 10
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ _site
 .jekyll-cache
 .asciidoctor
 verification-results
+bin/
--- a/archunit-junit/build.gradle
+++ b/archunit-junit/build.gradle
@@ -96,4 +96,3 @@ ext.configureJUnitArchive = { configureDependencies ->
 this.with addCleanThirdPartyTask
 
 javadoc.enabled = false
-uploadArchives.enabled = false
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/AccessTarget.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/AccessTarget.java
@@ -20,11 +20,11 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import com.google.common.base.Predicate;
 import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
-import com.google.common.collect.FluentIterable;
 import com.google.common.collect.ImmutableList;
 import com.tngtech.archunit.PublicAPI;
 import com.tngtech.archunit.base.ChainableFunction;
@@ -48,7 +48,6 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.Iterables.getOnlyElement;
 import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
 import static com.tngtech.archunit.base.DescribedPredicate.equalTo;
-import static com.tngtech.archunit.base.Guava.toGuava;
 import static com.tngtech.archunit.core.domain.JavaConstructor.CONSTRUCTOR_NAME;
 import static com.tngtech.archunit.core.domain.properties.HasName.Functions.GET_NAME;
 
@@ -344,9 +343,9 @@ public abstract class AccessTarget implements HasName.AndFullName, CanBeAnnotate
         @Override
         @PublicAPI(usage = ACCESS)
         public ThrowsClause<CodeUnitCallTarget> getThrowsClause() {
-            List<ThrowsClause<JavaCodeUnit>> resolvedThrowsClauses = FluentIterable.from(resolve())
-                    .transform(toGuava(JavaCodeUnit.Functions.Get.throwsClause()))
-                    .toList();
+            List<ThrowsClause<JavaCodeUnit>> resolvedThrowsClauses = resolve().stream()
+                    .map(unit -> JavaCodeUnit.Functions.Get.throwsClause().apply(unit))
+                    .collect(Collectors.toList());
 
             if (resolvedThrowsClauses.isEmpty()) {
                 return ThrowsClause.empty(this);
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,3 @@
 org.gradle.parallel=true
 archunit.version=0.23.0-SNAPSHOT
+org.gradle.jvmargs=--add-exports jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED
--- a/.github/workflows/update-gradle-wrapper.yml
+++ b/.github/workflows/update-gradle-wrapper.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Install Gradle JDK
         uses: actions/setup-java@v1
         with:
-          java-version: 14
+          java-version: 17
 
       - name: Update Gradle Wrapper
         uses: gradle-update/update-gradle-wrapper-action@v1.0.15
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,14 +4,14 @@ Contributions are very welcome. The following will provide some helpful guidelin
 
 ## How to build the project
 
-ArchUnit requires at least JDK 15 to build.
+ArchUnit requires at least JDK 17 to build.
 The following is just an example input/output from a Unix command line.
 Windows users should use `gradlew.bat` instead.
 
 ```
 $ cd /path/to/git/clone/of/ArchUnit
 $ ./gradlew showJdkVersion
-Configured JDK: 15
+Configured JDK: 17
 $ ./gradlew build
 ```
 
--- a/archunit/build.gradle
+++ b/archunit/build.gradle
@@ -108,14 +108,6 @@ task jdk15Test(type: Test) {
     testClassesDirs = sourceSets.jdk15test.output.classesDirs
     classpath = sourceSets.jdk15test.runtimeClasspath
     jvmArgs += "--enable-preview"
-
-    // workaround for issue in Gradle 6.7, see https://github.com/gradle/gradle/issues/14714
-    exclude '**/ClassFileImporterRecordsTest\$*RecordToImport.class'
-    exclude '**/GivenClassesThatRecordsTest\$*SomeRecord.class'
-    exclude '**/GivenMembersDeclaredInClassesThatRecordsTest\$*SomeRecord.class'
-    exclude '**/ShouldClassesThatRecordsTest\$*SomeRecord.class'
-    exclude '**/ShouldOnlyByClassesThatRecordsTest\$*RecordAccessingRecord.class'
-    exclude '**/ShouldOnlyByClassesThatRecordsTest\$*SomeRecord.class'
 }
 
 [jar, test]*.dependsOn compileJdk9mainJava
--- a/build-steps/testing/testing.gradle
+++ b/build-steps/testing/testing.gradle
@@ -1,6 +1,5 @@
 def unquote = { string -> string.replaceAll(/^"(.*)"$/, '$1')}
 def testJdksDefinition = [
-        [suffix: "Jre7", javaVersion: JavaVersion.VERSION_1_7, jdkProp: "java7Home"],
         [suffix: "Jre8", javaVersion: JavaVersion.VERSION_1_8, jdkProp: "java8Home"],
         [suffix: "Jre9", javaVersion: JavaVersion.VERSION_1_9, jdkProp: "java9Home"],
         [suffix: "Jre10", javaVersion: JavaVersion.VERSION_1_10, jdkProp: "java10Home"],
--- a/build-steps/testing/testing.gradle
+++ b/build-steps/testing/testing.gradle
@@ -7,7 +7,9 @@ def testJdksDefinition = [
         [suffix: "Jre12", javaVersion: JavaVersion.VERSION_12, jdkProp: "java12Home"],
         [suffix: "Jre13", javaVersion: JavaVersion.VERSION_13, jdkProp: "java13Home"],
         [suffix: "Jre14", javaVersion: JavaVersion.VERSION_14, jdkProp: "java14Home"],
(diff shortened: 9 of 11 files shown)


gradle.properties
```diff
@@ -1,2 +1,3 @@
 org.gradle.parallel=true
 archunit.version=0.23.0-SNAPSHOT
+org.gradle.jvmargs=--add-exports jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED
```
codecholeric, line 3: Why do we need this only now, but not before? :thinking: Is it related to JDK 8 or JDK 17 or upgrading Gradle? In any case an explanation in the respective commit message would be cool :slightly_smiling_face:

archunit/src/main/java/com/tngtech/archunit/core/domain/AccessTarget.java
```diff
@@ -344,9 +343,9 @@ public JavaClass getRawReturnType() {
         @Override
         @PublicAPI(usage = ACCESS)
         public ThrowsClause<CodeUnitCallTarget> getThrowsClause() {
-            List<ThrowsClause<JavaCodeUnit>> resolvedThrowsClauses = FluentIterable.from(resolve())
-                    .transform(toGuava(JavaCodeUnit.Functions.Get.throwsClause()))
-                    .toList();
+            List<ThrowsClause<JavaCodeUnit>> resolvedThrowsClauses = resolve().stream()
```
codecholeric, line 346: It seems a little random to just replace this one usage of `FluentIterable` :thinking: Maybe drop this from the commit and we have some commit/PR later that replaces all `FluentIterable` by Java Streams?

archunit-junit/build.gradle
```diff
@@ -96,4 +96,3 @@ ext.configureJUnitArchive = { configureDependencies ->
 this.with addCleanThirdPartyTask
 
 javadoc.enabled = false
-uploadArchives.enabled = false
```
codecholeric, line 99: Why has this been removed? Are you sure it's not publishing `archunit-junit` as a separate archive now?

.gitignore
```diff
@@ -13,3 +13,4 @@ _site
 .jekyll-cache
 .asciidoctor
 verification-results
+bin/
```
codecholeric, line 16: This does not seem to be related to the commit? I would make it a separate commit then and explain for which case you ignore it (I guess you're using Eclipse? Or what tool creates a `bin` folder?)

.github/workflows/build.yml
```diff
@@ -83,15 +77,9 @@ jobs:
           - macos-latest
           - windows-latest
         java_version:
-          - 7
           - 8
-          - 9
```
codecholeric, line 88: Do we really need to drop all the tests for other JDK versions? I would think dropping 7 is good enough? I know most are end of life, but I still like to make sure ArchUnit works if dropped in such a projekt :man_shrugging:

codecholeric:
Thanks for the initiative :smiley: We will still release one or two times supporting Java 7 and then go to version `1.0.0`, drop Java 7 support and upgrade Gradle. So this will still be on ice for a little while :wink: 
Other than that I added some quick spontaneous feedback from glancing over it as comments. One thing I think would be nice is to have separate commits, i.e. do one thing per commit. One to upgrade to JDK 8, one commit to upgrade Gradle, and if there are other unrelated things then also separate commits. And explain everything that is not immediately obvious within the respective message :slightly_smiling_face: (I added some example with the `jvmargs` where I think I know why you add it, but I'm not sure why it is necessary exactly now :man_shrugging:) It just makes reasoning about these things in half a year so much easier...

archunit/src/main/java/com/tngtech/archunit/core/domain/AccessTarget.java
timtebeek: I agree it seems random, but this one gave me a compilation error locally; When I revert it I once again get:
```
> Task :archunit:compileJava FAILED
/home/tim/Documents/workspace/archunit/archunit/src/main/java/com/tngtech/archunit/core/domain/AccessTarget.java:349: error: incompatible types: ImmutableList<ThrowsClause<CAP#1>> cannot be converted to List<ThrowsClause<JavaCodeUnit>>
                    .toList();
                           ^
  where CAP#1 is a fresh type-variable:
    CAP#1 extends JavaCodeUnit from capture of ? extends JavaCodeUnit
1 error
```

That's running on JDK 17-open as installed through sdkman:
```
openjdk version "17" 2021-09-14
OpenJDK Runtime Environment (build 17+35-2724)
OpenJDK 64-Bit Server VM (build 17+35-2724, mixed mode, sharing)
```

I had tried to change the assignment, but that failed for me as well; figured a first move away from Guava wouldn't hurt on Java 8+.

Do you see any other options to resolve this issue?

.gitignore
timtebeek: Indeed using Eclipse based IDE; was accidental that this was included, sorry about that! Will revert.

archunit-junit/build.gradle
timtebeek: When running on Gradle 7.3 I get the following error message is this line is present:
```
> Configure project :archunit-java-modules-test
Project :archunit-java-modules-test => 'example_module' Java module

FAILURE: Build failed with an exception.

* Where:
Build file '/home/tim/Documents/workspace/archunit/archunit-junit/build.gradle' line: 99

* What went wrong:
A problem occurred evaluating project ':archunit-junit'.
> Could not get unknown property 'uploadArchives' for project ':archunit-junit' of type org.gradle.api.Project.
```
Figured as the flag was disabled, not being able to find the property would lead to the same result.
Although some migration might still be in order if it was replaced by another now enabled by default task.

gradle.properties
timtebeek: These were successively encountered when running on JDK 17; they are needed since JDK 17 further restricts access to internals. For instance Spotless fails with a message such as:
```
Caused by: java.lang.IllegalAccessError: class com.google.googlejavaformat.java.RemoveUnusedImports (in unnamed module @0x3e573cb9) cannot access class com.sun.tools.javac.util.Context (in module jdk.compiler) because module jdk.compiler does not export com.sun.tools.javac.util to unnamed module @0x3e573cb9
        at com.google.googlejavaformat.java.RemoveUnusedImports.removeUnusedImports(RemoveUnusedImports.java:188)
        ... 139 more
```

.github/workflows/build.yml
timtebeek: Of course we wouldn't strictly have to drop support for the other JDK versions; I'll restore the end of life versions just in case.

Practically speaking I think it'd be quite rare for an issue not to show up in 8, 11 & 17, while still posing issues in say 10 or 14. And even if that would be the case, I imagine the recommended change would be to up- or downgrade to a supported JDK. Comparatively, [ASM seems to only test on JDK 11](https://gitlab.ow2.org/asm/asm/-/blob/master/.gitlab-ci.yml#L1), and any issues there are likely to cascade here.

timtebeek:
Thanks for the detailed feedback already! I agree that separate commits would have been better, especially when you want to `git blame` a year from now. I've replied in the threads for now, but could split those up into separate commits (after likely first messing it up a few times in git).

One thing I think could use a bit of attention already is the test failures when running on JDK 17. Any idea what might be causing those to fail? Could be worthwhile to ensure JDK 17 at least also works for now (although I've not had any issues there yet), in addition to Java 7.

archunit/src/main/java/com/tngtech/archunit/core/domain/AccessTarget.java
codecholeric: Ah, okay, that seems odd :joy: But no, then it makes sense to have it also in the JDK 17 commit, because it's directly related :+1: Would just maybe also add it as a detail to the commit message that this one `FluentIterable` had to be replaced, since it doesn't compile anymore with JDK 17. So it is clear when someone looks at the commit later on why only this was picked to be replaced :slightly_smiling_face:

archunit-junit/build.gradle
codecholeric: Yes, we need to look if it was replaced by a different API then. It would be weird if it was disabled by default, so I would be afraid that it publishes this now as an archive...

.gitignore
codecholeric: No problem :slightly_smiling_face: You can also still add a commit where you add it with an explanation for the next Eclipse user, if you want. I mean, it's not harmful :man_shrugging:

.github/workflows/build.yml
codecholeric: Yes, it might be a little paranoid, but you never know :joy: The part about the classpath scanning is a little tricky and has sometimes changed drastically.

codecholeric:
I think the problem with the JDK 17 test is this hardcoded list here `build-steps/testing/testing.gradle:2` :see_no_evil: I don't know if we should make this a little more generic, but it's also tricky to not make it insane as JDKs slowly develop. Anyway, that creates dynamic Gradle tasks and without that the task `testJre17` doesn't exist what makes the build fail...

gradle.properties
codecholeric: Ah, okay :+1: Then probably would also be good to add that explanation to the commit message of the JDK upgrade commit :slightly_smiling_face:

.github/workflows/build.yml
timtebeek: No problem at all; restored all the outdated JDKs and even added in 16 just to be sure.

.gitignore
timtebeek: Pushed now as a separate commit.

archunit-junit/build.gradle
timtebeek: Added a separate commit with link the the docs:
https://docs.gradle.org/current/userguide/upgrading_version_6.html#removal_of_the_uploadarchives_task
It seems the functionality was moved to plugins, so we're unlikely to inadvertently upload these artifacts now. :)

timtebeek: Although I do see the maven-publish plugin used in: `/archunit/build-steps/release/publish.gradle`.
Can't hurt to double check for anyone more familiar with Gradle and artifact publishing.

archunit/src/main/java/com/tngtech/archunit/core/domain/AccessTarget.java
timtebeek: Pushed it now as a separate commit; If on the other hand we decide to first & separately drop Java 7, it would also be possible to convert more/all FluentIterables to Streams before adding JDK 17.

gradle.properties
timtebeek: Pushed separately with comment.

timtebeek:
Updated the MR; there's a few issues to sort out still.

Firstly there's issues running the `task jdk15Test` on Java 17, as it complains about the enable preview on Java 17, while still needed for Java 15. Not sure about the proper way to resolve that.

Secondly there's now test failures on JDK 15:
```
MethodCallChainTest. run_test_cases[6: CallChainTestCase{start: [FiveStepsInterfaceChildStep1, FiveStepsHierarchyImplementationStep1], numberOfInvocations: [1], expectedResult: [FiveStepsInterfaceChildStep2, FiveStepsHierarchyImplementationStep2]}]
MethodCallChainTest. run_test_cases[7: CallChainTestCase{start: [FiveStepsInterfaceChildStep1, FiveStepsHierarchyImplementationStep1], numberOfInvocations: [2], expectedResult: [FiveStepsInterfaceChildStep3, FiveStepsHierarchyImplementationStep3]}]
MethodCallChainTest. run_test_cases[9: CallChainTestCase{start: [FiveStepsInterfaceChildStep1, FiveStepsHierarchyImplementationStep1], numberOfInvocations: [4], expectedResult: [FiveStepsInterfaceChildStep5, FiveStepsHierarchyImplementationStep5]}]
```

Thirdly there's these additional failures on JDK 17:
```
ClassFileImporterAnnotationsTest. parameters_of_meta_annotations_are_transitively_imported[0: JavaClass{name='com.tngtech.archunit.core.importer.ClassFileImporterAnnotationsTest$MetaAnnotatedClass'}]
ClassFileImporterAnnotationsTest. parameters_of_meta_annotations_are_transitively_imported[1: JavaField{com.tngtech.archunit.core.importer.ClassFileImporterAnnotationsTest$ClassWithMetaAnnotatedField.metaAnnotatedField}]
ClassFileImporterAnnotationsTest. parameters_of_meta_annotations_are_transitively_imported[2: JavaMethod{com.tngtech.archunit.core.importer.ClassFileImporterAnnotationsTest$ClassWithMetaAnnotatedMethod.metaAnnotatedMethod()}]
ClassFileImporterAnnotationsTest. parameters_of_meta_annotations_are_transitively_imported[3: JavaConstructor{com.tngtech.archunit.core.importer.ClassFileImporterAnnotationsTest$ClassWithMetaAnnotatedConstructor.<init>()}]
ClassFileImporterAnnotationsTest. parameters_of_meta_annotations_are_transitively_imported[4: JavaParameter{owner='com.tngtech.archunit.core.importer.ClassFileImporterAnnotationsTest$ClassWithMetaAnnotatedConstructorParameter.<init>(java.lang.String)', index='0', type='java.lang.String'}]
ClassFileImporterAnnotationsTest. parameters_of_meta_annotations_are_transitively_imported[5: JavaParameter{owner='com.tngtech.archunit.core.importer.ClassFileImporterAnnotationsTest$ClassWithMetaAnnotatedMethodParameter.method(java.lang.String)', index='0', type='java.lang.String'}]
```

codecholeric:
I think #755 will allow us to already upgrade Gradle, no matter if we drop Java 7 support or not :smiley: I'll look into it once it has been merged...

timtebeek:
Good to see #755 merged, even if it's heavily conflicting with this one.
Don't think there's much to salvage in this MR; probably best to start over. :upside_down_face:

PR #736, opened 2021-12-08
Issue: #735 Throw Exception if no components are found in diagram

### LINKED ISSUE #735: adhereToPlantUmlDiagram swallows unparseable diagrams
Some days ago I defined an ArchUnit rule that uses a plantUML via adhereToPlantUmlDiagram. 
I asked myself whether ArchUnit also accepts the generated PNG, so I changed passing the diagram.puml to passing the diagram.png and everything seems to work fine. 
Later I wondered why a violation was not detected by that ArchUnit rule. I played around and realized that if the file does not contain valid plantUML source it is ignored silently. 
That's bad because if the file does not contain valid content (e.g. you pass files of the wrong type or the file became corrupted) the test should fail, not pass.

pfichtner:
An empty diagram never makes sense to define architecture rules, but accepting it can easily hide a user error where an incompatible file has been passed.

Resolves #735


codecholeric:
Kewl, thanks a lot for the initiative!! :smile: Looks good to me and makes perfect sense. I've replaced the Guava `Files` API by the `java.nio` one, since the Guava one seems to be deprecated :+1:

Issue #477, opened 2020-11-24
Investigate Gradle's "Toolchains for JVM projects" feature


rweisleder:
Gradle 6.7 introduced a new feature called "Toolchains for JVM projects". We should investigate this feature to simplify our build process, especially the multi-jdk-tests for local development.

From what I understand, we can declare the required Java version for specific tasks (compile, test, etc.). Gradle then tries to locate an already installed JDK or installs one by downloading it from AdoptOpenJDK. (Note: AdoptOpenJDK doesn't provide JDK 7 😉)
The build is then independent of the JDK which runs Gradle itself.

For our multi-jdk-tests, it's currently required to install all JDKs manually and declare the javaHome-properties by hand. This could simplify issues regarding the support of new versions of Java.

See also:
- [Release Notes](https://docs.gradle.org/6.7.1/release-notes.html#new-jvm-ecosystem-features)
- [Documentation](https://docs.gradle.org/6.7.1/userguide/toolchains.html)

codecholeric:
I did look into this a little (for ~45 min) because Gradle 6.x doesn't run on JDK 16 so I was looking for ways to decouple the JDK Gradle runs on from the build and then also the test. It looked quite promising (as if it would also radically simplify the setup of the multi jdk test). But unfortunately in the end I got stuck at https://github.com/diffplug/spotless/issues/724. I.e. Spotless can only format code if the JDK that Gradle is running on matches. It might still work out right now, if we use JDK 15 for Gradle and JDK 16 for the build, but it didn't leave a good enough feeling for me to pursue this further at the moment.
For now I'd tend to wait for Gradle 7 and then upgrade to JDK 16.
I.g. it would be great to use toolchains though, maybe I'll revisit this when I have more time (or if things like Spotless just work out of the box). If anybody has a nice solution for this though, let me know!

codecholeric:
Also it would be quite nice to simplify the JDK setup for contributors, because with toolchains everyone could use their default JDK to run Gradle and the correct build and test JDK would just be installed automatically. At least that's how it looked to me, when I tested it locally :man_shrugging:

codecholeric:
For anyone interested I pushed my current hacking here: https://github.com/TNG/ArchUnit/commit/5511aa52af7406b85d0fcce212091714b2e5e189
[...]

PR #755, opened 2021-12-30
Use Toolchains to decouple Gradle build and test JDK

codecholeric:
This will introduce [Gradle Toolchains](https://docs.gradle.org/current/userguide/toolchains.html) into the ArchUnit build. The benefit is that we can decouple the JRE used to run Gradle from the JDK used to compile specific source sets and the JRE with which we run the test tasks. There was a lot of custom build script logic (`multiJdkTest` and similar) to facilitate running tests with different JREs, but one test gap was always that the code was compiled with the newest JDK necessary to support all features (at the moment JDK 15, limited by the need to be compatible with JDK 7).

With this PR the behavior is now the following:

* no extra option configured for the build -> set JVM compatibility to minimum (at the moment Java 7) and use the newest JDK possible to compile (at the moment JDK 15, since otherwise we run into problems to be compatible with Java 7)
* with the option `-PtestJavaVersion=x` the compile will run with JDK `x` wherever possible. If `x` is too low (e.g. source set needs JDK 9 but `x < 9`) it will use the minimum JDK compatible with the source set. Also tests will now be run with a JRE `x` or the minimum compatible JRE and the Maven integration test will be run with JDK `x`.

There were a couple of incompatibilities that only showed when really using and older JDK as opposed to just setting the target compatibility to a lower level and using a newer JDK.

Note that at the moment we can still not run tests compiled with a newer JDK version than Gradle 6 supports, because the Gradle test plugin does some introspection of the class files and crashes when it doesn't understand the class file format (supposedly to identify which classes pass to the JUnit runner). I.e. it then passes on wrong classes as test classes.

However, this will open up the way to upgrade Gradle even if while we still need to be Java 7 compatible.

Resolves: #477


archunit/src/main/java/com/tngtech/archunit/core/domain/AccessTarget.java
```diff
@@ -440,8 +440,9 @@ public JavaClass getRawReturnType() {
         @Override
         @PublicAPI(usage = ACCESS)
         public ThrowsClause<CodeUnitCallTarget> getThrowsClause() {
+            @SuppressWarnings("RedundantTypeArguments") // Some JDK versions need the type argument, or it will not compile
             List<ThrowsClause<JavaCodeUnit>> resolvedThrowsClauses = FluentIterable.from(resolve())
-                    .transform(toGuava(JavaCodeUnit.Functions.Get.throwsClause()))
+                    .transform(toGuava(JavaCodeUnit.Functions.Get.<JavaCodeUnit>throwsClause()))
```
timtebeek, line 445: Ah this [tripped me up as well](https://github.com/TNG/ArchUnit/pull/724/commits/3dd094d200ec257f17380bc3d77cc902eb274c5e); nice to see another way to resolve it!

PR #760, opened 2022-01-03
Prevent infinite recursion of javaClass.getAllRawInterfaces()

### LINKED ISSUE #754: archunit has a stackoverflowError, when scan a class which has been proguard
archunit has a stackoverflowError, when scan a class which has been proguard.  the origin class is "androidx.core.util.Pools"
[log.zip](https://github.com/TNG/ArchUnit/files/7792733/log.zip)

hankem:
Only recurse into getAllRawInterfaces()/getAllSubclasses() when interface/subclass has not been added before.

ProGuard seems to produce classes that appear to be their own interface.
For those, an unconditional recursion in the `Supplier<Set<JavaClass>> allRawInterfaces` led to a `StackOverflowError`.

Resolves #754

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
@@ -130,7 +130,7 @@ public class JavaClass
                     result.addAll(subclass.getAllSubclasses());
                 }
             }
-            return result;
+            return ImmutableSet.copyOf(result);
         }
     });
     private EnclosingDeclaration enclosingDeclaration = EnclosingDeclaration.ABSENT;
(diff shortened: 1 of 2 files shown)


hankem:
Ooops, I had missed https://github.com/TNG/ArchUnit/issues/754#issuecomment-1003529722 .
But we could still drop the first commit and merge the other two (unrelated) ones. 😉

codecholeric:
Closing this in favor of #792. I have cherry-picked the commit about `ImmutableSet`, dropped the one that makes `getAllRawInterfaces()` resilient against weird bytecode (as mentioned in https://github.com/TNG/ArchUnit/issues/754#issuecomment-1003529722) and replaced the commit that tests the deprecated methods by a cleanup of these methods (they are all deprecated for a while, I think the only effort we should put into them is removing them :wink:)

PR #774, opened 2022-01-14
Verify that the rule input is not empty

### LINKED ISSUE #178: Verify input set is not empty in ArchRule
It's easy to write a wrong rule like
`classes().thatAreDeclaredInAPackage("..nonexisting..).should()...`

Therefore it would be nice to have some additional syntax around the should() to verify that the input for the following conditions is not empty. This can not be implemented by means of a custom predicate, since the predicates are never called for an empty input set.

Basically I'm thinking of something like this:
`classes().thatAreDeclaredInAPackage("..nonexisting..).mustExistAndShould()...`

That rule would already fail before any predicate is executed, if the number of classes is empty when starting to evaluate the conditions.

If you feel this has no value for the general archunit users, I'm happy with response where I need to dig around to implement this myself locally. Thanks

oberprah:
The input for the should-part of a rule should normally not be empty, since then the test will always be green because it doesn't test anything.
With this change, it is now possible to verify that the input of a rule is not empty. If the input is empty, an `AssertionError` is thrown.

The verification if the input is empty can also be controlled via the configuration.
By setting the value `archRule.failOnEmptyShould` to `true` or `false`, we can turn on or off the verification.
The default value is `true` which means this is a breaking change.
However, with the configuration, the old behavior can easily be restored.

Resolve #178

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/lang/ArchRule.java
+++ b/archunit/src/main/java/com/tngtech/archunit/lang/ArchRule.java
@@ -23,6 +23,7 @@ import java.util.regex.Pattern;
 
 import com.google.common.base.Predicate;
 import com.google.common.collect.ImmutableSet;
+import com.tngtech.archunit.ArchConfiguration;
 import com.tngtech.archunit.Internal;
 import com.tngtech.archunit.PublicAPI;
 import com.tngtech.archunit.base.Optional;
@@ -36,9 +37,11 @@ import com.tngtech.archunit.lang.syntax.elements.ClassesShould;
 import com.tngtech.archunit.lang.syntax.elements.ClassesThat;
 import com.tngtech.archunit.lang.syntax.elements.GivenClasses;
 
+import static com.google.common.collect.Iterables.isEmpty;
 import static com.google.common.io.Resources.readLines;
 import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
 import static com.tngtech.archunit.base.ClassLoaders.getCurrentClassLoader;
+import static java.lang.Boolean.TRUE;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
@@ -175,6 +178,8 @@ public interface ArchRule extends CanBeEvaluated, CanOverrideDescription<ArchRul
         }
 
         private static class SimpleArchRule<T> implements ArchRule {
+            private static final String FAIL_ON_EMPTY_SHOULD_PROPERTY_NAME = "archRule.failOnEmptyShould";
+
             private final Priority priority;
             private final ClassesTransformer<T> classesTransformer;
             private final ArchCondition<T> condition;
@@ -206,6 +211,8 @@ public interface ArchRule extends CanBeEvaluated, CanOverrideDescription<ArchRul
             @Override
             public EvaluationResult evaluate(JavaClasses classes) {
                 Iterable<T> allObjects = classesTransformer.transform(classes);
+                verifyNoEmptyShouldIfEnabled(allObjects);
+
                 condition.init(allObjects);
                 ConditionEvents events = new ConditionEvents();
                 for (T object : allObjects) {
@@ -213,6 +220,21 @@ public interface ArchRule extends CanBeEvaluated, CanOverrideDescription<ArchRul
                 }
                 condition.finish(events);
                 return new EvaluationResult(this, events, priority);
+            }
+
+            private void verifyNoEmptyShouldIfEnabled(Iterable<T> allObjects) {
+                if (isEmpty(allObjects) && isFailOnEmptyShouldEnabled()) {
+                    throw new AssertionError("Rule failed to check any classes. " +
+                            "This means either that no classes have been passed to the rule at all, " +
+                            "or that no classes passed to the rule matched the `that()` clause. " +
+                            "To allow rules being evaluated without checking any classes you can set the ArchUnit property " +
+                            FAIL_ON_EMPTY_SHOULD_PROPERTY_NAME + " = " + false);
+                }
+            }
+
+            private boolean isFailOnEmptyShouldEnabled() {
+                return ArchConfiguration.get().getPropertyOrDefault(FAIL_ON_EMPTY_SHOULD_PROPERTY_NAME, TRUE.toString())
+                        .equals(TRUE.toString());
             }
 
             @Override
--- a/archunit/src/test/java/com/tngtech/archunit/lang/ArchRuleTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/lang/ArchRuleTest.java
@@ -1,13 +1,23 @@
 package com.tngtech.archunit.lang;
 
+import java.io.File;
+import java.io.IOException;
+import java.util.Iterator;
+import java.util.List;
+import java.util.SortedSet;
+import java.util.TreeSet;
+
 import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
 import com.google.common.io.Files;
+import com.tngtech.archunit.ArchConfiguration;
 import com.tngtech.archunit.core.domain.JavaClass;
 import com.tngtech.archunit.core.domain.JavaClasses;
 import com.tngtech.archunit.core.domain.JavaClassesTest;
+import com.tngtech.archunit.core.importer.testexamples.SomeClass;
 import com.tngtech.archunit.lang.ArchConditionTest.ConditionWithInitAndFinish;
 import com.tngtech.archunit.lang.syntax.ArchRuleDefinition;
+import com.tngtech.archunit.testutil.ArchConfigurationRule;
 import org.hamcrest.Description;
 import org.hamcrest.TypeSafeMatcher;
 import org.junit.After;
@@ -15,13 +25,6 @@ import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
-
-import java.io.File;
-import java.io.IOException;
-import java.util.Iterator;
-import java.util.List;
-import java.util.SortedSet;
-import java.util.TreeSet;
 
 import static com.google.common.collect.Lists.newArrayList;
 import static com.tngtech.archunit.core.domain.TestUtils.importClasses;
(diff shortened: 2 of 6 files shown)


archunit/src/main/java/com/tngtech/archunit/lang/ArchRule.java
```diff
@@ -59,6 +61,10 @@
  * @see com.tngtech.archunit.library.dependencies.Slices.Transformer
  */
 public interface ArchRule extends CanBeEvaluated, CanOverrideDescription<ArchRule> {
+
+    @PublicAPI(usage = ACCESS)
+    String VERIFY_INPUT_NOT_EMPTY_PROPERTY_NAME = "archRule.verifyInputNotEmpty";
```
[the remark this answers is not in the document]
codecholeric: Maybe we can name the property a little different? Because "input" is a little ambiguous for me. Generally rules work the following way:

```
allImportedClasses -> rule.that(..) -> filteredClasses -> rule.should(..) -> checks filteredClasses
```

So "input" could also refer to `allImportedClasses`, but what we're testing is even more strict (it would fail if `allImportedClasses` is empty but also on `filteredClasses` being empty). Maybe `failOnEmptyCheck`? Or `failOnEmptyShould`? :thinking:

archunit/src/test/java/com/tngtech/archunit/lang/ArchRuleTest.java
```diff
@@ -41,6 +42,7 @@
     @Before
     public void setUp() {
         ignoreFile().delete();
+        ArchConfiguration.get().reset();
```
codecholeric, line 45: There is a rule for that, that automatically sanitizes the configuration: `ArchConfigurationRule`
Then you don't need to manually reset the configuration, etc. You can check out the usage in other tests :slightly_smiling_face:

archunit/src/test/java/com/tngtech/archunit/lang/ArchRuleTest.java
```diff
@@ -159,6 +161,20 @@ public void finish(ConditionEvents events) {
         assertThat(condition.eventsFromFinish.getViolating()).hasSize(1);
     }
 
+    @Test
+    public void empty_input_throws_an_assertion_exception_by_default() {
+        thrown.expect(AssertionError.class);
```
codecholeric, line 166: I always have a little bad feeling about testing `AssertionErrors` like that. Because that is also the error used by JUnit itself to mark tests as failed. What about using `try .. catch` here and also asserting the error message? Because at the moment this test would be green for any test failure whatsoever, no?

archunit/src/test/java/com/tngtech/archunit/lang/ArchRuleTest.java
```diff
@@ -159,6 +161,20 @@ public void finish(ConditionEvents events) {
...
+
+        classes().should().haveSimpleName("NoClassesShouldBePassedAsInput").evaluate(importClassesWithContext());
```
codecholeric, line 168: You could create the opposite of `com.tngtech.archunit.lang.ArchRuleTest#ALWAYS_BE_VIOLATED`. I.e. a condition that never causes any error by just being empty. Then it's clear that this rule can never fail in a regular way. Also, I would at least once test that the filtering to 0 classes breaks the rule and not only an empty input set (compare my comment from above about the rule process). It's an important aspect that a non-empty set of classes that is filtered down to 0 classes also causes this failure.

archunit/src/test/java/com/tngtech/archunit/lang/ArchRuleTest.java
```diff
@@ -159,6 +161,20 @@ public void finish(ConditionEvents events) {
         assertThat(condition.eventsFromFinish.getViolating()).hasSize(1);
     }
 
+    @Test
+    public void empty_input_throws_an_assertion_exception_by_default() {
```
codecholeric, line 165: It's an `assertion_**error**` :wink: Also, given the other comments probably "rule_that_fails_to_check_any_class..."? :thinking:

archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/GivenClassesThatTest.java
```diff
@@ -469,6 +471,8 @@ public void areNotMetaAnnotatedWith_predicate() {
 
     @Test
     public void implement_type() {
+        ArchConfiguration.get().setProperty(VERIFY_INPUT_NOT_EMPTY_PROPERTY_NAME, Boolean.FALSE.toString());
```
codecholeric, line 474: What about just setting this to `false` for the whole test class? Optimally just by using `ArchConfigurationRule` so the configuration is automatically reset after each test. I think for these tests it's okay to switch the default behavior for the whole class, that reduces clutter.

codecholeric (changes requested):
Thanks a lot for this :smile: I'm sure this will be super useful for users!
I looked over it and added some comments :slightly_smiling_face: I like the first commit you did separately and explained well! :+1: In generally the commits are described very well :slightly_smiling_face: I would still suggest we squash the 2nd and the 3rd commit. Because the 3rd commit is a direct consequence of the 2nd one, and also we don't want to have any state in the repository that doesn't run successfully (i.e. no matter which commit I check out, I want `gradlew build` to pass successfully).

archunit/src/main/java/com/tngtech/archunit/lang/ArchRule.java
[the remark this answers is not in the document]
oberprah: > I don't think I would make this public API.

Yes, totally agree. This was just a quick fix for a failing `PublicAPIRules`-rule 😅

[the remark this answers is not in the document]
oberprah: I like your suggested renaming. I prefer `failOnEmptyShould` over `failOnEmptyCheck` since, for me, with `check` it is not totally clear what check is meant. I renamed it also accordingly

archunit/src/main/java/com/tngtech/archunit/lang/ArchRule.java
```diff
@@ -215,6 +222,16 @@ public EvaluationResult evaluate(JavaClasses classes) {
                 return new EvaluationResult(this, events, priority);
             }
 
+            private void verifyInputNotEmpty(Iterable<T> allObjects) {
+                if (isVerificationForEmptyInputActive() && Iterables.isEmpty(allObjects)) {
+                    throw new AssertionError("The input for a rule should not be empty, since then the test will always be green and does not really test something");
```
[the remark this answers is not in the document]
oberprah: Renamed it.

What do you mean with _hint_, adding some documentation to the user guide? I think that's a good idea, but I have a few questions. Where exactly should I add it? Probably adding a new section in the "Advanced Configuration" section?

By looking at the code, for me, it seems, that the user guide is kind of duplicated. There exists the `index.html` which includes everything from the user guide, but then there are also the `.adoc` files which contain the same information. What exactly is there the difference?

archunit/src/test/java/com/tngtech/archunit/lang/ArchRuleTest.java
oberprah: Ah, cool. Didn't know that 👌🏽

archunit/src/test/java/com/tngtech/archunit/lang/ArchRuleTest.java
oberprah: Yes, good point. I extended the test, so that also the error message is asserted.

archunit/src/test/java/com/tngtech/archunit/lang/ArchRuleTest.java
oberprah: Created `ALWAYS_BE_VALID` and added a test, where a non-empty set of classes is filtered down to 0.

archunit/src/test/java/com/tngtech/archunit/lang/ArchRuleTest.java
oberprah: I renamed the test methods, but I'm not sure if they are good enough. Maybe you have better suggestions?😄

archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/GivenClassesThatTest.java
oberprah: Yes, for this class I agree that it is a good solution.

However, I'm not sure If I understood you correctly regarding the `ArchConfigurationRule`. The rule we use to clean up the config after each test, right? But to set our property to `false` we need to use a `@Before` or can this be done also by using the rule?

With `@Before` (and `@After` inside the rule) we are changing the configuration on each test. Would it in this case not be better to use a `@BeforeClass`/`@AfterClass` so that it is only changed once for the test?

archunit/src/main/java/com/tngtech/archunit/lang/ArchRule.java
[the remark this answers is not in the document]
codecholeric: By "hint" I meant add it to the exception message :wink: I think it's nice when the exception message already tells you how you can change the behavior in this case.
Nevertheless, it might also be a good idea to add it to the user guide. The HTML is generated and should never be modified manually, the source of truth is the `adoc` (the reason is that GitHub pages doesn't whitelist the Jekyll Asciidoc plugin :disappointed: Otherwise, the HTML could be generated on the fly on page deployment).
So yes, the "Advanced Configuration" section would also be my intuition where I would put it, feel free to add it to the `adoc` file! The HTML you can ignore for now, it will be updated during the release process anyway.

archunit/src/main/java/com/tngtech/archunit/lang/ArchRule.java
```diff
@@ -215,6 +222,16 @@ public EvaluationResult evaluate(JavaClasses classes) {
                 return new EvaluationResult(this, events, priority);
             }
 
+            private void verifyNotEmptyIfActive(Iterable<T> allObjects) {
+                if (isPropertyFailOnEmptyShouldActive() && Iterables.isEmpty(allObjects)) {
+                    throw new AssertionError("Rule failed to check any classes");
```
codecholeric, line 227: As mentioned, I was thinking about sth. like "Rule failed to check any classes. To allow rules being executed without checking any classes adjust the ArchUnit configuration to `archRule.failOnEmptyShould=false`" or sth. like that.

archunit/src/test/java/com/tngtech/archunit/lang/ArchRuleTest.java
```diff
@@ -250,6 +279,14 @@ public void check(JavaClass item, ConditionEvents events) {
                 }
             };
 
+    private static final ArchCondition<JavaClass> ALWAYS_BE_VALID =
+            new ArchCondition<JavaClass>("always be valid") {
+                @Override
+                public void check(JavaClass item, ConditionEvents events) {
+                    events.add(new SimpleConditionEvent(item, true, "I'm valid"));
```
codecholeric, line 286: You don't need to add an event here :wink: The condition will also pass if there is no event at all...

archunit/src/main/java/com/tngtech/archunit/lang/ArchRule.java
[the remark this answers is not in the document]
codecholeric: Regarding `@PublicAPI`: I probably didn't express this very well :upside_down_face: What I meant is don't make this public at all. You can move it into `SimpleArchRule`, which is a class (and also the real user of this property), and make it private. In the test I would rather use the raw string then :man_shrugging:

codecholeric (approved):

PR #778, opened 2022-01-27
implement containNumberOfElements for members 

### LINKED ISSUE #179: Make containNumberOfElements() available for all Java elements

oberprah:
The method `containNumberOfElements` can be used to ensure that the `that`-clause matched a certain number of elements.
E.g. `classes().that().implement(SomeBusinessInterface.class).should().containNumberOfElements(lessThanOrEqualTo(1))`.
Until now, this method only existed for classes, but not for members.
With this commit, we implemented it now also for members.

Closes #179

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/lang/conditions/ArchConditions.java
+++ b/archunit/src/main/java/com/tngtech/archunit/lang/conditions/ArchConditions.java
@@ -1083,8 +1083,8 @@ public final class ArchConditions {
     }
 
     @PublicAPI(usage = ACCESS)
-    public static ArchCondition<JavaClass> containNumberOfElements(DescribedPredicate<? super Integer> predicate) {
-        return new NumberOfElementsCondition(predicate);
+    public static <T extends HasName.AndFullName> ArchCondition<T> containNumberOfElements(DescribedPredicate<? super Integer> predicate) {
+        return new NumberOfElementsCondition<>(predicate);
     }
 
     /**
@@ -1321,9 +1321,9 @@ public final class ArchConditions {
         }
     }
 
-    private static class NumberOfElementsCondition extends ArchCondition<JavaClass> {
+    private static class NumberOfElementsCondition<T extends HasName.AndFullName> extends ArchCondition<T> {
         private final DescribedPredicate<Integer> predicate;
-        private final SortedSet<String> allClassNames = new TreeSet<>();
+        private final SortedSet<String> allElementNames = new TreeSet<>();
 
         NumberOfElementsCondition(DescribedPredicate<? super Integer> predicate) {
             super("contain number of elements " + predicate.getDescription());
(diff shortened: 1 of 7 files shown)


archunit/src/main/java/com/tngtech/archunit/lang/conditions/ArchConditions.java
```diff
@@ -1321,33 +1326,68 @@ public void check(JavaClass javaClass, ConditionEvents events) {
...
     }
 
+    private static class NumberOfClassesCondition extends NumberOfElementsCondition<JavaClass> {
+        NumberOfClassesCondition(DescribedPredicate<? super Integer> predicate) {
+            super(predicate);
+        }
+
+        @Override
+        protected String getUniqueName(JavaClass item) {
+            return item.getName();
```
codecholeric, line 1366: You can also use `JavaClass.getFullName`, it's the same. GIven that, it should be enough to parameterize "classes" vs "members", which should make the subclasses obsolete, don't you think?

archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/MembersShouldTest.java
```diff
@@ -653,6 +660,65 @@ public void haveNameNotEndingWith(ArchRule rule, String suffix) {
...
+                $(greaterThan(999)),
+                $(lessThanOrEqualTo(0)),
+                $(lessThanOrEqualTo(1)),
+                $(greaterThanOrEqualTo(3)),
+                $(greaterThanOrEqualTo(999)));
+    }
+
+    @Test
+    @UseDataProvider("containNumberOfElements_rules_two_members")
+    public void containNumberOfElements_two_methods_with_same_name(DescribedPredicate<Integer> predicate) {
```
codecholeric, line 711: I don't exactly understand why we need both these tests? One test that has a constructor, field and method seems good enough? :thinking: What's the additional test coverage we're getting here? The "full name" is not the same, because the full name includes the parameter names :man_shrugging:

codecholeric (changes requested):
Nice work, looks very good :smiley: :+1: I have only one comment about the two test methods you can look at.
Also the commit / PR message sounds a little confusing to me :thinking: The phrase "that a should-set fulfills a given predicate" sounds very generic, doesn't that hold on pretty much every rule? Maybe something like "can be used to ensure that the `that`-clause matched a certain number of elements"? :thinking:

archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/MembersShouldTest.java
oberprah: I added this test, because I was not sure how the method `getFullName` works, and thus I wanted to test, that for two methods with the same name, but different parameters, the "full name" is different. But I agree, that we can probably remove this test, since we do not want to test in this test that the method `getFullName` works as expected.

codecholeric (approved):
Sorry that I only realized it now, but I saw that the tests are really a little overkill :zany_face: I know that you looked at the test for classes and thought "that @codecholeric knew what he was doing", but I'm afraid those tests were a little over the top back then :joy: Because, it makes no sense to test 10 different predicates, when each predicate is already tested in isolation and the rule only relies on generic predicate semantics. On the other hand, the positive case was missing (i.e. that it passes if the predicate matches). I've adjusted the test for the classes in a pre-commit and then fixed up the members test to match the classes test...
Thanks a lot for your contribution!! Really appreciated :smile:

PR #779, opened 2022-01-27
enable running build with JDK 17

timtebeek:
Figured give this another try now that [toolchains where added in #755](https://github.com/TNG/ArchUnit/pull/755).
Continues on from a [previous attempt](https://github.com/TNG/ArchUnit/pull/724), but now with Java 7 retained.
Also bumps Gradle wrapper to 7.3.3, after bumps were disabled in #565.

Was not yet sure if I should [bump maxSupportedJavaVersion to 17](https://github.com/timtebeek/ArchUnit/blob/java-17/build.gradle#L82);
When I tried that locally I got:
```
> Task :archunit:compileJdk15testJava FAILED
error: invalid source release 15 with --enable-preview
  (preview language features are only supported for release 17)
```
Maybe some conditionals are in order to set that flag conditionally.

Any feedback appreciated!


codecholeric:
Thanks a lot! I also think this could work now :slightly_smiling_face: About the `--enable-preview`, I actually think we should change the source set `testJdk15` to `testJdk16` and only start testing records from JDK 16 on. Then we don't need the compiler flag anymore, and at this point I don't think we need to continuously test that JDK 15 preview records are resolved correctly.

timtebeek:
> About the `--enable-preview`, I actually think we should change the source set `testJdk15` to `testJdk16` and only start testing records from JDK 16 on.

As perfect solution indeed; it all works out of the box on JDK 17 locally now. :) Ready for review! :)

codecholeric:
I looked over it, looks very good :smile: I took the liberty to slightly restructure the commits, because I think in the future it will be nice to see as one unit "what was necessary for Gradle 7 upgrade", "what was necessary for JDK 16", ... with additional details about the separate aspects in the commit message. Other than that this will finally bring ArchUnit back to the present and remove that horrible dependency on JDK 15 :partying_face: 
Thanks for all your effort! I know it's tough constraints with JDK 7 support, etc., which makes staying current quite tedious at times.

Now I just need to figure out why GH Actions die with a weird `Permission Denied` error since upgrading to Gradle 7.4 :grimacing:

timtebeek:
Great, thanks for the help! I'd intentionally left out Gradle 7.4 as that also carries some possible improvements with consolidating test reports, that I thought would be best to tackle separately. But if you can make it work then great!

codecholeric:
Hmm, now I'm puzzled :confused: Even if I downgrade to Gradle 7.3.3 again the build fails so weirdly now. Maybe something in GH Actions is broken :thinking: But strangely an older PR with Toolchains runs successfully...
Quite strange, since it always seems to be some "permission problem" like `can't create .ok file` or similar :grimacing: 
But I checked the diff before I force-pushed the new commit structure and the only real difference was the Gradle version :thinking:

codecholeric:
I guess somehow the upgrade from Gradle `7.3.3` to `7.4` indeed breaks the CI on GitHub Actions :exploding_head: I guess it must be some part of the changes of the Toolchains: https://docs.gradle.org/7.4/release-notes.html
Just very obscure, it's always some permission problems to write some file, which AFAIS has not really changed :thinking: 
Anyway, I'm gonna revert that upgrade to 7.3.3 then. Because the upgrade to Gradle 7 and the possibility to run the build with JDK 17 is way more valuable for the moment than to use the latest minor version. We'll tackle the other problem at a later time :+1:

codecholeric:
Okay, at least I noticed that the same thing actually happens locally if I try to provision a new JDK 9 with Gradle `7.4` and Toolchains. So at least it's reproducible locally and has nothing to do with GitHub Actions :man_shrugging:

PR #781, opened 2022-02-02
Load classes from TCCL

stuartwdouglas:
This allows ArchUnit to work with Quarkus continuous testing, which does not have a flat class path.

Quarkus continuous testing runs tests repeatedly in the same VM when class files are changed, which means that the tests need to be loaded in a new ClassLoader every time (which is set to the Thread context ClassLoader). Without this patch ArchUnit cannot load the classes. 

Signed-off-by: Stuart Douglas <stuart.w.douglas@gmail.com>


famod:
Hi @codecholeric! This is a bit of a blocker for Quarkus so we'd really appreciate if you'd take a look. Thanks a lot!

codecholeric:
Thanks a lot for the input! :smiley: Sorry for being a little slow, I'm on holiday in an area with very limited internet :grimacing: Can we extend the PR description a little bit? Just for documentation purposes, what was the problem (how did it manifest, e.g. exception stacktrace) and how does the change solve it?
I think I understand it (Quarkus loads classes with a different `ClassLoader` than the `AppClassLoader` and that one is set as context `ClassLoader`?) but for other users stumbling over this I still think it would be nice :slightly_smiling_face: 

I guess this is hard to test, huh? :thinking: Because it would be somewhat nice to ensure this doesn't break in the future. Do you think it makes sense to add a little Quarkus test module (similar to https://github.com/TNG/ArchUnit/tree/main/archunit-java-modules-test) to ensure this in the long run? (and also avoid regression in any other area?)

In any case I think it would make sense to encapsulate this somehow. There is already `com.tngtech.archunit.base.ClassLoaders` that serves a similar purpose. Would it make sense to add a `loadClass` method there, that does it correctly? (i.e. first try the context `ClassLoader`, then fall back?)

If I understand this correctly, then there is no problem if we load ArchUnit's own classes with the default `ClassLoader` (e.g. https://github.com/TNG/ArchUnit/blob/main/archunit/src/main/java/com/tngtech/archunit/core/PluginLoader.java#L83), but only if we reach out to Quarkus application/test classes?

stuartwdouglas:
I tried to do what you suggest with ClassLoaders and it caused some architecture based tests to fail (first it was accessing types via reflection, I added the annotation and then it complained about layers not being respected).

stuartwdouglas:
Testing is fairly complex, you would need to setup a complete test runner and testing environment I think to run JUnit from inside the test. At the moment Quarkus does not work (it needs the linked PR as well as this one), so testing with Quarkus is not an option yet.

codecholeric:
> I tried to do what you suggest with ClassLoaders and it caused some architecture based tests to fail (first it was accessing types via reflection, I added the annotation and then it complained about layers not being respected).

I see, the problem is that `MayResolveTypesViaReflection` resides in the core layer. However, I think this is only historically, there is no reason for such a central annotation to not reside in the `base` layer. I've refactored it real quick, now it's legal :slightly_smiling_face:

codecholeric:
I understand the testing problem. I tried to add some simplified tests for the problem, at least now for each of your changes, if I roll it back, some test fails. Can you check my changes if that makes sense? (and if it somewhat comes close to the real Quarkus problem?) I tried to model the tests according to some simplified setup mirroring the Quarkus problem (I hope).
If everything looks good to you I would squash the fixups and merge :slightly_smiling_face:

famod:
@codecholeric is there a snapshot somewhere that I can use to test? Last time I tried to get something installed in my local maven repo from a grade based project was a very unpleasant experience...

stuartwdouglas (commented):
Looks good to me. I think that class loader arrangement should be very similar to what we are doing in Quarkus.

codecholeric:
@famod only on call at the moment, which I just did :wink: You can use

```
<repositories>
    <repository>
        <id>sonatype-snapshots</id>
        <url>https://oss.sonatype.org/content/repositories/snapshots</url>
    </repository>
</repositories>
``` 

And pull in version `0.23.0-quarkus-SNAPSHOT`

Can you give me a thumbs up then when you tested it? Then I'll merge...

famod:
Things definitely improved, thanks! Still seeing two issues though: https://github.com/quarkusio/quarkus/pull/23358#issuecomment-1040881430
The second one looks like a Quarkus issue but I'm not sure about the first one. There might still be some classloading issue.
Is there an easy way to see which classes an AU test has gathered?

codecholeric:
If your using the JUnit 5 support the simplest way is probably to add an extra method and dump all imported classes somewhere:

```
@ArchTest
static void printAllClasses(JavaClasses classes) {
    classes.stream().map(JavaClass::getName).forEach(System.out::println);
}
```

Classpath scanning is always a little tricky :grimacing: So if Quarkus does things fundamentally different than the standard JRE, there might be problems lurking somewhere. I can outline real quick what ArchUnit does:

* IF Java Version < 9
  * take all `URLs` from `URLClassLoaders` in the hierarchy of the context `ClassLoader` or the current `ClassLoader`
  * add all URLs from the `java.class.path` system property and the `sun.boot.class.path` system property
* ELSE
  * take all URLs from the `java.class.path` system property and the `sun.boot.class.path` system property
  * add all module URLs from `ModuleFinder.ofSystem()`
  * add all module URLs from modules defined via `jdk.module.path`

This has in parts "historically grown" when bugs occurred, but now for a while it has been quiet :wink: One reason is that I never found a comprehensive guide anywhere with "this is the way to find all sources to classes the current `ClassLoader` has access to". Relying on `URLClassLoader` in earlier versions was already shady (but worked pretty well) and `ClassLoader` per se unfortunately doesn't offer much help (I can understand why, because in the end the interface doesn't demand classes to have an accessible source, so offering an API to retrieve class file sources makes i.g. no sense, but still for most cases there simply is a byte blob source somewhere :man_shrugging:)

famod:
@codecholeric thanks, really appreciated. Turns out to be a test result reporting issue. I will add another comment to the Quarkus PR.

famod:
@codecholeric I added a fix for the reporting issue to the Quarkus PR. I think this PR here can now safely be merged. Thanks!

Do you have an ETA for the release of 0.23?

codecholeric:
@famod I want to merge #785, then I would release. Likely this weekend.

PR #794, opened 2022-02-15
Add predefined ImportOption to exclude package-info.class files

### LINKED ISSUE #793: Maven's compiler plugin 3.10.0 breaks testing
Hello

Thank you for this awesome library!
❤️ 

This is not an actual issue, but I see that _Discussions_ is not activated for this repository.

I wanted to share some info, recently, [maven-compiler-plugin 3.10.0](https://github.com/apache/maven-compiler-plugin/releases/tag/maven-compiler-plugin-3.10.0) was released.
It brings forth some new features and bug fixes and accidentally broke my existing tests with the unexpected appearance of _pacakge-info.class_ files.

Here's the [failing workflow](https://github.com/TomerFi/alexa-skill-shabbat-times/runs/5196280297?check_suite_focus=true#step:4:439).
This, of course, works well with version _3.9.x_ of the _maven-compiler-plugin_.

As a quick fix, I created a simple [custom ImportOption implementation](https://github.com/TomerFi/alexa-skill-shabbat-times/commit/51ce485a7504f6ea9e23f2698befedc0b7f022ed) and used it within the _AnalyzeClasses_ annotation.

I think it will be useful for other users too,
I'll soon open a PR here adding this _ImportOption_ implementation.

See you there.
😉 👋🏻

TomerFi:
Resolves #793

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
@@ -76,6 +76,17 @@ public interface ImportOption {
             public boolean includes(Location location) {
                 return doNotIncludeArchives.includes(location);
             }
+        },
+        /**
+         * @see DoNotIncludePackageInfos
+         */
+        DO_NOT_INCLUDE_PACKAGE_INFOS {
+            private final DoNotIncludePackageInfos doNotIncludePackageInfos = new DoNotIncludePackageInfos();
+
+            @Override
+            public boolean includes(Location location) {
+                return doNotIncludePackageInfos.includes(location);
+            }
         };
 
         static final PatternPredicate MAVEN_TEST_PATTERN = new PatternPredicate(".*/target/test-classes/.*");
@@ -135,6 +146,18 @@ public interface ImportOption {
         @Override
         public boolean includes(Location location) {
             return !location.isArchive();
+        }
+    }
+
+    /**
+     * Excludes {@code package-info.class} files.
+     */
+    final class DoNotIncludePackageInfos implements ImportOption {
+        private static final Pattern PACKAGE_INFO_PATTERN = Pattern.compile(".*package-info\\.class$");
+
+        @Override
+        public boolean includes(Location location) {
+            return !location.matches(PACKAGE_INFO_PATTERN);
         }
     }
 }
--- a/archunit/src/test/java/com/tngtech/archunit/core/importer/ImportOptionsTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/core/importer/ImportOptionsTest.java
@@ -2,6 +2,7 @@ package com.tngtech.archunit.core.importer;
 
 import java.io.File;
 import java.io.IOException;
+import java.net.URISyntaxException;
 import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.List;
@@ -9,6 +10,7 @@ import java.util.List;
 import com.google.common.collect.ImmutableList;
 import com.tngtech.archunit.core.importer.ImportOption.DoNotIncludeArchives;
 import com.tngtech.archunit.core.importer.ImportOption.DoNotIncludeJars;
+import com.tngtech.archunit.core.importer.ImportOption.DoNotIncludePackageInfos;
 import com.tngtech.archunit.core.importer.ImportOption.DoNotIncludeTests;
 import com.tngtech.archunit.core.importer.ImportOption.OnlyIncludeTests;
 import com.tngtech.java.junit.dataprovider.DataProvider;
@@ -24,6 +26,7 @@ import static com.google.common.collect.Iterables.concat;
 import static com.google.common.collect.Iterables.getLast;
 import static com.tngtech.archunit.core.importer.ImportOption.Predefined.DO_NOT_INCLUDE_ARCHIVES;
 import static com.tngtech.archunit.core.importer.ImportOption.Predefined.DO_NOT_INCLUDE_JARS;
+import static com.tngtech.archunit.core.importer.ImportOption.Predefined.DO_NOT_INCLUDE_PACKAGE_INFOS;
 import static com.tngtech.archunit.core.importer.ImportOption.Predefined.DO_NOT_INCLUDE_TESTS;
 import static com.tngtech.archunit.core.importer.ImportOption.Predefined.ONLY_INCLUDE_TESTS;
 import static com.tngtech.java.junit.dataprovider.DataProviders.$;
@@ -155,6 +158,29 @@ public class ImportOptionsTest {
         assertThat(doNotIncludeArchives.includes(locationOf(Object.class)))
                 .as("includes Jrt location")
                 .isFalse();
+    }
+
+    @DataProvider
+    public static Object[][] do_not_include_package_info_classes() {
+        return testForEach(new DoNotIncludePackageInfos(), DO_NOT_INCLUDE_PACKAGE_INFOS);
+    }
+
+    @Test
+    @UseDataProvider("do_not_include_package_info_classes")
+    public void detect_package_info_class(ImportOption doNotIncludePackageInfoClasses) throws URISyntaxException {
+        Location packageInfoLocation = Location.of(getClass().getResource(testExampleResourcePath("package-info.class")).toURI());
+        assertThat(doNotIncludePackageInfoClasses.includes(packageInfoLocation))
+                .as("doNotIncludePackageInfoClasses includes package-info.class")
+                .isFalse();
+
+        Location thisClassLocation = locationOf(getClass());
+        assertThat(doNotIncludePackageInfoClasses.includes(thisClassLocation))
+                .as("doNotIncludePackageInfoClasses includes test class location")
+                .isTrue();
+    }
+
+    private String testExampleResourcePath(String resourceName) {
+        return "/" + getClass().getPackage().getName().replace(".", "/") + "/testexamples/" + resourceName;
     }
 
     private static Location locationOf(Class<?> clazz) {
--- a/archunit/src/test/java/com/tngtech/archunit/core/importer/testexamples/package-info.java
+++ b/archunit/src/test/java/com/tngtech/archunit/core/importer/testexamples/package-info.java
@@ -1,0 +1,5 @@
+// The annotation forces the creation of a compiled `package-info.class`
+@SomeAnnotation(mandatory = "any", mandatoryEnum = SOME_VALUE)
+package com.tngtech.archunit.core.importer.testexamples;
+
+import static com.tngtech.archunit.core.importer.testexamples.SomeEnum.SOME_VALUE;


archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
```diff
@@ -137,4 +149,16 @@ public boolean includes(Location location) {
             return !location.isArchive();
         }
     }
+
+    /**
+     * The {@code maven-compiler-plugin 3.10.0} fixed a bug that made {@code package-info.class}
+     * files hidden from {@code ArchUnit}.
+     * This is a {@link ImportOption} impelmentation for excluding {@code package-info.class} files.
```
hankem, line 156: There's a typo in `impelmentation`.

archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
```diff
@@ -137,4 +149,16 @@ public boolean includes(Location location) {
             return !location.isArchive();
         }
     }
+
+    /**
+     * The {@code maven-compiler-plugin 3.10.0} fixed a bug that made {@code package-info.class}
```
codecholeric, line 154: I don't think we need this context information inside of the Javadoc? This sounds to me like the reason you stumbled over it, but `package-info.class` has been there long before, I have already used it with ArchUnit, Gradle and IntelliJ long ago :man_shrugging: This is for me context information that could go into the commit message or PR message, but I don't think it's relevant for users who want to read the Javadoc of this `ImportOption`. In the Javadoc I would rather expect to see an explanation what `package-info` actually is (maybe with a link to the JLS) and that this `ImportOption` filters it out :thinking:

archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
```diff
@@ -137,4 +149,16 @@ public boolean includes(Location location) {
...
+     * files hidden from {@code ArchUnit}.
+     * This is a {@link ImportOption} implementation for excluding {@code package-info.class} files.
+     */
+    final class DoNotIncludePackageInfoClasses implements ImportOption {
+        @Override
+        public boolean includes(Location location) {
+            return !location.matches(Pattern.compile(".*package-info\\.class$"));
```
codecholeric, line 161: We should not call `Pattern.compile(..)` on every `includes(..)` invocation. That is quite inefficient. We should compile the `Pattern` one time into a static constant and then reuse it on every `includes(..)` invocation (which might in the end be called tens or hundreds of thousands of times for big projects).

I was also wondering if we should add a simple `Location.endsWith(..)` public API to save some performance :thinking: But I have realized that we're (ab-)using patterns also for some other simple cases like `MAVEN_TEST_PATTERN`. Don't know if there is an internal optimization that makes this as efficient as `contains`, maybe we should do some measurements at some point (unrelated to this PR :wink:)

archunit/src/test/java/com/tngtech/archunit/core/importer/ImportOptionsTest.java
```diff
@@ -157,6 +161,30 @@ public void detects_archives_correctly(ImportOption doNotIncludeArchives) {
...
 
+    @DataProvider
+    public static Object[][] do_not_include_package_info_classes() {
+        return testForEach(new DoNotIncludePackageInfoClasses(), DO_NOT_INCLUDE_PACKAGE_INFO_CLASSES);
+    }
+
+    @Test
+    @UseDataProvider("do_not_include_package_info_classes")
+    public void detect_package_info_class(ImportOption doNotIncludePackageInfoClasses) {
+        NormalizedUri mockUri = mock(NormalizedUri.class);
```
codecholeric, line 172: I would not mock things like this. I think mocking is great for collaborators, but not great for domain objects (you lose intrinsic invariants, i.e. you can create unrealistic objects, etc., and the tests are harder to maintain on refactorings).
Is there any reason you don't simply use something like

```
Location location = Location.of(URI.create("file:///path/to/package/and/file/package-info.class"));
```

Then the whole extra class `StubLocation` is also obsolete... :man_shrugging:

archunit/src/test/java/com/tngtech/archunit/core/importer/ImportOptionsTest.java
```diff
@@ -157,6 +161,30 @@ public void detects_archives_correctly(ImportOption doNotIncludeArchives) {
...
+        StubLocation location = new StubLocation(mockUri);
+
+        given(mockUri.toString())
+                .willReturn("/path/to/package/and/file/package-info.class");
+        assertThat(doNotIncludePackageInfoClasses.includes(location))
+                .as("with path for package-info.class file")
```
codecholeric, line 178: I think it's better if the description really refers to the argument of `assertThat(..)`. Because this way you would get an error like `with path for package-info.class file > expected false but was true`. But what should that mean? It would likely be better to have sth. like `doNotIncludePackageInfoClasses.includes(locationOfPackageInfo)`, then seeing `expected false but was true` makes sense, no? :thinking:

archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
```diff
@@ -137,4 +149,16 @@ public boolean includes(Location location) {
             return !location.isArchive();
         }
     }
+
+    /**
+     * The {@code maven-compiler-plugin 3.10.0} fixed a bug that made {@code package-info.class}
+     * files hidden from {@code ArchUnit}.
+     * This is a {@link ImportOption} implementation for excluding {@code package-info.class} files.
+     */
+    final class DoNotIncludePackageInfoClasses implements ImportOption {
```
codecholeric, line 158: Do you think we really need the suffix `Classes`? :thinking: Or would `DoNotIncludePackageInfos` be good enough? Just wondering what is consistent, because it's also called `DoNotIncludeTests` and not `DoNotIncludeTestClasses` :thinking:

archunit/src/test/java/com/tngtech/archunit/core/importer/ImportOptionsTest.java
```diff
@@ -186,4 +214,31 @@ FolderPattern expectTestFolder() {
             return new FolderPattern(folders, false);
         }
     }
+
+    private static class StubLocation extends Location {
```
codecholeric, line 218: As mentioned I think we can do without this :slightly_smiling_face:

archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
```diff
@@ -76,6 +76,18 @@ public boolean includes(Location location) {
             public boolean includes(Location location) {
                 return doNotIncludeArchives.includes(location);
             }
+        },
+
+        /**
+         * Do not include package-info.class files
```
codecholeric, line 82: Reading the Javadocs of these classes now I think the best thing would be to write comprehensive Javadoc on the class and then simply `@see ImportOptionClass` here :thinking: Because `Predefined` and the classes themselves (that you would use with JUnit testing support) should give the same information. Unfortunately the given Javadocs are not exactly great examples :joy: We should clean that up at some point and always refer from `Predefined` to the class that then has the full Javadoc...

codecholeric (changes requested):
Thanks a lot for your contribution :smile: As always I have way to many review comments :see_no_evil: I hope you're still motivated to look into them and let me know what you think!

TomerFi:
> Thanks a lot for your contribution smile As always I have way to many review comments see_no_evil I hope you're still motivated to look into them and let me know what you think!

Not a problem - a good code review is always a good thing.
I'll make the required adjustments by the weekend, most likely tomorrow.

archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
TomerFi: good point.

TomerFi: good point.
cleaned it up in https://github.com/TNG/ArchUnit/pull/794/commits/90c126f6acba68e0602e316c5d576001ee5f20ce

archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
TomerFi: yeah, I guess I missed this, good catch.
fixed in https://github.com/TNG/ArchUnit/pull/794/commits/90c126f6acba68e0602e316c5d576001ee5f20ce

archunit/src/test/java/com/tngtech/archunit/core/importer/ImportOptionsTest.java
TomerFi: that's funny - at the start, I actually did something similar to this and I changed it to the stubbing class instead of including an extra file.
but you are right - it is better like that.
fixed in https://github.com/TNG/ArchUnit/pull/794/commits/90c126f6acba68e0602e316c5d576001ee5f20ce

archunit/src/test/java/com/tngtech/archunit/core/importer/ImportOptionsTest.java
TomerFi: yes, makes perfect sense.
fixed in https://github.com/TNG/ArchUnit/pull/794/commits/90c126f6acba68e0602e316c5d576001ee5f20ce

archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
TomerFi: good point.
fixed in https://github.com/TNG/ArchUnit/pull/794/commits/90c126f6acba68e0602e316c5d576001ee5f20ce

archunit/src/test/java/com/tngtech/archunit/core/importer/ImportOptionsTest.java
TomerFi: this class is no longer with us.
fixed in https://github.com/TNG/ArchUnit/pull/794/commits/90c126f6acba68e0602e316c5d576001ee5f20ce

codecholeric (approved):
Cool, thanks a lot :smile: I rebased on `main` and fixed it up a tiny bit to be consistent with existing style (e.g. we don't use prefixes for the commit message and it shouldn't exceed 70 chars). Hope that's okay with you! (just wanna make sure it ends up in the next release :slightly_smiling_face:)

TomerFi:
> Cool, thanks a lot smile I rebased on `main` and fixed it up a tiny bit to be consistent with existing style (e.g. we don't use prefixes for the commit message and it shouldn't exceed 70 chars). Hope that's okay with you! (just wanna make sure it ends up in the next release slightly_smiling_face)

@codecholeric  Sure, it's your repo - thanks a lot for sharing it and accepting my small contribution.
😻 

I'm just curious...

Because for my projects, I usually do the exact opposite.
I'm a big fan of [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/),
I use the prefix (also known as a type) to categorize my commits, it helps me automate version bumps.
I also used to automate changelog in the same manner,
but today GH releases do an awesome changelog based on PRs.

May I ask why the 70 chars limit?
Does it work well with dependabo'st PRs?

PR #795, opened 2022-02-15
we ensure that the `stored.rules`-file is sorted alphabetically

oberprah:
In the file `stored.rules` there are all frozen rules stored, along with an uuid to a file where than all frozen violations for that rule are stored.
Currently, the content of that file is not sorted in any way.
Therefore, adding a new frozen rule adds a new line to that file and can lead to a reorder of all lines.
The reorder of the file is especially than a problem, when we want to see the changed to the previous version as a diff.
Ideally by adding one new rule also only one line should change in the file.
By sorting the file alphabetically this is given.

Used the following post to sort the file: https://stackoverflow.com/a/55957344

Signed-off-by: Hannes Oberprantacher <h.oberprantacher@gmail.com>


codecholeric:
As discussed, while I see the reasoning behind sorting these files I'm not super happy with extending `java.util.Properties`, because I don't like the maintenance aspect of this (this could break with any JDK upgrade and I don't feel comfortable with extending such core JDK classes, feels "hacky" to me). Did you make any progress considering an extension point for users instead? Where users could hook in code like this themselves?

oberprah:
Yes, I totally agree this approaches feels very hacky. 

However, I haven't tried yet to create an extension point for users. I'll see if I can get around to it next week, otherwise I'll close the PR.

PR #800, opened 2022-02-17
Add `javaClass` to `FieldSource`

famod:
Relates to #781 and would make the integration into Quarkus continuous testing a bit easier, see https://github.com/quarkusio/quarkus/pull/23358#discussion_r809498247

/cc @codecholeric

PS: Naming is inspired by `org.junit.platform.engine.support.descriptor.MethodSource.javaMethod`.

### DIFF
--- a/archunit-junit/junit5/engine-api/src/main/java/com/tngtech/archunit/junit/FieldSource.java
+++ b/archunit-junit/junit5/engine-api/src/main/java/com/tngtech/archunit/junit/FieldSource.java
@@ -25,17 +25,22 @@ import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
 
 @PublicAPI(usage = ACCESS)
 public final class FieldSource implements TestSource {
-    private final String className;
+    private final Class<?> javaClass;
     private final String fieldName;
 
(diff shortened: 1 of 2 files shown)


codecholeric (approved):
Thanks for the PR :smiley: I fixed it up a little bit because I don't think having both `className` and `javaClass` makes too much sense (retrieving the class name from the `Class<?>` is not an expensive operation for which we would need to cache :man_shrugging:). Figured you want to have it merged ASAP, best before the release :wink: 

I have to admit that I have mixed feelings about the naming `getJavaClass()`, because while it's consistent with `MethodSource` it's pretty inconsistent with the rest of ArchUnit :thinking: (within ArchUnit a method named this way would always refer to the domain object `JavaClass`...). But in the end I came to the conclusion that the consistency with `MethodSource` probably still has more weight here, since this is a pure API for devs to integrate it into the JUnit Platform :+1:

famod:
Thanks @codecholeric! I was expecting this to be a bit controversial but I came to the same conclusions as you.

> Figured you want to have it merged ASAP, best before the release

That would be awesome!

PR #801, opened 2022-02-18
Deprecate unwanted `ConditionEvents` API

codecholeric:
At the moment the implementation of `ConditionEvents` is fairly suboptimal. But at the same time changes are hard, because the public API gives too much access. In particular, that `ConditionEvents` allow to publicly retrieve all allowed events wastes a lot of memory. Because usually there are a lot more allowed events than violating events, but the allowed events should be irrelevant for the vast majority of (all?) users.

To improve this situation we change `ConditionEvents` to an interface which should make it more flexible to achieve everything that was achievable so far but to not lock us into a concrete implementation. But we still need to eliminate all public API that forces to keep allowed events. We thus deprecate these methods for now so we can remove them later on.

### DIFF
--- a/archunit/build.gradle
+++ b/archunit/build.gradle
@@ -88,7 +88,8 @@ dependencies {
     runtimeOnly sourceSets.jdk9main.output
 }
 
-compileJdk9mainJava.with {
+compileJdk9mainJava {
+    dependsOn(compileJava)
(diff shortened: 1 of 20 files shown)


codecholeric:
Gonna close this PR for now because I'm not sure yet how the future API will look like. We can break some minor stuff with 1.0 so I'll try to do the refactoring I have in mind then.

Issue #704, opened 2021-10-27
Feature request: Add a beCalledInClassesThat conjunction


cpoissonnier:
I wanted to implement a rule like this one : 

```
@ArchTest
    static final ArchRule events_must_be_created_by_aggregates =
            noConstructors().that().areDeclaredInClassesThat().areAssignableTo(Event.class).should().beCalledInClassesThat().areNotAssignableFrom(Aggregate.class)
                    .because("the aggregate should manage its own lifecycle and events");
```

But it seems that it is now possible without custom predicate.
Is that something that could be added in the ready-to-use conjunctions ?

For reference, the related Stackoverflow question : https://stackoverflow.com/questions/69723542/how-to-check-that-a-constructor-is-called-in-the-right-classes-with-archunit/69728892#69728892

JKLedzion:
Hi, I’ve started working on this issue (hope it’s Ok and it’s still open) and I’ve added a PR with initial part which I’ve done. Work is in progress but it would be great to receive some feedback if it goes in right direction.

I’ve started from adding method onlyBeCalledByClassesThat(DescribedPredicate predicate) to CodeUnitsShould.
Now I’m working on possibility to have a rule like: 
should().onlyBeCalled().{byClassesThat()/methodsThat()/constructorsThat()}.

Could you also let me know if my understanding of the issue is correct for below points?
1. I’ve implemented condition “code unit should only be called by”, but few possibilities came to my mind, not sure if I’ve decided on the correct one:
    * code unit should be called only by classes that,
    * code unit should be called by at least one class that,
    * code unit should be called by all classes that (but can be called also by other)
    * code unit should be called by all classes that (and can’t be called by any other).
2. should().onlyBeCalled().byMethodsThat() - here I’m checking if methods calling selected code unit meet provided criteria but still this code unit can be called by constructors not meeting criteria. Is it correct?

I would appreciate any feedback.
Justyna

PR #810, opened 2022-02-24
#704: added onlyBeCalledByClassesThat conjuntion - initial part

JKLedzion:
Initial part of work for issue #704. Work is in progress but it would be great to receive some feedback if it goes in right direction.

I’ve started from adding method onlyBeCalledByClassesThat(DescribedPredicate predicate) to CodeUnitsShould.
Now I’m working on possibility to have a rule like: 
should().onlyBeCalled().{byClassesThat()/methodsThat()/constructorsThat()}.

Could you let me know if my understanding of the issue is correct for below points?
1. I’ve implemented condition “code unit should only be called by”, but few possibilities came to my mind, not sure if I’ve decided on the correct one:
    * code unit should be called only by classes that,
    * code unit should be called by at least one class that,
    * code unit should be called by all classes that (but can be called also by other)
    * code unit should be called by all classes that (and can’t be called by any other).
2. should().onlyBeCalled().byMethodsThat() - here I’m checking if methods calling selected code unit meet provided criteria but still this code unit can be called by constructors not meeting criteria. Is it correct?

I would appreciate any feedback.
Justyna

Signed-off-by: Justyna Kubica-Ledzion <justyna.kubica-ledzion@digitalnewagency.com>

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaCodeUnit.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaCodeUnit.java
@@ -172,6 +172,9 @@ public abstract class JavaCodeUnit
     }
 
     @PublicAPI(usage = ACCESS)
+    public abstract Set<? extends JavaCall<?>> getCallsOfSelf();
+
+    @PublicAPI(usage = ACCESS)
     public Set<JavaMethodCall> getMethodCallsFromSelf() {
         return methodCalls;
     }
@@ -222,6 +225,11 @@ public abstract class JavaCodeUnit
 
     @PublicAPI(usage = ACCESS)
     public boolean isConstructor() {
+        return false;
+    }
+
+    @PublicAPI(usage = ACCESS)
+    public boolean isMethod() {
         return false;
     }
 
@@ -354,6 +362,16 @@ public abstract class JavaCodeUnit
                 }
             };
         }
+
+        @PublicAPI(usage = ACCESS)
+        public static DescribedPredicate<JavaCodeUnit> method() {
+            return new DescribedPredicate<JavaCodeUnit>("method") {
+                @Override
+                public boolean apply(JavaCodeUnit input) {
+                    return input.isMethod();
+                }
+            };
+        }
     }
 
     @PublicAPI(usage = ACCESS)
@@ -377,6 +395,15 @@ public abstract class JavaCodeUnit
                     }
                 };
             }
+
+            @PublicAPI(usage = ACCESS)
+            public static final ChainableFunction<JavaCodeUnit, Set<? extends JavaCall<?>>> GET_CALLS_OF_SELF =
+                    new ChainableFunction<JavaCodeUnit, Set<? extends JavaCall<?>>>() {
+                        @Override
+                        public Set<? extends JavaCall<?>> apply(JavaCodeUnit input) {
+                            return input.getCallsOfSelf();
+                        }
+                    };
         }
     }
 }
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaMethod.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaMethod.java
@@ -70,6 +70,12 @@ public class JavaMethod extends JavaCodeUnit {
         return annotationDefaultValue;
     }
 
+    @Override
+    @PublicAPI(usage = ACCESS)
+    public boolean isMethod() {
+        return true;
+    }
+
     @PublicAPI(usage = ACCESS)
     public Set<JavaMethodCall> getCallsOfSelf() {
         return getReverseDependencies().getCallsTo(this);
--- a/archunit/src/main/java/com/tngtech/archunit/lang/conditions/ArchConditions.java
+++ b/archunit/src/main/java/com/tngtech/archunit/lang/conditions/ArchConditions.java
@@ -53,6 +53,7 @@ import com.tngtech.archunit.core.domain.JavaModifier;
 import com.tngtech.archunit.core.domain.properties.HasAnnotations;
 import com.tngtech.archunit.core.domain.properties.HasModifiers;
 import com.tngtech.archunit.core.domain.properties.HasName;
+import com.tngtech.archunit.core.domain.properties.HasOwner;
 import com.tngtech.archunit.core.domain.properties.HasOwner.Functions.Get;
 import com.tngtech.archunit.core.domain.properties.HasOwner.Predicates.With;
 import com.tngtech.archunit.core.domain.properties.HasSourceCodeLocation;
@@ -91,6 +92,9 @@ import static com.tngtech.archunit.core.domain.JavaClass.Predicates.simpleNameCo
 import static com.tngtech.archunit.core.domain.JavaClass.Predicates.simpleNameEndingWith;
 import static com.tngtech.archunit.core.domain.JavaClass.Predicates.simpleNameStartingWith;
 import static com.tngtech.archunit.core.domain.JavaClass.Predicates.type;
+import static com.tngtech.archunit.core.domain.JavaCodeUnit.Functions.Get.GET_CALLS_OF_SELF;
+import static com.tngtech.archunit.core.domain.JavaCodeUnit.Predicates.constructor;
+import static com.tngtech.archunit.core.domain.JavaCodeUnit.Predicates.method;
 import static com.tngtech.archunit.core.domain.JavaConstructor.CONSTRUCTOR_NAME;
 import static com.tngtech.archunit.core.domain.JavaMember.Predicates.declaredIn;
 import static com.tngtech.archunit.core.domain.JavaModifier.FINAL;
@@ -1205,6 +1209,35 @@ public final class ArchConditions {
         return new DoesConditionByPredicate<>(declareThrowableOfType);
     }
 
+    @PublicAPI(usage = ACCESS)
+    public static ArchCondition<JavaCodeUnit> onlyBeCalledByClassesThat(DescribedPredicate<? super JavaClass> predicate) {
+        ChainableFunction<JavaAccess<?>, JavaCodeUnit> origin = JavaAccess.Functions.Get.origin();
+        ChainableFunction<HasOwner<JavaClass>, JavaClass> owner = Get.owner();
+        return new CodeUnitOnlyCallsCondition<>("only be called by classes that " + predicate.getDescription(),
+                origin.then(owner).is(predicate), GET_CALLS_OF_SELF);
+    }
+
+    @PublicAPI(usage = ACCESS)
+    public static ArchCondition<JavaCodeUnit> onlyBeCalledByCodeUnitsThat(DescribedPredicate<? super JavaMember> predicate) {
+        ChainableFunction<JavaAccess<?>, ? extends JavaCodeUnit> origin = JavaAccess.Functions.Get.origin();
+        return new CodeUnitOnlyCallsCondition<>("only be called by code units that " + predicate.getDescription(),
+                origin.is(predicate), GET_CALLS_OF_SELF);
+    }
+
+    @PublicAPI(usage = ACCESS)
+    public static ArchCondition<JavaCodeUnit> onlyBeCalledByMethodsThat(DescribedPredicate<? super JavaMember> predicate) {
+        ChainableFunction<JavaAccess<?>, ? extends JavaCodeUnit> origin = JavaAccess.Functions.Get.origin();
+        return new CodeUnitOnlyCallsCondition<>("only be called by methods that " + predicate.getDescription(),
+                origin.is(method().and(predicate)), GET_CALLS_OF_SELF);
+    }
+
+    @PublicAPI(usage = ACCESS)
+    public static ArchCondition<JavaCodeUnit> onlyBeCalledByConstructorsThat(DescribedPredicate<? super JavaMember> predicate) {
+        ChainableFunction<JavaAccess<?>, ? extends JavaCodeUnit> origin = JavaAccess.Functions.Get.origin();
+        return new CodeUnitOnlyCallsCondition<>("only be called by constructors that " + predicate.getDescription(),
+                origin.is(constructor().and(predicate)), GET_CALLS_OF_SELF);
+    }
+
     private static <T extends HasDescription & HasSourceCodeLocation> String createMessage(T object, String message) {
         return object.getDescription() + " " + message + " in " + object.getSourceCodeLocation();
     }
@@ -1223,7 +1256,7 @@ public final class ArchConditions {
             new IsConditionByPredicate<>("a local class", JavaClass.Predicates.LOCAL_CLASSES);
 
     private static class HaveOnlyModifiersCondition<T extends HasModifiers & HasDescription & HasSourceCodeLocation>
-            extends AllAttributesMatchCondition<T> {
+            extends AllAttributesMatchCondition<T, JavaClass> {
 
         private final Function<JavaClass, ? extends Collection<T>> getHasModifiers;
 
--- a/archunit/src/main/java/com/tngtech/archunit/lang/syntax/elements/CodeUnitsShould.java
+++ b/archunit/src/main/java/com/tngtech/archunit/lang/syntax/elements/CodeUnitsShould.java
@@ -70,6 +70,19 @@ public interface CodeUnitsShould<CONJUNCTION extends CodeUnitsShouldConjunction<
     CONJUNCTION notHaveRawParameterTypes(Class<?>... parameterTypes);
 
     /**
+     * Asserts that only certain objects call the code units selected by this rule.
+     * <br><br>
+     * E.g.
+     * <pre><code>
+     * {@link ArchRuleDefinition#codeUnits() codeUnits()}.{@link GivenCodeUnits#should() should()}.{@link CodeUnitsShould#onlyBeCalled() onlyBeCalled()}.{@link OnlyBeCalledSpecification#byClassesThat() byClassesThat()}.{@link ClassesThat#belongToAnyOf(Class[]) belongToAnyOf(SomeClass.class)}
+     * </code></pre>
+     *
+     * @return A syntax element that allows restricting how code units can be called
+     */
+    @PublicAPI(usage = ACCESS)
+    OnlyBeCalledSpecification<CONJUNCTION> onlyBeCalled();
+
+    /**
      * Asserts that {@link JavaCodeUnit JavaCodeUnits} have the specified fully qualified raw parameter type names.
      * <br><br>
      * E.g.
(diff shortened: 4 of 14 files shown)


codecholeric:
Thanks a lot for tackling this, we really appreciate it :smile: Also good idea to reach out early so we can discuss unclear points :+1: 

To answer your questions:
1) To be consistent with the rest "should only be called by classes that" should follow your first point. I.e. every class that calls the respective code unit should match the predicate. Otherwise we would e.g. use "any class that" or sth. like that. But we should also add Javadoc in the end to explain this precisely :slightly_smiling_face: 
2) If you say "only be called by methods that" I would forbid constructor calls. You are right that both is valid (again, that's why we should definitely add Javadoc), you could as well ignore constructor calls and only assert method calls. But we will also have "only be called by code units that", and that one will match methods and constructors, so given that I think it makes sense to forbid constructors when we talk about "only...by methods". Also, if we ponder how you would write the condition if you really want to forbid constructors. Then you would have to write something like "only be called by methods that ... are not constructors" and that just sounds weird to me :wink:

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaCodeUnit.java
```diff
@@ -171,6 +171,12 @@ public JavaClass getRawReturnType() {
         return fieldAccesses;
     }
 
+    @PublicAPI(usage = ACCESS)
+    public abstract Set<? extends JavaAccess<?>> getCallsOfSelf();
```
codecholeric, line 175: Couldn't the bound here be `JavaCall` instead of `JavaAccess`?

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaCodeUnit.java
```diff
@@ -359,6 +370,30 @@ private Get() {
                     }
                 };
             }
+
+            /**
+             * @see #getCallsOfSelf()
+             */
+            @PublicAPI(usage = ACCESS)
+            public static final ChainableFunction<JavaCodeUnit, Set<? extends JavaAccess<?>>> GET_CALLS_TO_SELF =
```
codecholeric, line 378: By convention the constant name should be the exact method name in upper case. I.e. if the getter is `getCallsOfSelf` then the function should be `GET_CALLS_OF_SELF`

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaConstructor.java
```diff
@@ -64,11 +64,22 @@ public boolean isConstructor() {
...
+        return false;
+    }
+
     @PublicAPI(usage = ACCESS)
     public Set<JavaConstructorCall> getCallsOfSelf() {
         return getReverseDependencies().getCallsTo(this);
     }
 
+    @PublicAPI(usage = ACCESS)
+    public Set<JavaConstructorCall> getMethodCallsOfSelf() {
```
codecholeric, line 79: This method is confusing to me because "get method calls...` sounds to me as if it would return `JavaMethodCall`. But instead it's returning `JavaConstructorCall` which is no "method call". I know where you're coming from, but I think we need to find a different naming here :thinking: Maybe it would be something like `getCallsOfSelfByMethods` or `...FromMethods`? :thinking: Or if we only need it in one place we don't add a new API method and simply filter in the one place :man_shrugging: We should ponder carefully if this is a common use case that should go in the API or not...

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaConstructor.java
```diff
@@ -64,11 +64,22 @@ public boolean isConstructor() {
         return true;
     }
 
+    @Override
+    @PublicAPI(usage = ACCESS)
+    public boolean isMethod() {
+        return false;
```
codecholeric, line 70: You don't need to override this because the default is `false` anyway, no?

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaMethod.java
```diff
@@ -70,11 +70,28 @@
         return annotationDefaultValue;
     }
 
+    @Override
+    @PublicAPI(usage = ACCESS)
+    public boolean isConstructor() {
+        return false;
```
codecholeric, line 76: Same here, the default is `false` anyway. As long as we go with a default (which you could question :man_shrugging:) we should use it then I guess...

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaCodeUnit.java
JKLedzion: Changed.

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaCodeUnit.java
JKLedzion: Corrected.

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaConstructor.java
JKLedzion: Method was removed as all calls should be analyzed not only calls done by methods.

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaConstructor.java
JKLedzion: Yes, I've missed that.

JKLedzion:
Thanks @codecholeric for quick review and feedback.

I’ve made corrections and added methods _onlyBeCalled().byConstructorsThat(DescribedPredicate predicate)_ /  _onlyBeCalled().byCodeUnitsThat(DescribedPredicate predicate)_.

Ad. 2. I’ve implemented logic as follows:
_.should().onlyBeCalled().byMethodsThat_ - calls done by methods are assessed against provided predicate, calls done by constructors are treated as violation of the rule. Method _getMethodCallsOfSelf()_ was removed as all calls should be analyzed.

Currently below listed posibilities are implemented but I’m not sure if all of them are needed.
Maybe only following pattern should be left: 
_{constructors()/methods()/codeUnits()}.should().onlyBeCalled().by…_
WDYT?

constructors()/methods()/codeUnits().should().onlyBeCalledByClassesThat(DescribedPredicate predicate)
constructors()/methods()/codeUnits().should().onlyBeCalledByClassesThat()
constructors()/methods()/codeUnits().should(onlyBeCalledByClassesThat(DescribedPredicate predicate)
constructors()/methods()/codeUnits().should().onlyBeCalled().byClassesThat(DescribedPredicate predicate)
constructors()/methods()/codeUnits().should().onlyBeCalled().byClassesThat()
constructors()/methods()/codeUnits().should().onlyBeCalled().byMethodsThat(DescribedPredicate predicate)
constructors()/methods()/codeUnits().should().onlyBeCalled().byConstructorsThat(DescribedPredicate predicate)
constructors()/methods()/codeUnits().should().onlyBeCalled().byCodeUnitsThat(DescribedPredicate predicate)

JKLedzion:
Thanks for confirmation.
I’ve removed _onlyBeCalledByClassesThat_ from _CodeUnitsShould_ leaving only _onlyBeCalled().byClassesThat_.
You mentioned earlier adding precise JavaDoc for this new methods. I put them in _OnlyBeCalledSpecification_ interface but I’m not sure if I was precisely enough 😄  WDYT?

archunit/src/main/java/com/tngtech/archunit/lang/syntax/elements/CodeUnitsShould.java
```diff
@@ -69,6 +69,19 @@
     @PublicAPI(usage = ACCESS)
     CONJUNCTION notHaveRawParameterTypes(Class<?>... parameterTypes);
 
+    /**
+     * Asserts that only certain objects call the code units selected by this rule.
+     * <br><br>
+     * E.g.
+     * <pre><code>
+     * {@link ArchRuleDefinition#codeUnits() codeUnits()}.{@link GivenCodeUnits#should() should()}.{@link CodeUnitsShould#onlyBeCalled()} onlyBeCalled().{@link OnlyBeCalledSpecification#byClassesThat(DescribedPredicate)}  byClassesThat(equivalentTo(SomeClass.class))}
```
codecholeric, line 77: Instead of the version with predicate `byClassesThat(equivalentTo(SomeClass.class))` maybe we can give an example with the fluent API, e.g. `byClassesThat().areAnnotatedWith(SomeAnnotation.class)` or something like that? :thinking:

archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/CodeUnitsShouldTest.java
```diff
@@ -277,6 +285,345 @@ public void throwable_type_predicates(ArchRule rule, Collection<String> expected
         assertThat(actualMembers).containsOnlyElementsOf(expectedMembers);
     }
 
+    @DataProvider
+    public static Object[][] restricted_constructor_calls_by_classes_rules() {
+        return $$(
+                $(constructors().should(onlyBeCalledByClassesThat(equivalentTo(ClassCorrectlyCallingMethodAndConstructor.class)))),
```
codecholeric, line 291: Why not use `JavaClass.Predicates.belongToAnyOf(..)` as the predicate, instead of `equivalentTo`? Then it would be consistent with the fluent API case two lines below...

archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/CodeUnitsShouldTest.java
```diff
@@ -47,14 +53,16 @@
...
+import static com.tngtech.archunit.testutil.Assertions.assertThat;
 import static com.tngtech.java.junit.dataprovider.DataProviders.$;
 import static com.tngtech.java.junit.dataprovider.DataProviders.$$;
 import static java.util.regex.Pattern.quote;
-import static org.assertj.core.api.Assertions.assertThat;
 
 @RunWith(DataProviderRunner.class)
 public class CodeUnitsShouldTest {
 
+    private static final String METHOD_CALL_WRONGLY = "callWrongly()";
```
codecholeric, line 64: You could maybe move that constant into the class `ClassWronglyCallingMethodAndConstructor`, then it would be easier to see what it refers to without doing a text search :slightly_smiling_face:

archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/CodeUnitsShouldTest.java
```diff
@@ -277,6 +285,345 @@ public void throwable_type_predicates(ArchRule rule, Collection<String> expected
...
+                        quote(ClassWronglyCallingMethodAndConstructor.class.getName()),
+                        quote(METHOD_CALL_WRONGLY),
+                        quote(ClassWithMethodAndConstructor.class.getName()),
+                        quote(CONSTRUCTOR_ONE_ARG)));
+    }
+
+    @DataProvider
+    public static Object[][] restricted_code_unit_calls_by_methods_rules() {
+        return $$(
+                $(codeUnits().should().onlyBeCalled().byMethodsThat(declaredIn(ClassCorrectlyCallingMethodAndConstructor.class))),
```
codecholeric, line 368: What about switching the order of the two data points so it matches the ones before where `onlyBeCalledByClassesThat` always comes first?

codecholeric (changes requested):
Thanks a lot for all this hard word, very nice :heart_eyes: To me your Javadoc looks good, I added some minor suggestions to adjust it, but in general it's exactly how I imagined it :smiley: Doesn't need to be a novel, but the important point (e.g. that `byConstructors` excludes methods) should be prominent :+1:
I added some comments about minor things I noticed that could maybe be improved if you're still motivated?
In a future PR we can also fix the generics on `MembersThat`/`CodeUnitsThat`/... so that it can be offered as a fluent API completion of `byCodeUnitsThat()` :slightly_smiling_face: (But I consider that out of scope for this PR)

JKLedzion:
Yes, still motivated😃 but I will be able to work on changes from Wednesday next week.

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaCodeUnit.java
```diff
@@ -359,6 +377,18 @@ private Get() {
                     }
                 };
             }
+
+            /**
+             * @see #getCallsOfSelf()
```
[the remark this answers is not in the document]
JKLedzion: Removed for consistency.

archunit/src/main/java/com/tngtech/archunit/lang/conditions/CodeUnitCallsCondition.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Collection;
+
+import com.tngtech.archunit.base.DescribedPredicate;
+import com.tngtech.archunit.base.Function;
+import com.tngtech.archunit.core.domain.JavaAccess;
+import com.tngtech.archunit.core.domain.JavaCodeUnit;
+import com.tngtech.archunit.lang.ArchCondition;
+import com.tngtech.archunit.lang.ConditionEvents;
+
+class CodeUnitCallsCondition extends ArchCondition<JavaCodeUnit> {
```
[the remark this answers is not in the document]
JKLedzion: I've made a change as suggested, hope I've understood it correctly.

archunit/src/main/java/com/tngtech/archunit/lang/syntax/OnlyBeCalledConstructorSpecificationInternal.java
```diff
@@ -0,0 +1,62 @@
...
+
+import com.tngtech.archunit.base.DescribedPredicate;
+import com.tngtech.archunit.base.Function;
+import com.tngtech.archunit.core.domain.JavaClass;
+import com.tngtech.archunit.core.domain.JavaMember;
+import com.tngtech.archunit.lang.conditions.ArchConditions;
+import com.tngtech.archunit.lang.syntax.elements.ClassesThat;
+import com.tngtech.archunit.lang.syntax.elements.OnlyBeCalledSpecification;
+
+class OnlyBeCalledConstructorSpecificationInternal implements OnlyBeCalledSpecification<ConstructorsShouldInternal> {
```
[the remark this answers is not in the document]
JKLedzion: Done

archunit/src/main/java/com/tngtech/archunit/lang/syntax/elements/CodeUnitsShould.java
JKLedzion: I've used example from tests.

archunit/src/main/java/com/tngtech/archunit/lang/syntax/elements/OnlyBeCalledSpecification.java
```diff
@@ -0,0 +1,60 @@
...
+     * @return A syntax conjunction element, which can be completed to form a full rule
+     */
+    @PublicAPI(usage = ACCESS)
+    CONJUNCTION byCodeUnitsThat(DescribedPredicate<? super JavaMember> predicate);
+
+    /**
+     * @return A syntax element that allows restricting which classes the call should be from
+     */
+    @PublicAPI(usage = ACCESS)
+    ClassesThat<CONJUNCTION> byClassesThat();
```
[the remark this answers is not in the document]
JKLedzion: Sorting of methods changed.

archunit/src/main/java/com/tngtech/archunit/lang/conditions/ArchConditions.java
```diff
@@ -1205,6 +1209,35 @@ public static AllDependenciesCondition onlyHaveDependenciesWhere(DescribedPredic
         return new DoesConditionByPredicate<>(declareThrowableOfType);
     }
 
+    @PublicAPI(usage = ACCESS)
+    public static ArchCondition<JavaCodeUnit> onlyBeCalledByClassesThat(DescribedPredicate<? super JavaClass> predicate) {
+        ChainableFunction<JavaAccess<?>, JavaCodeUnit> origin = JavaAccess.Functions.Get.origin();
+        ChainableFunction<HasOwner<JavaClass>, JavaClass> owner = Get.owner();
+        return new CodeUnitCallsCondition(Joiner.on(" ").join("only be called by classes that", predicate.getDescription()),
```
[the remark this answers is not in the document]
JKLedzion: > Wouldn't it be simpler to just write `"only be called by classes that " + predicate.getDescription()`?
Yes, it's simpler - changed :smiley:

> But if we unify `CodeUnitCallsCondition` it would likely just take description and predicate into the constructor separately anyway 🤷‍♂️ (at least that's how the respective `Classes...` does it)
I've left it as it is for now as I'm not sure how to change it. There are 4 methods in _ArchConditions_ using the same object _CodeUnitOnlyCallsCondition_, each with different description (i.e. "only be called by classes that “, “only be called by methods that "), that is why I'm passing description here. Also I’m passing extended predicate (i.e. origin.is(method().and(predicate)) so extracting predicate description inside _CodeUnitOnlyCallsCondition_ constructor would end up with description: “… should only be called by methods that is method and custom predicate”.

JKLedzion:
Hi, I've added modification for almost all comments, hope my understanding was correct 😄 

There is one which I'm not sure about:
[But if we unify CodeUnitCallsCondition it would likely just take description and predicate into the constructor separately anyway 🤷‍♂️ (at least that's how the respective Classes... does it)](https://github.com/TNG/ArchUnit/pull/810#discussion_r818374172)

There are 4 methods in _ArchConditions_ using the same class _CodeUnitOnlyCallsCondition_, each with different description (i.e. "only be called by classes that “, “only be called by methods that ")
that is why I'm passing description as a parameter to _CodeUnitOnlyCallsCondition_ constructor.
Also I’m passing extended predicate (i.e. _origin.is(method().and(predicate)_) so extracting predicate description inside _CodeUnitOnlyCallsCondition_ constructor would end up with description: “_… should only be called by methods that is method and custom predicate_”.

archunit/src/main/java/com/tngtech/archunit/lang/conditions/ArchConditions.java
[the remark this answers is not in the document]
codecholeric: Yeah, sorry, I got confused there :upside_down_face: It's already exactly how the other conditions do it, so never mind :slightly_smiling_face:

archunit/src/main/java/com/tngtech/archunit/lang/conditions/CodeUnitCallsCondition.java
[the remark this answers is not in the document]
codecholeric: Yes, you understood perfectly :+1: Nice and unified now :slightly_smiling_face: In the end if the name is as generic as `AllAttributesMatch`, then it also makes sense that the owner of these attributes is a generic type :+1:

codecholeric (approved):
Thanks for addressing all the comments and all your hard work!! I've squashed it now and adjusted the message a little bit to make it ready to be merged :slightly_smiling_face: 
Gonna merge as soon as CI is green :rocket:

codecholeric:
BTW: I've already fixed up the generics on `MembersThat` and others to make it reusable -> #832. I hope by that change it would now be easy to add this later on, i.e. that it's possible to add `onlyBeCalled().byMethodsThat()` in an easy way.

Issue #807, opened 2022-02-21
'Never found a JavaCodeUnit' for Kotlin 'groupingBy' in 0.23.0


pkubowicz:
Using 'groupingBy' in Kotlin:
```kotlin
val duplicatedIso = countries.groupingBy { it.iso }
```
works in ArchUnit 0.22.0 but crashes 0.23.0:
```
java.lang.IllegalStateException: Never found a JavaCodeUnit that matches supposed origin
CodeUnit{name='keyOf', descriptor=(Ljava/lang/Object;)Ljava/lang/Object;, declaringClassName='com.example.CountryValidator$failOnDuplicatedIso$$inlined$groupingBy$1'}
	at com.tngtech.archunit.core.importer.AccessRecord$Factory.getOrigin(AccessRecord.java:325)
	at com.tngtech.archunit.core.importer.AccessRecord$Factory.access$1200(AccessRecord.java:67)
	at com.tngtech.archunit.core.importer.AccessRecord$Factory$10.get(AccessRecord.java:314)
	at com.tngtech.archunit.core.importer.AccessRecord$Factory$10.get(AccessRecord.java:311)
	at com.tngtech.archunit.thirdparty.com.google.common.base.Suppliers$NonSerializableMemoizingSupplier.get(Suppliers.java:167)
	at com.tngtech.archunit.core.importer.AccessRecord$Factory$RawAccessRecordProcessed.getOrigin(AccessRecord.java:282)
	at com.tngtech.archunit.core.importer.ClassGraphCreator.tryProcess(ClassGraphCreator.java:146)
```

A project that reproduces the bug: https://github.com/pkubowicz/archunit-accessrecord-bug (see GitHub action result).

codecholeric:
Thanks a lot for raising the issue and in particular for taking the effort to provide a sample :smiley: This is likely due to the more strict search for matching methods to disambiguate bridge methods from real methods :thinking: I guess in Kotlin there is some compiler magic that doesn't match anymore now, but I'll look into it :+1:

codecholeric:
I think I found the culprit: It seems to happen if the generic signature of the return type of a method is out of sync with the descriptor :grimacing: I.e. The descriptor says the return type is `Object`, but the signature says it's `String` from the generic context. I'll try to release a fix within the next couple of days...

Issue #804, opened 2022-02-21
Never found a JavaCodeUnit that matches supposed origin CodeUnit for org.junit.jupiter.api.assertDoesNotThrow


aablsk:
Hi!

Thank you for providing and supporting this awesome package!

After upgrading from 0.22.0 to 0.23.0 unfortunately we stumbled across an issue.

**Observation**
When running the `classes().that()...` ArchUnit throws an `IllegalStateException` (with below stacktrace) if one of the classes in the packages that are being analyzed calls the JUnit method `org.junit.jupiter.api.assertDoesNotThrow`.

_Unfortunately we can not simply exclude test classes from ArchUnit, as we have ArchUnit tests that verify that other tests have a certain coverage._

**Workaround**
We're using the AssertJ equivalent `assertThatCode {...}.doesNotThrowAnyException()` of `assertDoesNotThrow` instead.

**Additional information**
We're using Spring Boot (gradle-plugin version: 2.6.3) with kotlin, Amazon Corretto Java 17.
We're using org.junit.jupiter:junit-jupiter:5.8.2

**Hypothesis**
This might be related to one of the bug fixes in 0.23.0

**Minimal test example that leads to Exception**
```
// this class needs to be analyzed by ArchUnit for the exception to occur
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.assertDoesNotThrow

class AssertDoesNotThrowTest {
    @Test
    fun breaksArchUnit() {
        assertDoesNotThrow {
        }
    }
}

```

**Stacktrace of exception**
```
java.lang.IllegalStateException: Never found a JavaCodeUnit that matches supposed origin CodeUnit{name='get', descriptor=()Ljava/lang/Object;, declaringClassName='{REDACTED}.AssertDoesNotThrowTest$breaksArchUnit$$inlined$assertDoesNotThrow$1'}
	at com.tngtech.archunit.core.importer.AccessRecord$Factory.getOrigin(AccessRecord.java:325)
	at com.tngtech.archunit.core.importer.AccessRecord$Factory.access$1200(AccessRecord.java:67)
	at com.tngtech.archunit.core.importer.AccessRecord$Factory$10.get(AccessRecord.java:314)
	at com.tngtech.archunit.core.importer.AccessRecord$Factory$10.get(AccessRecord.java:311)
	at com.tngtech.archunit.thirdparty.com.google.common.base.Suppliers$NonSerializableMemoizingSupplier.get(Suppliers.java:167)
[...]

Issue #802, opened 2022-02-20
`Automatically resolving transitive class dependencies with the following configuration:` is logged *a lot* in 0.23.0


famod:
For a simple test I'm seeing this multiple times:
```
2022-02-20 12:47:59,624 INFO  [com.tng.arc.cor.imp.DependencyResolutionProcess] (main) Automatically resolving transitive class dependencies with the following configuration:
import.dependencyResolutionProcess.maxIterationsForMemberTypes = 1
import.dependencyResolutionProcess.maxIterationsForAccessesToTypes = 1
import.dependencyResolutionProcess.maxIterationsForSupertypes = -1
import.dependencyResolutionProcess.maxIterationsForEnclosingTypes = -1
import.dependencyResolutionProcess.maxIterationsForAnnotationTypes = -1
import.dependencyResolutionProcess.maxIterationsForGenericSignatureTypes = -1
```
To me it looks as if this should only be logged once or/and only with DEBUG?

codecholeric:
That is strange, it should only be logged one time and when I test this locally it also is :thinking: I mean, it is logged one time for every import. Do you see it multiple times for a single call of `ClassFileImporter.import...`?
In any case, maybe `DEBUG` level is fair, to me it seemed like the logging of where the `archunit.properties` is read from, but I guess it's a finer level, because it's not relevant for typical use cases. Guess I'll change it to `DEBUG`...
If it's annoying to you you can simply set `DependencyResolutionProcess` to `WARN` though for now...

famod:
We have multiple AU test classes and thus multiple `@AnalyzeClasses`, which I think explains it.

Switching to DEBUG would be nice, yes.

famod:
Hum, I debugged this a little and that logging method is hit twice for the following test class:
```
@AnalyzeClasses(locations = OurCustomLocationProvider.class, importOptions = { OnlyIncludeTests.class })
class TestBasicsArchTest {

    @ArchTest
    final ArchRule doNotUseJunitAssertions = noClasses()
            .that().areNotAssignableTo(TestBasicsArchTest.class)
            .should().dependOnClassesThat().areAssignableTo(org.junit.jupiter.api.Assertions.class)
            .because("use org.assertj.core.api.Assertions instead");

    @ArchTest
    final ArchRule testMethodsShouldBePublic = ArchRuleDefinition.methods()
            .that().areMetaAnnotatedWith(Testable.class)
            .should().bePublic();
}
```
Is this to be expected?

[...]

PR #812, opened 2022-02-25
Fix bug where call origins cannot be found on descriptor signature mismatch

codecholeric:
With ArchUnit `0.23.0` we fixed an ambiguity problem when resolving method call origins (bridge methods would cause two methods with same name and parameter types, so we would pick one randomly, which could be the wrong/synthetic one). Unfortunately, this broke some Kotlin use cases, because inline functions would cause the compiler to create synthetic classes with methods where the `signature` and the `descriptor` do not match (e.g. the signature says the return type is `String`, but the descriptor claims `Object`). In these cases the erasure of the generic type does not match the raw type. But so far for the return type we derived the raw return type as erasure from the generic return type.
This is now fixed by keeping raw and generic return type completely separate, since we obviously cannot derive one from the other (the JVM spec also clearly states that this is a valid case, that descriptor and signature do not need to match exactly and are not validated against each other).

Resolves: #804
Resolves: #807
Resolves: #802


archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
```diff
@@ -898,6 +900,19 @@ void call(Target target) {
...
     }
 
+    @Test
+    public void identifies_call_origin_if_signature_and_descriptor_deviate() {
+        // Kotlin inline functions cause the creation of extra classes where the signature of the respective method shows
+        // the real types while the descriptor shows the erased types. The erasure of the real types from the signature
+        // does then not match the descriptor in some cases.
+
+        JavaClass javaClass = getOnlyElement(new ClassFileImporter().importLocations(singleton(Location.of(
+                relativeResourceUri(getClass(), "testexamples/MismatchBetweenDescriptorAndSignature.class")))));
```
hankem, line 910: I wonder whether we should somehow conserve the source code from https://github.com/pkubowicz/archunit-accessrecord-bug/blob/main/app/src/main/kotlin/com/example/CountryValidator.kt – and if it's only in the commit message.

hankem (approved):

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
codecholeric: Good idea, I added it as a comment. Doing this also made me strip down the example more :joy:

PR #813, opened 2022-02-25
Add support to test caught exceptions (fixes #591)

### LINKED ISSUE #591: Add support to test caught exceptions
I would like to be able to assert that some method calls are wrapped in a try/catch block catching certain type of exceptions.

Typical use case would be to make sure that runtime `WebApplicationExceptions` are caught when calling a remote resource using JAX-RS.

I'd be happy to try to contribute with a PR, if it could eventually be merged.

crizzis:
This will add `TryCatchBlocks` to the ArchUnit core and introduce `JavaCodeUnit.getTryCatchBlocks()` to examine try-catch blocks that have been parsed from the bytecode of a method or constructor. We also add an extension `JavaAccess.getContainingTryBlocks()` to make it easy to verify that certain accesses in code are wrapped into certain try-catch blocks (e.g. "whenever method x is called there should be a try-catch block to handle exception case y").

Resolves: #591

Signed-off-by: Krzysztof Sierszeń <krzysztof.sierszen@digitalnewagency.com>

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/Formatters.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/Formatters.java
@@ -89,7 +89,7 @@ public final class Formatters {
      * @return A {@link List} of fully qualified class names of the passed {@link Class} objects
      */
     @PublicAPI(usage = ACCESS)
-    public static List<String> formatNamesOf(Iterable<Class<?>> paramTypes) {
+    public static List<String> formatNamesOf(Iterable<? extends Class<?>> paramTypes) {
         ImmutableList.Builder<String> result = ImmutableList.builder();
         for (Class<?> paramType : paramTypes) {
             result.add(paramType.getName());
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/TryCatchBlock.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/TryCatchBlock.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2014-2022 TNG Technology Consulting GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.tngtech.archunit.core.domain;
+
+import java.util.Set;
+
+import com.google.common.collect.ImmutableSet;
+import com.tngtech.archunit.PublicAPI;
+import com.tngtech.archunit.core.domain.properties.HasOwner;
+import com.tngtech.archunit.core.domain.properties.HasSourceCodeLocation;
+import com.tngtech.archunit.core.importer.DomainBuilders.TryCatchBlockBuilder;
+
+import static com.google.common.base.MoreObjects.toStringHelper;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
+import static com.tngtech.archunit.core.domain.properties.HasName.Utils.namesOf;
+
+@PublicAPI(usage = ACCESS)
+public final class TryCatchBlock implements HasOwner<JavaCodeUnit>, HasSourceCodeLocation {
+    private final JavaCodeUnit owner;
+    private final Set<JavaClass> caughtThrowables;
+    private final SourceCodeLocation sourceCodeLocation;
+    private final Set<JavaAccess<?>> accessesContainedInTryBlock;
+
+    TryCatchBlock(TryCatchBlockBuilder builder) {
+        this.owner = checkNotNull(builder.getOwner());
+        this.caughtThrowables = ImmutableSet.copyOf(builder.getCaughtThrowables());
+        this.sourceCodeLocation = checkNotNull(builder.getSourceCodeLocation());
+        this.accessesContainedInTryBlock = ImmutableSet.copyOf(builder.getAccessesContainedInTryBlock());
+    }
+
+    @Override
+    @PublicAPI(usage = ACCESS)
+    public JavaCodeUnit getOwner() {
+        return owner;
+    }
+
+    @PublicAPI(usage = ACCESS)
+    public Set<JavaClass> getCaughtThrowables() {
+        return caughtThrowables;
+    }
+
+    @Override
+    @PublicAPI(usage = ACCESS)
+    public SourceCodeLocation getSourceCodeLocation() {
+        return sourceCodeLocation;
+    }
+
+    @PublicAPI(usage = ACCESS)
+    public Set<JavaAccess<?>> getAccessesContainedInTryBlock() {
+        return accessesContainedInTryBlock;
+    }
+
+    @Override
+    public String toString() {
+        return toStringHelper(this)
+                .add("owner", owner.getFullName())
+                .add("caughtThrowables", namesOf(caughtThrowables))
+                .add("location", sourceCodeLocation)
+                .toString();
+    }
+}
--- a/archunit/src/main/java/com/tngtech/archunit/core/importer/AccessRecord.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/importer/AccessRecord.java
@@ -54,6 +54,8 @@ interface AccessRecord<TARGET extends AccessTarget> {
 
     int getLineNumber();
 
+    RawAccessRecord getRaw();
+
     @Internal
     interface FieldAccessRecord extends AccessRecord<FieldAccessTarget> {
         AccessType getAccessType();
@@ -259,6 +261,11 @@ interface AccessRecord<TARGET extends AccessTarget> {
             @Override
             public int getLineNumber() {
                 return record.lineNumber;
+            }
+
+            @Override
+            public RawAccessRecord getRaw() {
+                return record;
             }
         }
 
--- a/archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
@@ -1,16 +1,22 @@
 package com.tngtech.archunit.core.importer;
 
+import java.io.BufferedReader;
+import java.io.IOException;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
+import java.net.Socket;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.function.Predicate;
 
 import com.google.common.collect.ImmutableSet;
+import com.tngtech.archunit.base.ChainableFunction;
+import com.tngtech.archunit.core.domain.AccessTarget;
 import com.tngtech.archunit.core.domain.AccessTarget.ConstructorCallTarget;
 import com.tngtech.archunit.core.domain.AccessTarget.FieldAccessTarget;
 import com.tngtech.archunit.core.domain.AccessTarget.MethodCallTarget;
@@ -25,7 +31,10 @@ import com.tngtech.archunit.core.domain.JavaField;
 import com.tngtech.archunit.core.domain.JavaFieldAccess;
 import com.tngtech.archunit.core.domain.JavaMethod;
 import com.tngtech.archunit.core.domain.JavaMethodCall;
+import com.tngtech.archunit.core.domain.SourceCodeLocation;
+import com.tngtech.archunit.core.domain.TryCatchBlock;
 import com.tngtech.archunit.core.domain.properties.HasOwner;
+import com.tngtech.archunit.core.domain.properties.HasOwner.Functions.Get;
 import com.tngtech.archunit.core.importer.DomainBuilders.FieldAccessTargetBuilder;
 import com.tngtech.archunit.core.importer.testexamples.callimport.CallsExternalMethod;
 import com.tngtech.archunit.core.importer.testexamples.callimport.CallsMethodReturningArray;
@@ -73,11 +82,17 @@ import com.tngtech.archunit.core.importer.testexamples.integration.ClassD;
 import com.tngtech.archunit.core.importer.testexamples.integration.ClassXDependingOnClassesABCD;
 import com.tngtech.archunit.core.importer.testexamples.integration.InterfaceOfClassX;
 import com.tngtech.archunit.core.importer.testexamples.specialtargets.ClassCallingSpecialTarget;
+import com.tngtech.archunit.core.importer.testexamples.trycatch.ClassHoldingMethods;
+import com.tngtech.archunit.core.importer.testexamples.trycatch.ClassWithComplexTryCatchBlocks;
+import com.tngtech.archunit.core.importer.testexamples.trycatch.ClassWithSimpleTryCatchBlocks;
+import com.tngtech.archunit.core.importer.testexamples.trycatch.ClassWithTryCatchBlockWithoutThrowables;
+import com.tngtech.archunit.core.importer.testexamples.trycatch.ClassWithTryWithResources;
 import com.tngtech.java.junit.dataprovider.DataProviderRunner;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
 import static com.google.common.collect.Iterables.getOnlyElement;
+import static com.tngtech.archunit.core.domain.JavaClass.Predicates.equivalentTo;
 import static com.tngtech.archunit.core.domain.JavaConstructor.CONSTRUCTOR_NAME;
 import static com.tngtech.archunit.core.domain.JavaFieldAccess.AccessType.GET;
 import static com.tngtech.archunit.core.domain.JavaFieldAccess.AccessType.SET;
@@ -92,13 +107,17 @@ import static com.tngtech.archunit.core.importer.DomainBuilders.newMethodCallTar
 import static com.tngtech.archunit.testutil.Assertions.assertThat;
 import static com.tngtech.archunit.testutil.Assertions.assertThatAccess;
 import static com.tngtech.archunit.testutil.Assertions.assertThatCall;
+import static com.tngtech.archunit.testutil.Assertions.assertThatTryCatchBlock;
 import static com.tngtech.archunit.testutil.Assertions.assertThatType;
 import static com.tngtech.archunit.testutil.Assertions.assertThatTypes;
+import static com.tngtech.archunit.testutil.JavaCallQuery.methodCallTo;
 import static com.tngtech.archunit.testutil.ReflectionTestUtils.constructor;
 import static com.tngtech.archunit.testutil.ReflectionTestUtils.field;
 import static com.tngtech.archunit.testutil.ReflectionTestUtils.method;
 import static com.tngtech.archunit.testutil.TestUtils.relativeResourceUri;
+import static com.tngtech.archunit.testutil.assertion.TryCatchBlocksAssertion.tryCatchBlock;
 import static java.util.Collections.singleton;
+import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.toSet;
 
 @RunWith(DataProviderRunner.class)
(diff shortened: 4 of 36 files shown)


archunit/src/main/java/com/tngtech/archunit/core/domain/JavaCall.java
```diff
@@ -15,17 +15,38 @@
...
+
 import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
 
 public abstract class JavaCall<T extends CodeUnitCallTarget> extends JavaCodeUnitAccess<T> {
-    JavaCall(JavaAccessBuilder<T, ?> builder) {
+    private static final Function<JavaClassDescriptor, CaughtThrowable> TO_CAUGHT_THROWABLE_CONVERTER = new Function<JavaClassDescriptor, CaughtThrowable>() {
+
+        @Override
+        public CaughtThrowable apply(JavaClassDescriptor input) {
+            return new CaughtThrowable(input);
```
crizzis, line 38: In this minimal version, only the exception type is retained in `CaughtThrowable`

This is a pretty large PR as it is, I'd suggest that any enhancements be added in a separate PR

archunit/src/main/java/com/tngtech/archunit/core/importer/AccessRecord.java
```diff
@@ -63,16 +55,31 @@
...
 
-        static Factory<RawAccessRecord, AccessRecord<ConstructorCallTarget>> forConstructorCallRecord() {
-            return new Factory<RawAccessRecord, AccessRecord<ConstructorCallTarget>>() {
+        static Factory<RawAccessRecord.ForMethodCall, ConstructorCallRecord> forConstructorCallRecord() {
+            return new Factory<RawAccessRecord.ForMethodCall, ConstructorCallRecord>() {
                 @Override
-                AccessRecord<ConstructorCallTarget> create(RawAccessRecord record, ImportedClasses classes) {
-                    return new RawAccessRecordProcessed<>(record, classes, CONSTRUCTOR_CALL_TARGET_FACTORY);
+                ConstructorCallRecord create(RawAccessRecord.ForMethodCall record, ImportedClasses classes) {
+                    return new RawConstructorCallRecordProcessed(record, classes);
```
crizzis, line 82: Quite a lot of layers to punch through :) I didn't quite get the idea behind `RawAccessRecord`/`RawXXXRecordProcessed`

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
```diff
@@ -752,6 +753,66 @@ public void classes_know_method_calls_to_themselves() {
         assertThat(calls).as("Method calls to class").isEqualTo(expected);
     }
 
+    @Test
+    public void method_calls_know_which_exceptions_are_handled() {
+        JavaClasses classes = new ClassFileImporter().importUrl(getClass().getResource("testexamples/trycatch"));
+        JavaClass classHoldingMethods = classes.get(ClassHoldingMethods.class);
+        JavaClass classWithTryCatchBlocks = classes.get(ClassWithTryCatchBlocks.class);
+        JavaMethod setSomeInt = classHoldingMethods.getMethod("setSomeInt", int.class);
```
crizzis, line 761: I wish there was some metamodel to generate those automatically :|

archunit/src/main/java/com/tngtech/archunit/core/importer/AccessRecord.java
codecholeric: I can explain the idea quickly: `RawAccessRecord` represents the raw data as collected during bytecode analysis (i.e. origin and target are just strings or type descriptors, etc.). `RawAccessRecordProcessed` represents the next step, when this raw data has been resolved to domain objects (e.g. the target is now not a string anymore, but a `JavaMethodCallTarget` which has a `JavaClass` as its owner, the origin is a `JavaCodeUnit`, etc.).

codecholeric:
Thank you so much for all your hard work!! :smiley: I'm really sorry, but I feel we should have discussed it more ahead though, because now looking deeper into it there are a lot of questions coming to my mind where things don't seem as clear as I thought on first sight :thinking: Because somehow the discussion how to model this just stopped and we never really came to a final conclusion (I just threw in that I would prefer `CaughtException` over `JavaClass`, but I didn't think about how to model it deeper).

But still, let's start discussing this a little though, because as I said there are some basic points for me that influence everything. The most important one for me is the model. @hankem was asking if it should be something like `JavaCall.getCaughtThrowables()`, as you modeled it, but looking at it now this looks confusing to me. Because the "call" hasn't caught the throwable, no? The method has caught the throwable around the call, hasn't it? So naively I would expect `JavaCodeUnit.getCaughtThrowables()` where I would get the throwables the method/constructor has caught. But then `CaughtThrowable` would need the info what `JavaCalls` were covered, which would seem really weird on a type with that name. So looking at it now, I tend to think that what we really want is something like a `TryCatchBlock`, very similar to what you have created internally anyway to track the boundaries of the try-catch-block :thinking: 

The current model in the PR is very close to the issue, i.e. the concrete issue would be super easy to solve with this model. But I'm not sure it's the best domain model in general. Because, if for example I would want to write a rule on all the caught exception types. Then I would now have to track down every single call, gather all the exception types and deduplicate, because multiple calls in the same block all have the same `CaughtThrowables` attached :thinking: Also, as I said, `JavaCall.getCaughtThrowables()` doesn't feel intuitive to me from a semantic point of view, i.e. not knowing about the API I would not expect to find this on `JavaCall` :thinking: 

I'm wondering if the following wouldn't be a more intuitive model:

```
class TryCatchBlock {
  int tryBlockStartLineNumber;
  int tryBlockEndLineNumber;
  Set<JavaClass> caughtThrowableTypes;

  boolean tryBlockContains(JavaCall call) {
    // compare line numbers to check that the call is contained? Or should we keep references to all `JavaCall` instead?
  }
}

class JavaCodeUnit {
  // ...
  Set<TryCatchBlock> getTryCatchBlocks() {..}
}
```

This model would make it a little more tedious to solve the concrete issue (I would now have to check all the try-catch-blocks to see if the relevant call is contained and the caught throwables match), but to me it feels better as a generic model to also implement other rules users might have for try-catch-blocks :thinking: (but I might be wrong :man_shrugging:) But it would also be possible to e.g. add `JavaCall.getContainingTryCatchBlocks()` or something like that to make it easier to solve this issue but still have a more generic model :thinking: 

What do you think about these points?

crizzis:
@codecholeric Your comment raises a couple of different points, so let me try and unpack it step by step. 

>  @hankem was asking if it should be something like JavaCall.getCaughtThrowables(), as you modeled it, but looking at it now this looks confusing to me. Because the "call" hasn't caught the throwable, no? The method has caught the throwable around the call, hasn't it?

I had a hard time coming up with an appropriate name, and I agree `getCaughtThrowables` is not the best one. Perhaps `getThrowablesHandledByCaller` would be better?

> So naively I would expect JavaCodeUnit.getCaughtThrowables() where I would get the throwables the method/constructor has caught. (...) Because, if for example I would want to write a rule on all the caught exception types. Then I would now have to track down every single call, gather all the exception types and deduplicate, because multiple calls in the same block all have the same CaughtThrowables attached 🤔

I believe we are discussing two separate use cases. My PR focused on covering @cpollet's use case and would not be easily solved by your proposal. Your use case, in turn, also sounds completely valid and the call for having `TryCatch` blocks listed under `JavaCodeUnit` seems perfectly justifiable.  

Perhaps then, our proposed solutions are both correct, and the association between `JavaCall`s and `TryCatchBlock`s should simply be a bidirectional one (I believe that's what you're suggesting by `getContainingTryCatchBlocks`)?

What I would suggest now, is to keep this solution as it is, and add another PR that would in addition create a `TryCatchBlock` entity, with a set of `CaughtThrowable` attached. I believe some of the functionality provided in this PR will be useful in implementing that other scenario. 

WDYT?

codecholeric:
Yes, a bidirectional association was what I meant by adding `JavaCall.getContainingTryCatchBlocks()`. I think it is very nice to be able to get the containing catch-blocks of a `JavaCall`. I'm just not sure we really need `TryCatchBlock` **and** `CaughtThrowable` then. Because how would you use `CaughtThrowable`? You would retrieve it, then retrieve the `Throwable` type, right? But `TryCatchBlock` can offer exactly the same, you could get all caught exception types and `TryCatchBlock` could simply hold a `Set<JavaClass>` or throwable types then (I only voted for something like `CaughtException` around it before, because `Set<JavaClass>` as a top-level type to model this can't be extended with any domain logic)

I fully agree about your argument with the line numbers, it would likely work *most* of the time, but it would be shady. Because in theory you can also do crazy one-line things like this:

```
try { call(); } catch(Wohoo e) {} call(); try { call(); } catch(Yahoo e) {}
```

Maybe the line numbers of start and end could still be interesting, but I think we can leave them out and see if users really want them :+1: 

Coming back to how to model the types, as mentioned, I think one domain object would be good enough. I think the best would be to record the necessary info (like labels, etc.) on import, then construct 

```
class JavaCodeUnit {
  // ...
  Set<TryCatchBlock> tryCatchBlocks;
}

class TryCatchBlock {
  Set<JavaCall> containedCalls; // (or Set<JavaAccess> containedAccesses? see below)
  Set<JavaClass> caughtThrowableTypes;
  // ...
}

class JavaCall {
  // ...
  Set<TryCatchBlock> getContainingTryCatchBlocks() {
    Set<TryCatchBlock> allTryCatchBlocks = getOwner().getTryCatchBlocks();
    return filter(allTryCatchBlocks, contains(this));
  }
}
```

This would offer a convenient API. (any better name than `getContainingTryCatchBlocks()`? Should it be `TryBlocks`, because it's only the try-block we're interested in? But that might be confusing? Or `getTryCatchBlocksWhereThisCodeUnitIsContainedWithinTheTryBlock()` :joy: )

One thing I'm wondering about is if we should just make it `JavaAccess` instead of `JavaCall` :thinking: Because any `JavaAccess` could be contained within a try-catch-block, right? Just that for a field access I can't imagine a use case for the info out of my head :thinking: But the implementation effort would certainly not be more and maybe it would make the model simpler if we do it for all `JavaAccess`? :thinking:

hankem:
> Maybe the line numbers of start [...] could still be interesting, but [...] leave them out and see if users really want them

For conditions based on `TryCatchBlock`s, it would certainly be nice if this domain object implemented `HasSourceCodeLocation`.

> `class TryCatchBlock {`
> `  Set<JavaCall> containedCalls; // (or Set<JavaAccess> containedAccesses? see below)`

Shouldn't this `Set` be `calls`(or `accesses`... 🤷‍♂️)`ContainedInTryBlock`? I wonder whether we might want to extend `TryCatchBlock` to also include `calls`(or `accesses`... 🤷‍♂️)`ContainedInCatchBlock` at some point.

P.S.  I'm very sorry that https://github.com/TNG/ArchUnit/issues/591#issuecomment-830625777 might have led in a wrong direction!

codecholeric:
True, it could implement `HasSourceCodeLocation` giving the line number of the try-block-start :+1: And yes, I agree that we should always be careful about "contained in try-catch-block", because that's ambiguous, if it's the try- or the catch-block (or even the finally block? Is that also in there? :thinking:)

codecholeric:
Also no worries about the comment @hankem, you only asked a question and we never finished the discussion. I should probably be more careful about slapping the `help-wanted` label onto such issues and only do it, when the issue is fully analysed. Also sorry for that, I'm often on such a tight schedule that I'm lacking the time to refine each issue before I know if somebody actually has time to tackle it :disappointed:

codecholeric:
BTW @crizzis, we have some code style files for IntelliJ and Eclipse checked in at https://github.com/TNG/ArchUnit/tree/main/develop. You can use those to make sure the formatting is consistent and the imports don't change unnecessarily!

crizzis:
@codecholeric Fixed the code style, converted `CaughtThrowable` into `TryCatchBlock`, and made it implement `HasSourceCodeLocation`. 

Is it okay with you for the `JavaCodeUnit -> TryCatchBlock` to be implemented in another PR?

(Also, I guess it makes sense for `TryCatchBlock` to then implement `HasOwner` as well, right?)

codecholeric:
Sorry that I haven't gotten back to you on this for so long!! I read through the code, then decided I should support you more (since obviously this whole back and forth wasn't super motivating for you, sorry about that), then got sick for a while :grimacing: Anyway, I invested quite some time now to pretty much bend it into the direction of having try-catch-blocks on the `JavaCodeUnit` instead of on the `JavaCall`. The reason is that I don't think it makes sense to add it in a way that we need to refactor it completely again to go for the other vision. I want to share some feedback with you, so that you hopefully can also get something out of it :slightly_smiling_face: 

First of all the code was very nice to read :+1: :smiley: And also respect that you managed to bite through all those layers, understand and handle the ASM visitor API and then implement it correctly, that was certainly not easy! I'll try to provide some constructive feedback about things I noticed:

* The most important part for the future is to watch out for other qualitative properties besides functionality. Because, while your code was written very cleanly and did what was necessary on the functional level, it was also quite inefficient. I tried to do a performance test (running `com.tngtech.archunit.core.importer.ClassFileImporterSlowTest#creates_JavaPackages`) and what took 4 seconds on the main branch now crashed with an `OutOfMemoryError` after 6 minutes. I think the reason was excessive filtering + copying of collections (in `TryCatchTracker` and co) and the massive duplication of `TryCatchBlocks` by adding them as separate objects to every `JavaCall`. `JavaCall` is a high volume object, so it is very easy to add something there that makes the heap explode. This was one reason I wanted to have exactly one `TryCatchBlock` for each try-catch-block in the code and add it to `JavaCodeUnit`. In my refactoring I had to watch out quite some bit to achieve a similar performance than before, because this is just a high-volume processing where we need to look at all the labels, so it's very easy to add a bottleneck there. I added things like the `active` flag simply to shave of some processing time :grimacing: 
* Before you create a PR you should run the full build once locally (that would have shown you that there are some errors as well as that the build now suddenly takes a very long time with your change)
* Another point is about structuring commits. It is fine to refactor some code, like move these assertions from inner classes to toplevel, but you should always do such things in a preparatory separate commit ahead. Because it makes it massively harder to see what you really added to implement the feature versus parts that were just moved. Also, sometimes it might be worth it to ask about it ahead. For example some of the inner classes in the main code are a conscious trade-off to group them somehow (while unfortunately not being able to use packages because of visibility :disappointed:). Anyway, for now I extracted those pure preparatory refactorings into a separate commit for you...
* The class `LabelSpan` would not have belonged to the `domain` part, since that is for the ArchUnit domain model that users use. But `LabelSpan` was a purely internal class. On the other hand `TryCatchBlock` would have belonged there, because `JavaCall` has a reference and in general `domain` should not have dependencies on `import`
* There was some functional part missing. If a try-catch-block contained an inner return statement the block would not have been completed correctly (unfortunately there was no test for that, I found it by accident when doing performance tests with some additional assertions in place)
* Domain objects like `TryCatchBlock` should never return `JavaClassDescriptor` (which was the return type for the `caughtThrowables`), because that class is `@Internal` and thus not to be used by users. Instead it has to be `JavaClass` for these cases.

In any case, I would be happy if you could look at my proposed changes and tell me what you think about them! Maybe it can also be made clearer in some parts or written more efficiently. I'm in particular not super happy that we now have to pass the `TryCatchBlockBuilders` to all the access creation, but I couldn't find any cleaner way to associate the completed accesses to the `TryCatchBlocks` :disappointed: If you want to compare performance between `main` and that branch, that might also be good (since I only used my own local machine for a limited number of cases)
If everything looks good to you I would suggest you just squash my review changes onto yours to create one clean commit that adds the support for try-catch-blocks.

crizzis:
> Sorry that I haven't gotten back to you on this for so long!! I read through the code, then decided I should support you more (since obviously this whole back and forth wasn't super motivating for you, sorry about that), then got sick for a while 😬

No worries, life happens, hope things are better now ;)

Unfortunately, life happens to me too, and while I promise to have a look at the suggestions, what I can't promise is that it will happen any time soon, as I will have much less time now to contribute to open source stuff. 

The ultimate goal is of course to ship new, shiny functionality to users, and most of the proposed changes sound very reasonable, so if you feel the feature can't wait, I wouldn't like to make my lack of time a showstopper. Thanks for the constructive feedback, and I can always revisit the PR even after it is merged, the code is not set in stone, after all. 

Otherwise I will have to ask you for some patience, but either way, it's probably not a good idea to leave a PR alive for too long lest it becomes hopelessly conflicted with the main branch. I'm leaving the decision up to you. 

Cheers!

codecholeric:
Okay, don't stress about it! I still have to look at your other PR as well, which I couldn't find the time yet :see_no_evil:

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
```diff
@@ -754,6 +776,247 @@ public void classes_know_method_calls_to_themselves() {
...
+
+        JavaConstructor constructor = javaClass.getConstructor();
+        Set<TryCatchBlock> tryCatchBlocks = constructor.getTryCatchBlocks();
+
+        assertThat(tryCatchBlocks)
+                .hasSize(3)
+                .areExactly(1, tryCatchBlock()
+                        .declaredIn(constructor)
+                        .catching(Exception.class)
+                        .atLocation(ClassWithComplexTryCatchBlocks.class, 14)
```
hankem, line 895: It might be interesting that the outer `TryCatchBlock` 's `accessesContainedInTryBlock` contains the inner one's.
(But it's admittedly tedious to describe in the test.)

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
```diff
@@ -754,6 +776,247 @@ public void classes_know_method_calls_to_themselves() {
...
+    }
+
+    private static class Data_all_accesses_know_which_exceptions_are_handled {
+        @SuppressWarnings("unused")
+        static class Target {
+            Object field;
+
+            Target() {
+            }
+
```
hankem, line 944: The default constructor should be good enough. 😉
```suggestion
```

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
hankem: Never mind, I later saw that this is implicitly covered in the `@Test` `all_accesses_know_which_exceptions_are_handled`.

archunit/src/test/java/com/tngtech/archunit/testutil/assertion/CodeUnitAccessAssertion.java
```diff
@@ -0,0 +1,83 @@
...
+            public boolean matches(AccessTarget.CodeUnitAccessTarget target) {
+                return target.getName().equals(codeUnitName)
+                        && HasName.Utils.namesOf(target.getRawParameterTypes()).equals(formatNamesOf(parameterTypes));
+            }
+        });
+    }
+
+    public CodeUnitAccessAssertion isWrappedWithTryCatchFor(Class<? extends Throwable> throwableType) {
+        assertThat(access.getContainingTryBlocks())
+                .as("wrapped with try/catch for %s", throwableType.getSimpleName())
```
hankem, line 53: Shouldn't the description rather simply be `"containing try blocks"`? 
`caughtThrowableOfType(throwableType)` already contains `expectedThrowable.getSimpleName()`.
(Same below.)

archunit/src/test/java/com/tngtech/archunit/testutil/assertion/CodeUnitAccessAssertion.java
```diff
@@ -0,0 +1,83 @@
...
+                .areExactly(1, caughtThrowableOfType(throwableType));
+        return newAssertion(access);
```
hankem, line 55: Wouldn't `this` be good enough? 😉
(Same below.)

archunit/src/test/java/com/tngtech/archunit/testutil/JavaCallQuery.java
```diff
@@ -0,0 +1,63 @@
...
+public class JavaCallQuery {
+
+    private FluentIterable<JavaMethodCall> calls;
+
+    private JavaCallQuery(Set<JavaMethodCall> calls) {
+        this.calls = FluentIterable.from(calls);
+    }
+
+    public static JavaCallQuery methodCallTo(final JavaMethod method) {
+        return new JavaCallQuery(method.getCallsOfSelf()).that(hasTarget(method));
```
hankem, line 20: Isn't `that(hasTarget(method))` redundant to `method.getCallsOfSelf()`?

archunit/src/test/java/com/tngtech/archunit/testutil/JavaCallQuery.java
```diff
@@ -0,0 +1,63 @@
...
+
+public class JavaCallQuery {
+
+    private FluentIterable<JavaMethodCall> calls;
+
+    private JavaCallQuery(Set<JavaMethodCall> calls) {
+        this.calls = FluentIterable.from(calls);
+    }
+
+    public static JavaCallQuery methodCallTo(final JavaMethod method) {
```
hankem, line 19: I usually wouldn't mix a static method between fields / constructors / instance methods.
If that's the recommended entry point, why not put it at the very top of the class?

hankem (approved):
I haven't yet played with the new feature for myself, and I didn't understand the one test with the 3 try-catch-blocks, but other than that, the code looks good to me!

archunit/src/main/java/com/tngtech/archunit/core/domain/Formatters.java
```diff
@@ -92,7 +92,7 @@ public static List<String> formatNamesOf(Class<?>... paramTypes) {
      * @return A {@link List} of fully qualified class names of the passed {@link Class} objects
      */
     @PublicAPI(usage = ACCESS)
-    public static List<String> formatNamesOf(Iterable<Class<?>> paramTypes) {
+    public static List<String> formatNamesOf(Iterable<? extends Class<?>> paramTypes) {
```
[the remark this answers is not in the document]
codecholeric: That is part of the Java Generics variance behavior. At least AFAIS without the `? extends` you can't use it for

```
Formatters.formatNamesOf(ImmutableList.of(String.class, Serializable.class));
```

while you can use it for

```
Formatters.formatNamesOf(ImmutableList.of(String.class, Object.class));
```

The problem seems to be that in the former case the inferred type is `List<Class<? extends Serializable>>` which seems to conflict with `Iterable<Class<?>>`. So to make the API more convenient I changed the signature to allow subtypes of `Class<?>`, because then the former case compiles without the need to explicitly fix the type by `ImmutableList.<Class<?>>`. Should maybe add a test though and split out the commit :wink: (because I saw that the final state actually compiles without this change, so I think I added it at some point when I stumbled over it, but the reason afterwards wasn't necessary to be added to the PR after all :man_shrugging:)

archunit/src/main/java/com/tngtech/archunit/core/domain/TryCatchBlock.java
```diff
@@ -0,0 +1,75 @@
...
+public final class TryCatchBlock implements HasOwner<JavaCodeUnit>, HasSourceCodeLocation {
+    private final JavaCodeUnit owner;
+    private final Set<JavaClass> caughtThrowables;
+    private final SourceCodeLocation location;
+    private final Set<JavaAccess<?>> accessesContainedInTryBlock;
+
+    TryCatchBlock(TryCatchBlockBuilder builder) {
+        this.owner = checkNotNull(builder.getOwner());
+        this.caughtThrowables = ImmutableSet.copyOf(builder.getCaughtThrowables());
+        this.location = checkNotNull(builder.getSourceCodeLocation());
```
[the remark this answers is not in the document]
codecholeric: Absolutely agree :wink: Changed!

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
```diff
@@ -754,6 +776,247 @@ public void classes_know_method_calls_to_themselves() {
...
+                .containsExactlyInAnyOrder(
+                        SourceCodeLocation.of(javaClass, 7),
+                        SourceCodeLocation.of(javaClass, 13));
+    }
+
+    @Test
+    public void imports_try_catch_blocks_from_same_starting_label() {
+        assumeTrue(!System.getProperty("java.version").startsWith("1.7."));
+
+        // this try-catch-combination creates a try-catch and a try-finally block from the same starting label
```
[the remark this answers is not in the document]
codecholeric: Yes, it is starting label, and honestly I'm also confused why exactly it is like this. But this is the exception table:

```
Exception table:
         from    to  target type
            27    29    27   any
             0    26    37   Class java/lang/Exception
            27    37    37   Class java/lang/Exception
```

I just found this strange case when testing and thus thought it's better to add a test for this, especially cause it's quite odd :man_shrugging: I don't exactly understand why the compiler generates this exception table, reading the JVM spec it doesn't seem to make sense to me... But on the other hand the code also doesn't make much sense, wrapping a `return` statement alone into `try-finally` :joy: Just one of these odd cases that we nevertheless should make sure they don't crash anything :man_shrugging:

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
```diff
@@ -754,6 +776,247 @@ public void classes_know_method_calls_to_themselves() {
...
+
+        assertThat(tryCatchBlocks).extracting("sourceCodeLocation")
+                .containsExactlyInAnyOrder(
+                        SourceCodeLocation.of(javaClass, 7),
+                        SourceCodeLocation.of(javaClass, 13));
+    }
+
+    @Test
+    public void imports_try_catch_blocks_from_same_starting_label() {
+        assumeTrue(!System.getProperty("java.version").startsWith("1.7."));
```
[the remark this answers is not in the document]
codecholeric: Luckily I'll rebase this on the JDK 8 branch and this will be obsolete :wink: That's also why I hadn't squashed it...

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
```diff
@@ -754,6 +776,247 @@ public void classes_know_method_calls_to_themselves() {
...
+                        reader.close();
+                    }
+                } catch (Exception ignored) {
+                }
+            }
+        }
+
+        JavaMethod method = new ClassFileImporter().importClass(SomeClass.class).getMethod("method", int.class, boolean.class);
+
+        assertThat(method.getTryCatchBlocks()).hasSize(3);
```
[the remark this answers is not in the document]
codecholeric: Sorry, I also can't explain it :disappointed: I just saw that this is how it is. I tried to strip down the original example as much as possible and this is where I ended. Somehow the whole combination, the parameters, the sort of nesting, the local variables, all that together causes this weird bytecode. But I have no full explanation why :grimacing: (if anybody ever reads this and can explain, please comment :pray:)

archunit/src/test/java/com/tngtech/archunit/testutil/assertion/CodeUnitAccessAssertion.java
codecholeric: Yeah, you're right, I overlooked that :+1:

archunit/src/test/java/com/tngtech/archunit/testutil/JavaCallQuery.java
codecholeric: Yes, I think so :joy: Adjusted it!!

archunit/src/test/java/com/tngtech/archunit/testutil/JavaCallQuery.java
codecholeric: Yes, agreed, but I put it on the bottom, because I think that's more consistent within the code base (we have many other places where factory methods are at the bottom, but I don't recall one where they are on the top). Of course I can also see why it makes sense to put it on the top when it has constructor semantics, but I want to keep it consistent here...

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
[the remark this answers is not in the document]
hankem: I would probably add a code comment to clarifiy that this test does not document a requirement, but just the current behavior:

> // might not be intuitive, but that's what the compiler generates 🤷‍♂️

Issue #809, opened 2022-02-24
failOnEmptyShould conflicts with optionalLayer


bjpe:
Hi, I have an ArchUnit test for a layered architecture, which has some mandatory layers and some optional layers, which are currently empts (i.e., there are no classes in the referred package).

This works fine with ArchUnit 0.22.0, but after an update to ArchUnit 0.23.0 I got

```
java.lang.AssertionError: Rule failed to check any classes. This means either that no classes have been passed to the rule at all, or that no classes passed to the rule matched the `that()` clause. To allow rules being evaluated without checking any classes you can set the ArchUnit property archRule.failOnEmptyShould = false
```
Removing the access restriction for the optional, empty layer fixes the test, as well as setting
```
archRule.failOnEmptyShould=false
```
but without this property the optional layers seem not so optional anymore.

codecholeric:
Thanks for raising this, it seems to go into the same direction as #803, #806 and #808 :+1: But I also see that there is an additional component, because layered architecture is composed from multiple sub-rules we need to watch out to in any case set each optional layer sub-rule to allow empty should, no matter what the configuration says.
/cc @oberprah

Issue #808, opened 2022-02-21
Make `failOnEmptyShould` overridable per rule


famod:
`failOnEmptyShould` is a nice new feature, thanks for that.

Unfortunately we have to switch it off because we have a Maven `test-jar` that contains multiple AU tests that are run in each module of out multi-module Maven project and in some modules not all tests catch respective classes (because there none with such criteria).

It would be nice if there was e.g. `*Should.failOnEmpty(boolean)`, to be able to override this case by case.

codecholeric:
Thanks for raising the issue :+1: I agree that we need some way to make this configurable on a per rule basis (compare #806). The question is just where to put it exactly :thinking: 

I guess it could go in the beginning `ArchRuleDefinition.failOnEmptyShould(false).classes().that()...` or in the end `classes()...should()...allowEmptyShould(false)`. In any case this probably means we have to add it individually to various places :disappointed: Would be nice if this could easily be done in a generic way in one central place (i.e. add a new rule API and it will automatically support it), but I don't see a way for that. Maybe for the standard rules APIs (e.g. `classes()`, `methods()`, ...) we can introduce a common parent interface, because in the end `containNumberOfElements(..)` also is a rules method that makes sense no matter what type of element it is. We could introduce a parent `ClassesShould extend ObjectsShould` that different `Should` types could extend to inherit the parts that make sense for all rule APIs. One question is, if we also need to add this to the `layeredArchitecture()`/`onionArchitecture()` and `slices()` APIs :thinking:

What do you think @oberprah? Do you want to tackle this as a follow-up? :wink:

Issue #803, opened 2022-02-21
Adding tests to verify patterns are not followed


jamietanna:
As a follow-up to #774, it now is not _as easy_ to write a rule that verifies that we do _not_ follow a pattern, such as the below which ensures we don't use field injection with Spring's Dependency Injection:

```java
  @ArchTest
  void requireNoAutowiredFieldInjection(JavaClasses classes) {
    classesThatAreNotTests()
        .and()
        .containAnyFieldsThat(
            describe(
                "are Autowired by Spring",
                f -> f.tryGetAnnotationOfType(Autowired.class).isPresent()))
        .should()
        .containNumberOfElements(equalTo(0))
        .check(classes);
  }
```

This now would require running the whole set of tests with `archRule.failOnEmptyShould` which although fine, leads to the risk that we have other tests that need to assert i.e. `.containNumberOfElements(greaterThan(0))`.

codecholeric:
Thanks for raising the issue :slightly_smiling_face: Yes, I understand that we need to make this configurable per rule! We will tackle this ASAP! (see #808 and #806)

PR #816, opened 2022-02-26
add the possibility to allow empty should per rule

oberprah:
This will allow to override the configuration property `archRule.failOnEmptyShould` on a per-rule basis. We overlooked the use case for certain rules that simply expect to have no match when we introduced `failOnEmptyShould`. An example would be rules like `classes().that(areUnwanted()).should().containNumberOfElements(equalTo(0))`. So making this only configurable on a global level renders it useless for a certain set of users. Also, we broke `optionalLayers` in `LayeredArchitecture` and `OnionArchitecture` with the default configuration, since individual access checks are modelled as rules. `LayeredArchitecture` and `OnionArchitecture` already have the concept of `optionalLayers` which is pretty much the same as expressed by `allowEmptyShould`. We now simply translate `allowEmptyShould` to the existing concept, otherwise the two concepts clash (e.g. declare an `optionalLayer` for which the allowed accesses are internally modelled as separate rule, and it still rejects empty classes if `failOnEmptyShould` is set to `true).`

Resolves: #803
Resolves: #808
Resolves: #809


codecholeric (approved):
Thank you so much for your support :smile: As discussed I just took it and fixed up some parts this time without prior discussion so we can release this ASAP :+1: Feel free to comment on things you notice on my updated version!!

archunit/src/main/java/com/tngtech/archunit/lang/AllowEmptyShould.java
```diff
@@ -0,0 +1,48 @@
...
+            return true;
+        }
+    },
+    FALSE {
+        @Override
+        public boolean isAllowed() {
+            return false;
+        }
+    },
+    DEFAULT {
```
hankem, line 33: Maybe `AS_CONFIGURED` instead of `DEFAULT`, to more clearly distinguish from the other _default_ internal to this `AllowEmptyShould` mode:
```
getPropertyOrDefault(FAIL_ON_EMPTY_SHOULD_PROPERTY_NAME, TRUE
```

archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/ClassesShouldTest.java
```diff
@@ -1767,6 +1770,26 @@ public void onlyCall_should_report_success_if_targets_are_non_resolvable(ArchRul
...
+            public void call() {
+                classes().that().areInterfaces().should().bePublic()
+                        .check(new ClassFileImporter().importClasses(getClass()));
+            }
+        }).isInstanceOf(AssertionError.class)
+                .hasMessageContaining("Rule failed to check any classes");
+    }
+
+    @Test
+    public void should_allow_empty_should_if_overridden_by_rule() {
```
hankem, line 1786: Maybe ...`overridden_by_configuration`?

archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/MembersShouldTest.java
```diff
@@ -672,6 +680,26 @@ public void containNumberOfElements_fails_on_mismatching_predicate() {
                         + "com.tngtech.archunit.lang.syntax.elements.testclasses.SimpleFieldAndMethod.violated()]");
     }
 
+    @Test
+    public void should_fail_on_empty_should_by_default() {
+        assertThatThrownBy(new ThrowableAssert.ThrowingCallable() {
+            @Override
+            public void call() {
+                members().that().haveName("doesNotExist").should().bePublic()
```
hankem, line 688: We could extract a semantic constant
```java
private ArchRule ruleWithEmptyShould = members().that().haveName("doesNotExist").should().bePublic();
```

archunit/src/test/java/com/tngtech/archunit/lang/ArchRuleTest.java
```diff
@@ -169,32 +168,43 @@ public void finish(ConditionEvents events) {
...
-    public void evaluation_fails_because_of_empty_set_of_classes_with_default_fail_on_empty_should() {
-        archConfigurationRule.removeProperty(FAIL_ON_EMPTY_SHOULD_PROPERTY_NAME);
-
+    public void evaluation_fails_because_of_empty_set_of_elements_with_default_fail_on_empty_should() {
         thrown.expect(AssertionError.class);
         thrown.expectMessage("Rule failed to check any classes");
         thrown.expectMessage(FAIL_ON_EMPTY_SHOULD_PROPERTY_NAME);
 
-        classes().should(ALWAYS_BE_VALID).evaluate(importClasses());
+        createPassingArchRule().evaluate(importClasses());
```
hankem, line 176: _Unrelated to the changes in this PR:_ I was confused why this fails, until I realized that `importClasses()` actually imports no classes. 🙈 
We could use `importClasses(/* no classes */)` or even have a semantic method / variable.

archunit/src/test/java/com/tngtech/archunit/lang/CompositeArchRuleTest.java
```diff
@@ -83,6 +91,31 @@ public void priority_is_passed() {
...
     }
 
+    @Test
+    public void fails_on_empty_should_by_default() {
+        assertThatThrownBy(new ThrowingCallable() {
+            @Override
+            public void call() {
+                CompositeArchRule
+                        .of(classes().should().bePublic())
+                        .and(classes().that().haveFullyQualifiedName("DoesNotExist").should().bePublic())
```
hankem, line 101: We could extract a semantic constant
```java
    private ArchRule ruleWithEmptyShould = CompositeArchRule
            .of(classes().should().bePublic())
            .and(classes().that().haveFullyQualifiedName("DoesNotExist").should().bePublic());
```

archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/ClassesShouldTest.java
```diff
@@ -1767,6 +1770,26 @@ public void onlyCall_should_report_success_if_targets_are_non_resolvable(ArchRul
...
+        assertThatThrownBy(new ThrowingCallable() {
+            @Override
+            public void call() {
+                classes().that().areInterfaces().should().bePublic()
+                        .check(new ClassFileImporter().importClasses(getClass()));
+            }
+        }).isInstanceOf(AssertionError.class)
+                .hasMessageContaining("Rule failed to check any classes");
+    }
+
```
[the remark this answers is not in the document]
codecholeric: I left them out, because we tested all three cases in `ArchRuleTest` and this here is more a sanity check for users of `ArchRule.Factory`, but maybe you're right :thinking: Can't harm I guess...

archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/MembersShouldTest.java
codecholeric: The problem about extracting this further as directly locally above the line where it is used is that it needs the combination of the imported classes and the rule. The rule by itself at the moment can't be `emptyShould` per se, only if you take into consideration that we're importing a specific class.
But thinking about it we can just use `alwaysFalse` as predicate, then it's alright again :+1:

archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/ClassesShouldTest.java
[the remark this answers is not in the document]
hankem: My initial confusion was probably due to the fact that two parts change between those two tests and because I overlooked that the configuration sets `failOnEmptyShould` whereas the ArchRule has `allowEmptyShould`. 🙈

archunit/src/test/java/com/tngtech/archunit/lang/ArchRuleTest.java
codecholeric: Changed to separate method :wink:

hankem (approved):

archunit/src/main/java/com/tngtech/archunit/library/Architectures.java
```diff
@@ -294,6 +270,11 @@ public ArchRule because(String reason) {
             return ArchRule.Factory.withBecause(this, reason);
         }
 
+        @Override
+        public ArchRule allowEmptyShould(boolean allowEmptyShould) {
+            return withOptionalLayers(allowEmptyShould);
```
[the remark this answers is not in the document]
codecholeric: Yes, it's a trade-off. But I think users won't use this method on a `LayeredArchitecture` exactly because `optionalLayers` makes more sense here. Thinking about it, we could write this in a Javadoc comment :thinking:

archunit/src/main/java/com/tngtech/archunit/library/Architectures.java
```diff
@@ -223,6 +222,7 @@ private void checkEmptyLayers(JavaClasses classes, EvaluationResult result) {
         private EvaluationResult evaluateLayersShouldNotBeEmpty(JavaClasses classes, LayerDefinition layerDefinition) {
             return classes().that(layerDefinitions.containsPredicateFor(layerDefinition.name))
                     .should(notBeEmptyFor(layerDefinition))
+                    .allowEmptyShould(true)
```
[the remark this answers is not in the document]
codecholeric: I guess you mean `allowEmptyShould(false)`? (otherwise I don't understand the question, since we're already using `allowEmptyShould(true)` :wink:) So theoretically yes, but the other method is nicer, because it doesn't crash immediately but gather the violations. So with `notBeEmptyFor(layerDefinition)` you will get (possibly multiple) violations like `layer x is empty`. With `allowEmptyShould(false)` you would get one error "rule failed to check any classes" (but which rule? :grimacing: Since there are several nested rules involved). I get it that it's not super nice, but for me at the moment it's "okay", because it's a hidden implementation details encapsulated in one private method. I can add a comment though, maybe that helps to understand it :man_shrugging:
Thinking further about it, maybe it would be nice to add the description to the exception :thinking: Even though it's irrelevant here, but if there are multiple rules nested you would at least have a chance to understand which rule is the problem :thinking:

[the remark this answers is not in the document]
oberprah: Yes, I meant `allowEmptyShould(false)` 🤦🏼‍♂️

Ah ok, yes, then it makes totally sense for me to use it like it is.

PR #820, opened 2022-03-01
Modified the IntelliJ test pattern for test exclusion

### LINKED ISSUE #696: INTELLIJ_TEST_PATTERN doesn't work with multi module projects
Firstly, thanks for creating this library. It's an awesome and very useful tool. 

I've come across a small issue when trying to exclude test classes when importing classes. It seems that the IntelliJ test pattern in `ImportOption.java` doesn't work with multi-module projects. In the project I work in I had to use `.*/out/test/.*` to exclude test classes properly. Would be possible to look into this and possibly expand the pattern to support multi module projects?

JCook21:
Plain IntelliJ projects (i.e. set up without any build tool) now seem to use the pattern `out/test/$MODULE_NAME` to compile class files to. Since we can safely assume that everything built into `out/test` belongs to the test scope, we can adjust the IntelliJ pattern to exclude everything in `.*/out/test/.*` to catch all cases, no matter if set up with a build tool or directly with IntelliJ.

Resolves: #696


codecholeric:
Thanks a lot of taking care of this immediately, really appreciated :smile:

JCook21:
Don't mind at all and thanks for merging this in quickly. This was a nice, easy first contribution to the project and I'll see if I can find some other issues to help with. :)

Issue #719, opened 2021-11-17
Let `@AnalyzeClasses` default to scan the package of the annotated class


odrotbohm:
Currently I have to write:

```java
package com.acme.sample;

@AnalyzeClasses(packagesOf = MyTestClass.class) 
class MyTestClass { … }
```

to analyze all classes contained in the very same package as the test class is located in. It would be nice if an "empty" annotation would simply default to that.

```java
package com.acme.sample;

@AnalyzeClasses // scans com.acme.sample by default
class MyTestClass { … }
```

Currently, the suggested arrangement seems to kick of scanning of the entire JVM and classpath, which is unlikely to be a useful default, is it?

codecholeric:
I agree that just `@AnalyzeClasses` has a little bit of a surprising default behavior. The reason it is like this at the moment was that I originally considered the use case for this default to be for example

```
@AnalyzeClasses(importOptions = DoNotIncludeArchives.class)
```

I.e. I don't want to specify any packages, I just exclude all dependencies by saying to ignore any JAR or JRT... But this turned out to be a little unreliable, because it might then depend on the environment. Run it with the build tool, suddenly things are JARs that were just plain compiled class files on some path in your IDE :thinking:

Thinking about my own use cases I also pretty much always use `packagesOf` or something similar and seldomly the whole classpath with some filter. 

I see the benefit of taking the package of the class by default, because that would make it very simple to just put the test in the root and it will test the project. And it's consistent to similar cases of other frameworks.

Only problem is that for people following the above pattern this would be a breaking change that could lead to existing tests not testing anything anymore (because the test e.g. resides alone in some subpackage, etc.). But maybe with a dominant breaking change announcement in the release notes it's alright :thinking:

hankem:
Just an idea: Should we leave `@AnalyzeClasses` as it is and additionally provide some `@AnalyzeClassesInThisPackage`? 🤔

I don't know whether users would actually suffer from the suggested change, but it would certainly be annoying if somebody who doesn't know about the new behavior – because I don't think that the name `@AnalyzeClasses` communicates it – has to debug why a test behaves differently in one package than in another, or even worse: loses test coverage without noticing...

codecholeric:
Yes, that is also what I'm afraid of. But I'm not sure if yet another annotation really makes this better, because it still keeps surprising behavior of the default.
[...]

PR #828, opened 2022-03-06
make `@AnalyzeClasses` import package of test class by default

codecholeric:
In particular coming from a Spring framework background it is a somewhat surprising behavior that a plain `@AnalyzeClasses` without any options will import the whole classpath. The more intuitive behavior would be to import the package of the annotated class. Since importing the whole classpath is likely not what most users want, and the performance drain is large, we will change the behavior to import the package of the annotated class by default if no further locations are specified.
However, this might break some existing use cases where scanning the whole classpath in combination with `ImportOptions` was used consciously. I.e. something like

```
@AnalyzeClasses(importOptions = MySpecialFilter.class)
```

To make sure ArchUnit can still satisfy this usecase we introduce a new property `@AnalyzeClasses(wholeClasspath = true)`, which is `false` by default, but can be set to `true` to restore the old behavior.

Resolves: #719

### DIFF
--- a/archunit-junit/junit4/src/main/java/com/tngtech/archunit/junit/AnalyzeClasses.java
+++ b/archunit-junit/junit4/src/main/java/com/tngtech/archunit/junit/AnalyzeClasses.java
@@ -63,6 +63,12 @@ public @interface AnalyzeClasses {
     Class<? extends LocationProvider>[] locations() default {};
 
     /**
+     * @return Whether to import all classes on the classpath.
+     *         Warning: Without any further filtering this can require a lot of resources.
+     */
+    boolean wholeClasspath() default false;
+
+    /**
      * Allows to filter the class import. The supplied types will be instantiated and used to create the
      * {@link ImportOptions} passed to the {@link ClassFileImporter}. Considering caching, compare the notes on
      * {@link ImportOption}.
--- a/archunit-junit/src/test/java/com/tngtech/archunit/junit/ClassCacheTest.java
+++ b/archunit-junit/src/test/java/com/tngtech/archunit/junit/ClassCacheTest.java
@@ -6,6 +6,7 @@ import java.util.Set;
 
 import com.tngtech.archunit.core.domain.JavaClass;
 import com.tngtech.archunit.core.domain.JavaClasses;
+import com.tngtech.archunit.core.importer.ClassFileImporter;
 import com.tngtech.archunit.core.importer.ImportOption;
 import com.tngtech.archunit.core.importer.ImportOptions;
 import com.tngtech.archunit.core.importer.Location;
@@ -15,6 +16,7 @@ import com.tngtech.archunit.testutil.ArchConfigurationRule;
 import com.tngtech.java.junit.dataprovider.DataProvider;
 import com.tngtech.java.junit.dataprovider.DataProviderRunner;
 import com.tngtech.java.junit.dataprovider.UseDataProvider;
+import org.assertj.core.api.Condition;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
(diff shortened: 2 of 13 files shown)


codecholeric:
:warning: As the second part of the description explains, this will still be a breaking change for some users. Thus, we'll release it with `1.0` and we need to make sure to document it carefully as breaking change.

archunit-junit/junit4/src/main/java/com/tngtech/archunit/junit/AnalyzeClasses.java
```diff
@@ -62,6 +62,12 @@
      */
     Class<? extends LocationProvider>[] locations() default {};
 
+    /**
+     * @return Whether to import all classes on the classpath.
```
t-h-e, line 66: Maybe `scan` or `analyze` instead of `import`, to better distinguish the wording from the `ImportOptions`?

t-h-e (approved):
Looks good 👍 
Just one nit-picking comment.
Maybe instead of `wholeClasspath` -> `scanWholeClasspath` or `analyzeWholeClasspath`?

archunit-junit/src/test/java/com/tngtech/archunit/junit/ClassCacheTest.java
```diff
@@ -134,10 +136,38 @@ public void rejects_LocationProviders_without_default_constructor() {
...
+        assertThatTypes(classes).contain(getClass(), TestAnalysisRequest.class);
+        assertThatTypes(classes).doNotContain(ClassFileImporter.class);
+    }
+
+    @Test
+    public void if_whole_classpath_is_set_true_then_the_whole_classpath_is_imported() {
+        TestAnalysisRequest defaultOptions = new TestAnalysisRequest().withWholeClasspath(true);
+        Class<?>[] expectedImportResult = new Class[]{getClass()};
+        doReturn(new ClassFileImporter().importClasses(expectedImportResult))
+                .when(cacheClassFileImporter).importClasses(ArgumentMatchers.<ImportOptions>any(), ArgumentMatchers.<Collection<Location>>any());
```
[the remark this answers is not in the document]
codecholeric: Yeah, good old auto-complete gave me this one :joy: How can IntelliJ be soooo "smart", but then not see that the intelligent way would be to suggest the version that is consistent :stuck_out_tongue_closed_eyes:

archunit-junit/junit4/src/main/java/com/tngtech/archunit/junit/AnalyzeClasses.java
codecholeric: What about `Whether to look for classes on the whole classpath`? I just checked the other Javadocs and that seems consistent to `packages()`, etc...

codecholeric:
Thanks a lot for the review :smiley: About the naming of the method, I thought it feels consistent with the other methods :thinking: Because we also don't call it `scanPackages()`, but just `packages()`. The mental model was more "analyze classes -> packages", so here I thought "analyze classes -> whole classpath". Do you think users would have a problem to understand this in the context of `@AnalyzeClasses(..)`?

archunit-junit/junit4/src/main/java/com/tngtech/archunit/junit/AnalyzeClasses.java
t-h-e: Sounds good

t-h-e:
In the context of `@AnalyzeClasses` it is reasonable and should be understood.
In case of `ClassAnalysisRequest` all methods start with a verb (all with `get` for obvious reasons). I'm not a fan of `isWholeClasspath`, so therefore the suggestion `scanWholeClasspath`. But this is just personal preference, so do not hesitate to merge 😃

codecholeric:
> In the context of `@AnalyzeClasses` it is reasonable and should be understood.
> In case of `ClassAnalysisRequest` all methods start with a verb (all with `get` for obvious reasons)

Got it :slightly_smiling_face: I'll think about it :+1:

PR #833, opened 2022-03-12
Drop Java 7 support

marknp:
This will make Java 8 the minimum required Java version to run ArchUnit tests. We also refactored many internals to use Java 8 streams and lambdas instead of anonymous classes and replaced ArchUnit's own `Function`, `Predicate`, `Optional` and `Supplier` by the standard JDK ones. Furthermore, we upgraded most dependencies that were only hold back to be Java 7 compatible (like Guava and AssertJ).

Note that this is obviously a breaking change which will make ArchUnit unusable in Java 7 projects and will break all user code that depend on the mentioned ArchUnit types that were replaced by standard types. Thus, this change will only be released with version `1.0.0`.


marknp:
there are two commits to update mockito - should we not join this to one?

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
```diff
@@ -1575,16 +1576,17 @@ private Superclass(JavaType type) {
...
         }
 
         Optional<JavaClass> getRaw() {
             return rawType;
         }
 
+        @SuppressWarnings({"unchecked", "rawtypes"})
         Optional<JavaType> get() {
-            return type.or(rawType);
+            return type.isPresent() ? type : (Optional) rawType;
```
marknp, line 1574: cant we do `type.orElse(rawType)` here?

codecholeric:
Oh, those two Mockito-commits are some artifact from rebasing :see_no_evil: The former one should be squashed into the Guava-upgrade, because it has nothing to do with Mockito :grimacing: Good catch! :+1:

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
codecholeric: The problem is, that we want to combine 2 `Optionals` and that doesn't seem to be possible in a convenient way :unamused: 

Starting with Java 9 there would be `optional.or(() -> otherOptional)`, but with Java 8 we would AFAIS have to do something ugly like

```
Optional.ofNullable(type.orElse(rawType.orElse(null)))
```

Don't have a strong preference, to me both looks bad :joy: Maybe I'll use this ugly piece of code though, just so we can get rid of the `@SuppressWarnings`, because that also is something we should really avoid :+1:

marknp: I think "just so we can get rid of the @SuppressWarnings, because that also is something we should really avoid" is a totally reasonable way of thinking :D

PR #835, opened 2022-03-17
Add/Improve Javadoc for annotatedWith

rweisleder:
This commit also incorporates the changes from #540.


codecholeric:
Thanks a lot for the improvement :smile: I changed two little things:

1) add a `@param annotationTypeName` description additionally to `@see methodWithClass(Class)` that explains that this is the fully qualified class name (because I have seen users struggle with this and using the simple class name)
2) for the negated explanation for meta-annotated-with I inverted the part in `<p>`, because I thought it might be misunderstood. E.g.
`The assertion is also successful if members are not directly annotated with the supplied annotation type. ->
The assertion also fails if members are directly annotated with the supplied annotation type.` Because the former sounds like it might already be a sufficient condition to be not directly annotated with the annotation type, while in fact it has to be neither annotated directly **nor** meta-annotated.

Anyway, I hope that this additional documentation will help users, in particular with the not obvious part about meta-annotated also matching direct annotations :+1:

rweisleder:
LGTM, thanks!

Issue #855, opened 2022-04-27
Support creation of arbitrary ArchConditions from Predicates


u3r:
Hi everyone,

internally there are already `DoesConditionByPredicate, IsConditionByPredicate, HaveConditionByPredicate`.
What is missing is access to them, e.g. in the form of
```java
// ArchCondition.java
public <T extends HasDescription & HasSourceCodeLocation> ArchCondition<T> have( DescribedPredicate<T> predicate ) {
    return new HaveConditionByPredicate(predicate);
}

public <T extends HasDescription & HasSourceCodeLocation> ArchCondition<T> is( DescribedPredicate<T> predicate ) {
    return new IsConditionByPredicate(predicate);
}

public <T extends HasDescription & HasSourceCodeLocation> ArchCondition<T> does( DescribedPredicate<T> predicate ) {
    return new DoesConditionByPredicate(predicate);
 }
```
as well as the respective shortcuts in
`ClassesShould, ClassesShouldConjuntion, MethodsShould, MethodsShouldConjunction ....`
e.g. 
```java
/*IsConditionByPredicate*/ ClassesShould.be(DescribedPredicate<T> predicate)
/*HaveConditionByPredicate*/ ClassesShould.have(DescribedPredicate<T> predicate)
```

this would make writing own Archrules (by using mainly predicates) so much easier and more general.

Issue #437, opened 2020-09-27
Spotless complains about line endings on Windows with core.autocrlf=false


rweisleder:
Extracted this issue from PR #427, see also the discussion on that PR...

The ArchUnit build `gradlew build` fails currently on my Windows machine:
```
Execution failed for task ':archunit:spotlessJavaCheck'.
> The following files had format violations:
      archunit\src\jdk9main\java\com\tngtech\archunit\core\domain\Java9DomainPlugin.java
          @@ -1,36 +1,36 @@
          -/*\n
          - * Copyright 2014-2020 TNG Technology Consulting GmbH\n
          - *\n
...
          +/*\r\n
          + * Copyright 2014-2020 TNG Technology Consulting GmbH\r\n
          + *\r\n
...
      ... (24 more lines that didn't fit)
  Violations also present in 662 other files.
  Run 'gradlew.bat :archunit:spotlessApply' to fix these violations.
```
Spotless wants to change the line endings of all files in my working copy from \n to \r\n.

My Git is configured to `core.autocrlf=false`, so the line endings in my working copy are \n.

Creating a `.gitattributes` file with the content
```
*.java text eol=lf
```
worked for me, but failed on GitHub's Windows runner, see https://github.com/TNG/ArchUnit/runs/1166974355
```
Execution failed for task ':archunit-junit4:spotlessJavaCheck'.
> The following files had format violations:
      archunit-junit\junit4\src\main\java\com\tngtech\archunit\junit\AnalyzeClasses.java
          @@ -1,18 +1,18 @@
          -/*\r\n
          - * Copyright 2014-2020 TNG Technology Consulting GmbH\r\n
          - *\r\n
...
          +/*\n
          + * Copyright 2014-2020 TNG Technology Consulting GmbH\n
          + *\n
...
```
In this case, Spotless wants to change the line endings from \r\n to \n.

Spotless has an issue for this topic which is not resolved currently: https://github.com/diffplug/spotless/issues/540

stefanroeck:
I ran into the same issue, `.gitattributes` did the trick as mentioned above.
After that, the JavaClassDescriptorTest fails due to an encoding issue which can be fixed by adding these lines to `build.gradle`:
``` 
    compileJava.options.encoding = "UTF-8"
    compileTestJava.options.encoding = "UTF-8"
```

rweisleder:
[...]

PR #858, opened 2022-04-29
Feature/make arbitrary arch conditions from predicates

u3r:
This is a partial fix for  #855.

I've added two of the methods with some tests.
A couple of points though:

a) I had severe problems, getting this to build (see #437) and test (in IntelliJ lots of tests are executed even if only a few are selected). So formatting might be off

b) I have adapted one architecture check - my reasoning is that just converting from a `DescribedPredicate<T>` to a `ArchCondition<T>` is ok without contravariance. It makes stuff easier to diagnose and doesn't loose any power as ArchConditions are already handled contravariantly.

c) I have also added an 'inverse' test but I cannot get that to fail - beats me why.

I'm looking forward to your feedback!

### DIFF
--- a/archunit-integration-test/src/test/java/com/tngtech/archunit/PublicAPIRules.java
+++ b/archunit-integration-test/src/test/java/com/tngtech/archunit/PublicAPIRules.java
@@ -4,7 +4,9 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.lang.reflect.WildcardType;
+import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import com.tngtech.archunit.base.DescribedPredicate;
@@ -15,6 +17,7 @@ import com.tngtech.archunit.core.domain.JavaConstructor;
 import com.tngtech.archunit.core.domain.JavaMember;
 import com.tngtech.archunit.core.domain.JavaMethod;
 import com.tngtech.archunit.core.domain.properties.HasName;
+import com.tngtech.archunit.core.domain.properties.HasReturnType;
 import com.tngtech.archunit.junit.ArchTest;
 import com.tngtech.archunit.junit.ArchUnitRunner;
 import com.tngtech.archunit.lang.ArchCondition;
@@ -22,6 +25,7 @@ import com.tngtech.archunit.lang.ArchRule;
 import com.tngtech.archunit.lang.CompositeArchRule;
 import com.tngtech.archunit.lang.ConditionEvents;
 import com.tngtech.archunit.lang.SimpleConditionEvent;
+import com.tngtech.archunit.lang.conditions.ArchConditions;
 import com.tngtech.archunit.lang.syntax.ArchRuleDefinition;
 
 import static com.google.common.collect.Iterables.getLast;
@@ -120,6 +124,22 @@ public class PublicAPIRules {
                     .and().arePublic()
                     .and().doNotHaveName("adhereToPlantUmlDiagram")
                     .should().haveRawParameterTypes(thatArePublic());
+
+    @ArchTest
+    public static final ArchRule methods_converting_from_described_predicate_to_arch_condition_should_not_be_contravariant =
+          codeUnits()
+                  .that().haveRawParameterTypes(anyElementThat(is(assignableTo(DescribedPredicate.class))))
+                  .and(are(declaredIn(modifier(PUBLIC))))
+                  .and(are(not(declaredIn(annotatedWith(Internal.class)))))
+                  .and(are(declaredIn( ArchConditions.class )))
+                  .and(have(modifier(PUBLIC)))
+                  .and(have(HasReturnType.Predicates.rawReturnType(is(assignableTo(ArchCondition.class)))))
+                  .should(ArchConditions.not((haveContravariantPredicateParameterTypes())))
+                  .as(String.format(
+                          "Public API methods that take a %s<PARAM> and convert it to a %s should not declare the type parameter contravariantly (i.e. %s<? super PARAM>)",
+                          DescribedPredicate.class.getSimpleName(), ArchCondition.class.getSimpleName(), DescribedPredicate.class.getSimpleName()))
+                  .because(String.format(
+                          "%ss are already treated contravariantly in the API", ArchCondition.class.getSimpleName()));
 
     @ArchTest
     public static final ArchRule predicate_parameters_of_public_API_code_units_should_be_contravariant =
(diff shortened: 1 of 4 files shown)


u3r:
@codecholeric, do you have time for some feedback on this? Thanks!

codecholeric:
Hey, sorry that it took me so long, work pipeline problems :grimacing: Thanks a lot for your push here! :smiley: 
I had already thought about something like this as well, but I wanted to go a little different direction. The problem for me with these existing `ArchConditions` (and the reason why I kept them private) is that the description of the single violations typically does not match the one description to be added to the rule text.
For example assume `topLevelClasses()` returns a `DescribedPredicate`:

```
classes().should().be(topLevelClasses())
```

The `DescribedPredicate` will naturally have the description `top level classes` so it completes the rule text "be top level classes". However, if we use the description of this predicate for the violations it's just awkward:

```
Class <SomeClass> is not top level classes
```

These two conditions were tailored in a way that they worked internally, but for me they don't have the quality to make it into the public API :thinking: At least I would want a better violation description somehow.

Sometimes the description of the predicate is also fine, but it would at least be necessary to customize it (as these classes allow internally, exactly for cases like the above).

Originally I had thought of something like a more convenient lambda API on `ArchCondition`, so it's similar to `DescribedPredicate.describe(..)` :thinking: 

E.g.

```
ArchCondition.from(topLevelClasses())
  .describeViolatedAs(javaClass -> javaClass.getDescription() + " is no top level class")
```

I.e. have some optional additional methods like `describeViolatedAs(Function<T, String>)` and `describeAllowedAs(Function<T, String>)` to customize the descriptions of the single violations. That would probably mean though, that it couldn't be added to the fluent API :thinking: But I'm not sure if there is a good way that doesn't have the problem I mentioned above :man_shrugging: 

Regarding your problems, do you have exactly the problem mentioned in #437? Because according to https://github.com/TNG/ArchUnit/issues/437#issuecomment-1030610272 it should have been fixed by some upgrade, no? I wonder why it runs on Windows CI without any problem, maybe there the `autocrlf` is set differently.

u3r:
Hi @codecholeric,

thanks for the long answer even on a Sunday ;)
I understand  your reservations - I need to check how I was going to use them - it seems my rules mach better with those methods.
One alternative to think about is to default to singular handling like so: (pseudocode not checked against api)
`anyClass().that(hasSomeProperty()).shouldBe(someThing())` (with `anyClass()` being equivalent to `classes()` in all but description.

If this holds for most examples, it would
a) have nice to read output (your point) and
b) have a single short way to re-use predicates as conditions (my goal)

Regarding the other problems, I need to check again on my work machine.

codecholeric:
The only problem I see about `anyClass()` or similar is that it's quite confusing. At least I as a user would ponder "why is there `classes()` and `anyClass()`, what's the difference", etc. And then we would have `theClass()`, `anyClass()` and `classes()` where the first one is singular and the latter ones behave exactly the same :thinking:

codecholeric:
I would rather go with a way to make the condition creator more customizable. For simple cases the predicate might already be fine, then you don't have to do anything. But for complex cases you should be able to customize it to have correct wording. I also think having it as a factory method on `ArchCondition` would not be so bad, because that's consistent with `DescribedPredicate`.

u3r:
I am still checking in semi-blind, as even rebasing onto latest gives me errors on `./gradlew clean build`  (using JDK17)
(see [buildLog.txt](https://github.com/TNG/ArchUnit/files/8804162/buildLog.txt) )
Is this also supposed to auto-format? If not AND you are by chance using intellij, would you be open to checking in the formatter settings?)

As for the code:
I added some quick-and-dirty version of what I understood you wanted  (not polished or named nicely at all)
a) I've put in the lambda based approach you mentioned
b) I've also put in `withViolationString` that relies on the predicates description so creating this doesn't get too verbose
c) I am not sure where the plural form of the description is generated and how to test that realistically.
This would fail: 
```java
 assertThat (ArchCondition
              .from( JavaClass.Predicates.simpleName( "SimpleName" ) )
              .withViolationString( "does not have" ) )
              .hasDescription( "do not have simple name 'SimpleName'" );
```
but I don't know if it is something the builder should support or that is generated "around" as the ArchCondition is used.
d) The builder out of `ArchCondition.from()` is to test a fluent api (no probles so far in the simple case)
Alternatively it could of course be `ArchCondition.from(DescribedPredicate, FailureTextDescriptor)`; I am unsure what is nicer.
e) I quite liked the simplicity of the `ArchConditions.be/have` versions for their simplicity because at least for me they cover a LOT of cases without the worse readability of the more powerful variants.

codecholeric:
@u3r I would also guess what @hankem suggested about the `gradle.properties`. In particular the one inside of the ArchUnit project adds

```
org.gradle.jvmargs=--add-exports jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED
```

And AFAIS that was what was necessary at the time to make Spotless work unfortunately :disappointed: 

As for your question above, I'm not sure I completely understand. The description of the `ArchCondition` should be derived 1-to-1 from the `DescribedPredicate`. So `.from( JavaClass.Predicates.simpleName( "SimpleName" ) )` would lead to just `simple name "SimpleName"`. The rest should apply to the single violations, i.e. the details of the test when it fails. Like

```
Class <com.example.Foo> does not have simple name "SimpleName"
Class <com.example.Bar> does not have simple name "SimpleName"
...
```

u3r:
I'll take a look at the fails and suggestions but it will take me a while to get back to this - just as a heads up

u3r:
I'm working out the technical problems (thanks to @codecholeric  and @hankem) Indeed I need the global `gradle.properties` for the corporate proxy but adding the jvm args worked.

However I am still no satisfied with the api.
I've not  yet pushed the code for this. It is meant to discuss the actual syntax:
```java
    @ArchTest
   ArchRule demoRule = methods()
         .should( never(
               ArchCondition.from( HasName.Predicates.nameEndingWith( "s" ).forSubtype() )
                     .withViolationString( "has a" ) // customize single violation text (this one uses the original description)
                     /* This next section is needed, because we have no way to customize the original predicate description "name ending with 's'"
                        This is especially bad as you need to repeat the original description again creating noise during rule definition
                      */
                     .as( "have a name ending with 's'" )
         ) )
         .because( "We don't want that" );

   @ArchTest
   ArchRule demoRule2 = methods()
         .should( never(
               ArchCondition.from( ArchPredicates.have(HasName.Predicates.nameEndingWith( "s" )) ) // no need for (further) description customization
                     //again we need to repeat the original description again (as now it's prefix does not match) creating noise during rule definition
                     .describeViolatedAs( entity -> entity.getDescription() + "has a name ending with 's'" )
         ) )
         .because( "We don't want that" );

   @ArchTest
   ArchRule demoRule3 = methods()
         .should( never(
               ArchCondition.from( HasName.Predicates.nameEndingWith( "s" ).forSubtype() ) // how to document that ArchPredicates.have and friends may not be used
                     .withDescriptionPrefix("have a") // prefix for the predicate description (reusing that)
                     .withViolationPrefix( "has a" ) // prefix for the predicate description (reusing that) in the violation case (singular)
         ) )
         .because( "We don't want that" );

   @ArchTest
   ArchRule demoRule4 = methods()
         .should( never(
               ArchConditions.have( HasName.Predicates.nameEndingWith( "s" ).forSubtype() ) // how to document that ArchPredicates.have and friends may not be used
         ) )
         .because( "We don't want that" );
```

`demoRule` has a minimal `ArchConditionBuilder` but is not nice to handle
`demoRule2` just 'inverts' the badness of `demoRule`
`demoRule3` would be ok from a usage point of view - still very verbose in comparison to `demoRule4` but the methods could be added internally using `demoRule3`
`demoRule4` was my initial proposal (which I personally still like best due to it's brevity while handeling 70% of (my) use-cases with just `have()` and `be()` methods.

Before I start spending time with cleanup, naming and such please have a look and let's agree on
what approach to take and maybe what documentation to add, so a user can more easily navigate the pitfalls shown in (`demoRule 1 + 2`)

My personal favourite would be to implement something that supports `demoRule3`-like syntax AND the `ArchPredicates.have() and .be()` methods using that API to provide a 70% case with just one line of code (instead of 3)

codecholeric:
I'm against adding something to the official API that works nicely in 70% of the cases and badly otherwise. I think it's fine doing this as a user, but I'm not convinced of adding such a thing to the core for all to use. I can see the argument about the bloat, but for me that doesn't trump that an API should work 100% of the cases :man_shrugging: 

My ideas would be

a) instead of `ArchCondition.from(..)` let's use `ArchCondition.have(..)` and `ArchCondition.be(..)` as suggested by you, but make it customizable if necessary. I.e. allow optionally to define something like 
```
ArchCondition.be(INTERFACES)
  .describeEventAs((clazz, isSatisfied) -> 
    clazz.getDescription() + (satisfied ? " is " : " is no ") + "interface")
```
Then the case where it just works would simply be `should(ArchCondition.have(nameEndingWith(..)))`, cause you could omit the customization, but you can write correct event messages in all cases by customization.

b) To the fluent API I would not add `should().have/be(..)`, because there it is not customizable and only works in 70% of the cases. What I could imagine (if we really want to offer something like this right in the fluent API?) would be a more generic form like `should().satisfy(predicate)`. Which would then cause `Class <..> satisfies <simple name ending with 'Foo'>` or something like that. Because at least then it's a grammatically correct and meaningful sentence, even though it doesn't read as "natural language" as the usual ArchUnit rules.

codecholeric:
To move forward in smaller steps I have taken your PR, extracted a part from it and fixed it up a little here: #904
Because for me that part is non-controversial, that we want some convenient factory method to create conditions from predicates. After that we can decide how to move on with this PR here, i.e. should we transform it into something that adds fluent API or more convenience methods on top, or is it alright as it is :thinking: 
@u3r What do you think about #904? Does it look reasonable to you in its current state? I would be happy about a review :slightly_smiling_face:

u3r:
converted to #904, no more use for this

PR #862, opened 2022-05-06
add JUnit5-support for `@ArchTest` members in abstract base classes

### LINKED ISSUE #104: @ArchTest instance fields or static methods in abstract base classes

hankem:
So far, having `@ArchTest` fields or methods in an abstract base class would cause the JUnit 5 engine to crash. This also made the pattern to "share" tests via extending a common base class impossible, which some users would like to have.

Similar to #105 for JUnit 4, resolves #104.

### DIFF
--- a/archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/ArchUnitTestDescriptor.java
+++ b/archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/ArchUnitTestDescriptor.java
@@ -15,7 +15,9 @@
  */
 package com.tngtech.archunit.junit.internal;
 
+import java.lang.reflect.AccessibleObject;
 import java.lang.reflect.Field;
+import java.lang.reflect.Member;
 import java.lang.reflect.Method;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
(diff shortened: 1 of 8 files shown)


hankem:
Thanks; I think it's a great idea to pass the `owner` in first place!

Two comments on the Review commit: 
1. The change to `AbstractBaseClassWithLibraryWithAbstractBaseClass` belongs to the fixup! commit, doesn't it?
2. I'm not sure whether I understand the last part of the commit message correctly:
    > since instanceof is very compiler friendly(TM) ;-)

    Did you want to say that `instanceof` is _not human friendly?_

codecholeric:
Actually I decided to squash all commits together instead of fixing up each single commit with parts from the review commit :grimacing: In the end it's one consistent change anyway, and it looks good to me as a single commit.

PR #876, opened 2022-06-03
Improve `ArchCondition` evaluation

codecholeric:
At the moment evaluating `ArchConditions` is quite memory inefficient. E.g. for a rule like `noClasses().should().accessField(Foo.class, "bar")` we create a `ConditionEvent` for every single field access and qualify it with "allowed" or "violation". Thus, this is independent of some access actually being a violation but done for every access. Furthermore, all such events of all evaluated objects will be kept in memory until the test failure is reported. This will keep in the former case events for all field accesses of all classes in memory until the final failure (or success) is reported.

The background for this behavior is the complex evaluation logic of nested and inverted conditions. E.g. `noClasses()` will invert the events and thus make the events that were "allowed" before to be violations. To make matters worse we have combinations like `accessField(..).orShould()...` and aggregations like `noClasses().should().haveAny...` where `haveAny...` will be satisfied if there is any occurrence that supports the condition, but by negating it via `noClasses()` we effectively change the evaluation to "all occurrences must NOT satisfy the condition". So it is quite challenging to determine at which point the information about "allowed" events is finally obsolete, since we can arbitrarily aggregate and invert.

I tried to come up with a lazy solution, where we would use some `Supplier/Consumer` approach, but failed to come up with one that a) makes up a good public API for users, b) works for all those cases of aggregation including customized behavior by users and c) really saves substantial amounts of memory. So in the end I gave up on that and focused on optimizing the current approach.

This PR will address two points: a) improve the memory footprint of `ArchCondition` evaluation and b) improve the public API to make future optimizations easier and the code more maintainable (now is a very good opportunity since we are about to release a new major version, so we can break the public API a little bit).

## Performance Aspect

As for point a) the main optimization here is to throw out intermediate copying of events, and once we know we are done with event evaluation (i.e. we are now top-level about to create the result) to dump all references to now obsolete satisfied events. This had the following positive effect when testing `noClasses().should().accessField(System.class, "out")` on a big code base on my machine:

#### Heap allocation

image

#### Performance

Evaluating the rule could be done 89 times in 60 seconds.

#### Heap allocation

image

#### Performance

Evaluating the rule could be done 135 times in 60 seconds.

### Result

So altogether we see a smoother memory allocation with less GCs and a performance improvement of about 50% when evaluating such conditions. Thus, I consider the change worthwile.

## API Aspect

Before the public API included a final class `ConditionEvents`, this is now an interface to allow different implementations to be used internally. Also the API of this interface has been stripped of all non-essential methods, like `getAllowed()` which would return the allowed events and force us to always keep a reference to satisfied events as long as we hold `ConditionEvents`. I tried to strip the API down to really just the essentials, e.g. `containViolation()` or `getViolating()` and remove all methods that don't make sense in all contexts.


archunit/src/test/java/com/tngtech/archunit/lang/conditions/FieldAccessConditionTest.java
```diff
@@ -51,36 +51,33 @@ public void FieldAccessCondition_satisfied_on_both_get_and_set_field() {
...
     }
 
-    private void assertSatisfiedWithMessage(FieldAccessCondition getFieldCondition, JavaFieldAccess access,
-            String accessDescription) {
+    private void assertSatisfied(FieldAccessCondition getFieldCondition, JavaFieldAccess access) {
         ConditionEvents events = checkCondition(getFieldCondition, access);
-        boolean satisfied = !events.containViolation();
-
-        assertThat(satisfied).as("Events are satisfied").isTrue();
+        assertThat(!events.containViolation()).as("Events are satisfied").isTrue();
```
hankem, line 64: I'd get rid of the negation and stick to the API, which focusses on violations:
```suggestion
        assertThat(events.containViolation()).as("Events are violated").isFalse();
```

archunit/src/test/java/com/tngtech/archunit/lang/conditions/FieldAccessConditionTest.java
```diff
@@ -51,36 +51,33 @@ public void FieldAccessCondition_satisfied_on_both_get_and_set_field() {
...
+            FieldAccessCondition getFieldCondition,
+            JavaFieldAccess access,
+            String accessDescription
+    ) {
         ConditionEvents events = checkCondition(getFieldCondition, access);
-        boolean satisfied = !events.containViolation();
 
-        assertThat(satisfied).as("Events are satisfied").isFalse();
-        assertThat(events.getAllowed()).isEmpty();
+        assertThat(!events.containViolation()).as("Events are satisfied").isFalse();
```
hankem, line 75:
```suggestion
        assertThat(events.containViolation()).as("Events are violated").isTrue();
```

archunit/src/main/java/com/tngtech/archunit/lang/conditions/JoinCondition.java
```diff
@@ -0,0 +1,139 @@
...
+        List<ConditionWithEvents<T>> invert(List<ConditionWithEvents<T>> evaluatedConditions) {
+            return evaluatedConditions.stream().map(this::invert).collect(toList());
+        }
+
+        private ConditionWithEvents<T> invert(ConditionWithEvents<T> evaluation) {
+            SubEvents invertedEvents = new SubEvents();
+            Stream.concat(
+                    evaluation.events.getAllowed().stream(),
+                    evaluation.events.getViolating().stream()
+            ).forEach(event -> invertedEvents.add(event.invert()));
```
hankem, line 135: This could also be written as:
```suggestion
            ).map(ConditionEvent::invert).forEach(invertedEvents::add);            
```
but after a second thought, I'm not even convined that this is better any more, so just pick what you prefer.

archunit/src/main/java/com/tngtech/archunit/lang/conditions/SubEvents.java
```diff
@@ -0,0 +1,71 @@
...
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+
+import com.tngtech.archunit.lang.ConditionEvent;
+import com.tngtech.archunit.lang.ConditionEvents;
+
+final class SubEvents implements ConditionEvents {
```
hankem, line 26: Some thoughts about naming:

Do you think that `SubEvents` is a good name? (I didn't understand it.)
How about `InvertableConditionEvents` (as this class tracks `allowedEvents`, too)?

Similarly, I'd probably rename `ConditionEventsInternal` to something like `SimpleConditionEvents` or `LeanConditionEvents` and leave a comment that this class only tracks violations.

I was at first surprised that `ConditionEvents.Factory.create()` builds lean `ConditionEvents`, but then realized that this was okay as the `interface ConditionEvents` has changed and now really only deals with violations. So `ConditionEvents` might not be the best name any more: IMO it's rather a `ViolationsContainer` now, but I guess that you didn't want to break _all_ existing `ArchCondition`s (which have this type in the signature of the `check` method)... 😉

hankem (approved):

archunit/src/main/java/com/tngtech/archunit/lang/EvaluationResult.java
```diff
@@ -15,71 +15,123 @@
...
-        for (ConditionEvent event : part.events) {
-            events.add(event);
-        }
+        violations.addAll(part.violations);
     }
 
     /**
-     * @see ConditionEvents#handleViolations(ViolationHandler)
+     * Passes violations to the supplied {@link ViolationHandler}. The passed violations will automatically
+     * be filtered by the type of the given {@link ViolationHandler}. That is, if a
```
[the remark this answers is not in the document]
codecholeric: Yeah, sounds better :+1:

archunit/src/main/java/com/tngtech/archunit/lang/EvaluationResult.java
```diff
@@ -15,71 +15,123 @@
...
+     * <code>SomeClass</code> will be reported. Note that this will be unsafe for generics, i.e. ArchUnit
+     * cannot filter to match the full generic type signature. E.g.
+     * <pre><code>
+     * handleViolations((Collection&lt;Optional&lt;String&gt;&gt; objects, String message) ->
+     *     assertType(objects.iterator().next().get(), String.class)
+     * )
+     * </code></pre>
+     * might throw an exception if there are also {@code Optional<Integer>} violations. Thus, when using
+     * this method generic type parameters should always be substituted by wildcard types,
+     * except if it is clear that the type parameter will always be set to a well known type.
```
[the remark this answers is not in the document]
codecholeric: It is more of a "unless you know what you are doing" warning :thinking: Take a rule where you know the violations will always be `ThrowsClause` of `JavaMethod`. Then you could safely do 

```
handleViolations((ThrowsClause<JavaMethod> violatingObject, String message) -> ...)
```

But if you then catch a `ThrowsClause<JavaConstructor>` for any reason you will get an exception :man_shrugging: 
I'll try to rephrase it a little.

archunit/src/main/java/com/tngtech/archunit/lang/EvaluationResult.java
```diff
@@ -96,8 +148,8 @@ public Priority getPriority() {
      */
     @PublicAPI(usage = ACCESS)
     public EvaluationResult filterDescriptionsMatching(Predicate<String> linePredicate) {
-        ConditionEvents filtered = new ConditionEvents();
-        for (ConditionEvent event : events) {
+        ConditionEvents filtered = ConditionEvents.Factory.create();
+        for (ConditionEvent event : violations) {
             filtered.add(new FilteredEvent(event, linePredicate));
         }
         return new EvaluationResult(rule, filtered, priority);
```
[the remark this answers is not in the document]
codecholeric: Given that I forgot to transport the `informationAboutNumberOfViolations` I think this is a very good idea :see_no_evil: I adjusted it... But I did not make it public for now.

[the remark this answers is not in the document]
hankem: I saw that `informationAboutNumberOfViolations` is not set, but we anyways don't know what to set for the filtered `EvaluationResult`, do we?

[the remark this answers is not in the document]
codecholeric: Yes, I think that's a good idea. It also makes it more explicit that we drop the `informationAboutNumberOfViolations` here (which makes sense I think, since we filter out any number of possible violations, so even if there was an additional hint it's likely not accurate anymore). But I'll keep it out of the public API for now...

archunit/src/test/java/com/tngtech/archunit/lang/conditions/FieldAccessConditionTest.java
codecholeric: Absolutely :joy:

archunit/src/main/java/com/tngtech/archunit/lang/conditions/JoinCondition.java
codecholeric: Not much difference to me either :man_shrugging: So I'll just keep it...

archunit/src/main/java/com/tngtech/archunit/lang/conditions/SubEvents.java
codecholeric: But it's not completely true that it is always only a `ViolationsContainer` as long as you can add `ConditionEvents` with `satisfied == true`, no? I mean, you can always add satisfied and violated `ConditionEvents`, so I still think `ConditionEvents` is a fair name. What the interface now does not allow anymore is to *retrieve* informations about satisfied events. But it can still transport them. Nevertheless, since we can't *retrieve* satisfied `ConditionEvents` or are influenced in any way on the public surface, we don't need to track allowed events anymore in the standard case :man_shrugging: But I could imagine users doing the same as ArchUnit does internally, i.e. use their own implementation to track or handle more :man_shrugging:

codecholeric: About the other naming, like `SubEvents`, I have to admit that I haven't spent too much brain power on the naming as it was never public API :see_no_evil: For me the name came from "I'm collecting a subset of events with this container, then transform it and add it to the parent container". But earlier it was local variables that I called `subEvents`, not a whole type...

codecholeric: About `Lean` I'm not a super big fan, because that for me also doesn't convey much information :man_shrugging: Or in other words, it's quite fuzzy to me what this means...

codecholeric: The suffix `Internal` came from the convention we have for the syntax. I.e. public API is an interface, like `ConditionEvents.Factory.create()` in this case, and the implementation was usually called `xxxInternal` there. But yeah, maybe `SimpleConditionEvents` is fair, since it's the most straight forward implementation within ArchUnit without any extra properties / responsibilities.

hankem: Right, the naming does not surface the public API.
I'd still find it useful to have a clear semantics for ArchUnit's internal types, but a code comment or JavaDoc would be enough.

hankem: The difference to the syntax API with the `Internal` classes is that there is only one implementation.
Here we have two, so an ArchUnit developer needs to understand their semantics.

codecholeric: Unfortunately, `InvertibleConditionEvents` also sounds weird to me, because every `ConditionEvent` has an `invert()` method and is thus "invertible" :cry: But it's really hard to come up with a good name. Maybe in the end it will just be `Extended...` or something fuzzy like that :disappointed: But at least it's no API, so the name is irrelevant for users...

codecholeric: As discussed I'll rename `SubEvents` to `ViolatedAndSatisfiedConditionEvents` with an additional Javadoc. It's not the prettiest name, but at least it tells exactly what's the point :wink: And it's luckily no public API in the end :man_shrugging:

Issue #641, opened 2021-07-20
Add ability to execute a subset of rules from the via system properties


bobtiernay-okta:
When developing large applications, one common pattern is using  multiple `ArchRules` in a single class file. However, it appears as though it is not possible to only run a subset of these tests from the command line. It would be great if new system properties were exposed that allowed JUnit 4 / 5 runners to only execute rules by pattern, in similar fashion to how JUnit does for test classes.

crizzis:
I'm not sure I follow. What is wrong with the existing tooling support? Is it not possible to exclude test classes by category / file name?

Also, what's the point of grouping ArchUnit suites under one uber-suite if you're going to then exclude some of them anyway? How was the grouping helpful in the first place, then?

codecholeric:
One common way I have used several times is to group rules together and include them hierarchically, like

```
@AnalyzeClasses(..)
class ProjectOneArchRules {
    @ArchTest
    static final ArchTests serviceRules = ArchTests.in(ServiceRules.class);

    @ArchTest
    static final ArchTests controllerRules = ArchTests.in(ControllerRules.class);
}
```

This works completely fine in CI where I want to continuously execute that whole suite. But now let's say I'm confused why controller rule 5 is failing suddenly (let's assume the message is not self-explanatory this time :wink:). So now I want to execute only controller rule 5 locally to test it real quick. At the moment this is really tedious to achieve. That's what this issue is about.

crizzis:
@codecholeric Please have a look at [my draft PR](https://github.com/TNG/ArchUnit/pull/834/files) when you have a moment. 

The PR:
* introduces support for JUnit 4's `Assume` API
* introduces support for Jupiter's `@EnabledIfXxx` annotations (for method-flavored tests) and a set of bridging `@Conditions.EnabledIfXxx` annotations for field-based tests
* introduces support for Jupiter's `@Tag` annotation for method-flavored tests
* introduces support for Jupiter's `@Disabled` for method-flavored tests
* enables [Surefire test filtering](https://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html) for field-flavored tests (for JUnit Jupiter and JUnit Vintage, but **not** JUnit 4); note that modifications to Surefire are also needed - see [related Jira issue](https://issues.apache.org/jira/browse/SUREFIRE-2037)
[...]

PR #880, opened 2022-06-10
#641 Added tooling tests

crizzis:


archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/ArchUnitTestDescriptor.java
```diff
@@ -56,7 +56,7 @@ class ArchUnitTestDescriptor extends AbstractArchUnitTestDescriptor implements C
     private ClassCache classCache;
 
     private ArchUnitTestDescriptor(ElementResolver resolver, Class<?> testClass, ClassCache classCache) {
-        super(resolver.getUniqueId(), testClass.getSimpleName(), ClassSource.from(testClass), testClass);
+        super(resolver.getUniqueId(), testClass.getName(), ClassSource.from(testClass), testClass);
```
crizzis, line 59: Changed this because Jupiter would group field-flavored and method-flavored test results under different parent nodes (one under FQ class name and the other under simple class name)

archunit-tooling-test/build.gradle
```diff
@@ -0,0 +1,144 @@
...
+def deleteSafely(File file) {
+    if (file.exists()) {
+        println "Deleting ${file.toURI().toString()}"
+        delete file
+    } else {
+        println "Could not find ${file.toURI().toString()}"
+    }
+}
+
+task removeFromMavenLocal {
```
crizzis, line 103: This task cleans the local Maven repo to make sure we are running the Surefire and Gradle tests with an up-to-date version of ArchUnit jars

archunit-tooling-test/build.gradle
```diff
@@ -0,0 +1,144 @@
...
+    dependsOn ':archunit-junit5-engine-api:publishMavenJavaPublicationToMavenLocal'
+    dependsOn ':archunit-junit5-engine:publishMavenJavaPublicationToMavenLocal'
+    dependsOn ':archunit-junit5-api:publishMavenJavaPublicationToMavenLocal'
+    dependsOn ':archunit-junit5:publishMavenJavaPublicationToMavenLocal'
+    dependsOn ':archunit-junit4:publishMavenJavaPublicationToMavenLocal'
+    dependsOn ':archunit:publishMavenJavaPublicationToMavenLocal'
+}
+
+gradle.taskGraph.whenReady { graph ->
+    if (graph.hasTask(publishJUnitDepsToMavenLocal)) {
```
crizzis, line 131: Needed because all the `publishMavenJavaPublicationToMavenLocal` tasks depend on `test` in their respective subprojects, so basically, `publishJUnitDepsToMavenLocal` would re-trigger the tests for all of those subprojects that we need in the local Maven repo.

Perhaps there's a better way of doing the same, didn't really investigate

archunit-tooling-test/src/main/java/com/tngtech/archunit/tooling/engines/gradle/GradleProjectLayout.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    protected String getCompiledTestClassesTargetDirectory() {
+        return "build/classes/java/test";
+    }
+
+    @Override
+    protected String getBuildToolWrapperDirectory() {
+        return "gradle";
```
crizzis, line 25: The Gradle wrapper is not really used for now (note the explicit Gradle version specified inside `GradleEngine`)

archunit-tooling-test/src/test/java/com/tngtech/archunit/tooling/BaselineTest.java
```diff
@@ -0,0 +1,47 @@
...
+import com.tngtech.archunit.tooling.examples.RegularJUnit4Test;
+import com.tngtech.archunit.tooling.examples.RegularJUnit5Test;
+import org.junitpioneer.jupiter.cartesian.ArgumentSets;
+import org.junitpioneer.jupiter.cartesian.CartesianTest;
+import org.junitpioneer.jupiter.cartesian.CartesianTest.MethodFactory;
+
+import java.util.Map;
+import java.util.stream.Stream;
+
+public class BaselineTest extends BaseTest {
```
crizzis, line 12: This test isn't testing ArchUnit as such, it is only here to test the different test engines themselves

build.gradle
```diff
@@ -55,21 +55,37 @@ ext {
...
             junit_dataprovider  : [group: 'com.tngtech.java', name: 'junit-dataprovider', version: '1.11.0'],
             mockito             : [group: 'org.mockito', name: 'mockito-core', version: '4.4.0'],
             assertj             : [group: 'org.assertj', name: 'assertj-core', version: '3.22.0'],
             assertj_guava       : [group: 'org.assertj', name: 'assertj-guava', version: '3.4.0'],
 
+            findBugsAnnotations : [group: 'com.google.code.findbugs', name: 'annotations', version: '3.0.1'], // SpotBugs version 4 requires Java 9+
+
+            // Dependencies for tool integrations
+            surefireApi         : [group: 'org.apache.maven.surefire', name: 'surefire-api', version: '3.0.0-M7'],
+            gradleApi           : [group: 'dev.gradleplugins', name: 'gradle-api', version: '7.4.1'],
```
crizzis, line 74: `gradleApi` and `byteBuddy` are not yet used anywhere (they will be used for Gradle filtering)

codecholeric:
Hey, thanks for all the hard work, but can you please put in a detailed description of **why** (I just see *no description provided*)? My problem here is: This adds 2.5K lines of code that we will have to maintain in the future. And I don't exactly see the benefit yet. In particular, what would be important to me is to weigh this against the simple system property filter suggested in the issue and then give detailed reasons why not to go the simple way.
Because I think we could add a simple filtering based on system properties that would solve the issue creators problem with a couple of hundred LOC for prod and test code altogether. And it would work with any tool, be it Gradle, Maven or IntelliJ, because system properties can always be supplied. So if we would go this way here, where we need this massive complexity of test infrastructure, then we need to state a good reason for it IMHO. So can you outline this? What is your vision? Why do we need these tooling tests for it? What would we gain over the simple filtering mechanism?

codecholeric:
I implemented a simplified version in #1280. Since I have no capacity to work on this further, I'm gonna close this. Feel free to reopen if you want to pick this up or add further improvements.

Issue #887, opened 2022-06-14
ArchUnit: how can one silence references to arrays of basic types for method mayOnlyAccessLayers(...)?


mmoser18:
For a big GWT-based application I am defining an architecture check to prevent that our junior programmers accidentally reference classes that are not GWT-serializable and the resulting objects then can not make it "over the wire" (which then throws exceptions at runtime).

For this I am using a layer-check like so:
```
	@ArchTest
	void checkGWTAccess(JavaClasses classes) {
		LayeredArchitecture layers = Architectures.layeredArchitecture()
			.layer("Client").definedBy("..client..")
			.layer("Shared").definedBy("..shared..")
			... <many entries omitted here> ...
			;

		ArchRule rule = layers
			.whereLayer("Client").mayOnlyAccessLayers("Shared", "DomDtos", "DomBase", "DomConfig",  								
			                                          ... <many entries omitted here> ...
			                                         );
			;
		rule.check(classes);
```

"In principle" this works ok EXCEPT that I am still getting errors for all fields that are arrays of basic types or collections of basic types (i.e. e.g. `int[] digits;`, `char[] buffer;`, `List<Integer> allowedValues;`) as well as methods returning such arrays or collections (i.e. e.g. `int[] methodA(...) {...}`, `String[] methodB(...) {...}`, `List<Character> methodC(...) {...}`, etc.).

These error read e.g.: 
```
Field <com.example.client.ui.components.converter.TimeFieldConverter.timeNumberDigits> has type <[I> in (TimeFieldConverter.java:0)
```

IMHO this is a bug in ArchUnit. I can not fancy any reason arrays of basic types should possibly violate any layer-check rule. 

In case this is NOT a bug: 
How can one exclude these types from a `<layer>.mayOnlyAccessLayers(...)`-check? 

Or - alternatively - can one *include* these types to the list of allowed layers (if it's possible to identify the and/or declare them as "layer")?

The purpose is simply, to avoid getting any rule violations for these references, so that the tests run green.

codecholeric:
Yes, `LayeredArchitecture` is missing an option to specify how to deal with dependencies, similar to `PlantUmlArchCondition`. You can probably find an answer and how to mitigate your problem here though ->  #841
Does that help you?

mmoser18:
Re-reading the referenced append I finally found an expression that helped:

I added the following clause to the layeredArchitecture()-definition:
```
[...]

PR #892, opened 2022-06-18
Add mandatory dependency settings to `layeredArchitecture()`

codecholeric:
Ever since `layeredArchitecture()` supports rules on "forward dependencies" (i.e. `whereLayer(..).mayOnlyAccessLayers(..)`) there has been confusion about how to exclude false positives like `java.lang.Object`. While excluding these dependencies can be achieved by `ignoreDependency(..)` this does not seem to be an API that is easily discoverable.

We now make it mandatory to choose how to treat dependencies right in the beginning. This way users will be guided to those considerations and be made aware of pitfalls as well as have an easy predefined option how to ignore false positives.

This will be a breaking change, since existing `layeredArchitecture()` definitions will not compile anymore without adding the choice of how to treat dependencies.

Resolves: #887


kaebiscs:
Looks good to me. No remarks.

Issue #662, opened 2021-09-02
Feature proposal: Provide "extended" package matching support


Pfoerd:
### Story
As an ArchUnit test developer, I would like to be able to use an extended syntax for package identifiers, since the existing one based on PackageMatcher is very limited. At least, I would like to be able to define alternatives such as `(a|b)`. This allows to support architectures that for some reasons need a namespace for some components by using different packages even though they logically belong to the same architectural entity like a layer, domain, etc. This is especially useful if your code needs to satisfy some "overarching" structural requirements that come from outside your actual project, e.g. from an enterprise architecture.

### Sample Use Case
You have a structure `org.my.app.subdomain1` and `org.my.app.subdomain2` and want to logically place them in the same top-level domain `domain1` in your ArchUnit tests, but for certain reasons you are not allowed to introduce another package level.

### Known Alternatives:
Most ArchUnit APIs we use offer an alternative to the "straightforward" variant of just providing a `packageIdentifier` String e.g. by using custom `Predicate`s oder `SliceAssignment`s. But this makes the tests very verbose, complex and hard to understand.

### API proposals:
* Extend the existing PackageMatcher in a backward compatible way
* Alternatively a new "PackgeMatcherExt" (name WIP 😉) could be introduced, together with corresponding SliceAssignements and Predicates. The advantage of sending something like this upstream is that it could share the basic matching algorithms for the existing syntax with the existing PackageMatcher. This is currently not possible due to its current design (final monolithic class without interfaces). In addition, special API extensions could be introduced for using the "extended" package identifier syntax, e.g. something like `layer("domain").definedByExtendedPkgId("org.my.app.(subdomain1|2)..")` (name WIP 😉).

### Further ideas
Further "nice-to-haves" that could be achieved by such a solutions are non-capturing group alternatives like `(?:a|b)` and a "not" operator like `(!a|b).(?:!a)`

If you're interested, let me know, because we have an existing solution and would be happy to share the relevant parts with you and send it upstream.

P.S.: I really appreciate your work on ArchUnit, it's a really cool tool 😃!

Pfoerd:
I really like your approaches. I wonder if there is a way we can bring them all together, because I think all approaches have their pros and cons:
* add minor syntax extensions to support, for example, simple alternatives like `en.foo.(a|b)` and perhaps `en.(foo.a|bar)`. Since support for the latter can be considered a compatible extension to the former, it could be added incrementally
[...]

Issue #685, opened 2021-10-01
Slices across different base packages


whiskeysierra:
I'd like to cut slices in the following package structure:

- application
  - a
    - http
    - grpc
  - b
    - kafka
    - sns
- domain
  - api
    - a
    - b
  - logic
    - a
    - b
  - spi
    - c
    - d
- infrastructure
  - c
    - memory
    - postgres
  - d
    - file
    - s3

Slices that I'd like to have:

1. `application.a.http`, `application.a.grpc`, `domain.api.a` and `domain.logic.a`
2. `application.b.kafka`, `application.b.sns`, `domain.api.b` and `domain.logic.b`
3. `domain.spi.c`, `infrastructure.c.memory` and `infrastructure.c.postgres`
4. `domain.spi.d`, `infrastructure.d.file` and `infrastructure.d.s3`

If I could specify multiple package identifiers (when calling `slices().matching(packageIdentifier)`), then I would use (slice 1+2):

```
"..application.(*).*.."
"..domain.api.(*).."
"..domain.logic.(*).."
```

or respectively (slice 3+4):

```
"..domain.spi.(*).."
"..infrastructure.(*).*.."
```

Is there a readable way to express that with a single package identifier expression?

Alternatively I'm currently using this (Kotlin) implementation that supports multiple package identifiers:

```kotlin
private fun SlicesRuleDefinition.Creator.matching(vararg packageIdentifiers: String) =
    assignedFrom(MultiPackageMatchingSliceIdentifier(*packageIdentifiers))

private class MultiPackageMatchingSliceIdentifier(private vararg val packageIdentifiers: String) : SliceAssignment {
    override fun getIdentifierOf(javaClass: JavaClass): SliceIdentifier =
        packageIdentifiers
            .map {
                PackageMatcher.of(it)
                    .match(javaClass.packageName)
                    .map(TO_GROUPS)
                    .orElse(emptyList())
            }
            .filter { it.isNotEmpty() }
            .let { it.flatten() }
            .let { parts -> identifierOf(parts) }

    private fun identifierOf(parts: List<String>?) =
        if (parts.isNullOrEmpty()) ignore() else of(parts)

    override fun getDescription(): String {
        return packageIdentifiers.joinToString(", ") { "'$it'" }
    }
}
```

codecholeric:
[...]

PR #896, opened 2022-06-22
Add support for alternatives to package identifiers

Pfoerd:
Adds the ability to define alternatives such as `(a|b)` in package identifiers. This allows to support architectures that for some reasons need a namespace for some components by using different packages even though they logically belong to the same architectural entity like a layer, domain, etc. This is especially useful if your code needs to satisfy some "overarching" structural requirements that come from outside your actual project, e.g. from an enterprise architecture.

* `pack.[a.c|b*].d` matches `pack.a.c.d` or `pack.bar.d`, but neither `pack.a.d` nor `pack.b.c.d`
* `..service.(a|b*)..` matches `a.service.bar.more` and group 1 would be `bar`
* Nested alternatives like `(?:..application.(*).*..)|(?:..domain.api|logic.(*)..)` are not allowed to keep the syntax of package identifiers simple and readable.

Resolves: #662

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/PackageMatcher.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/PackageMatcher.java
@@ -22,7 +22,9 @@ import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.IntStream;
+import java.util.stream.Stream;
 
+import com.google.common.base.CharMatcher;
 import com.google.common.collect.ImmutableSet;
 import com.tngtech.archunit.PublicAPI;
 
@@ -41,6 +43,12 @@ import static java.util.stream.Collectors.toList;
  * <li><b>{@code '*.*.pack*..'}</b> matches <b>{@code 'a.b.packfix.c.d'}</b>,
  * but neither <b>{@code 'a.packfix.b'}</b> nor <b>{@code 'a.b.prepack.d'}</b></li>
  * </ul>
+ * You can also use alternations with the '|' operator within brackets. For example
+ * <ul>
+ * <li><b>{@code 'pack.[a.c|b*].d'}</b> matches <b>{@code 'pack.a.c.d'} or <b>{@code 'pack.bar.d'}</b>, but neither
+ * <b>{@code 'pack.a.d'}</b> nor <b>{@code 'pack.b.c.d'}</b></li>
+ * </ul>
+ * <p>
  * Furthermore, the use of capturing groups is supported. In this case '(*)' matches any sequence of characters,
  * but not the dot '.', while '(**)' matches any sequence including the dot. <br>
  * For example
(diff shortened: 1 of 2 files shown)


Pfoerd:
@codecholeric maybe this could be a good improvement for stable release 1.0 too :wink: ?

codecholeric:
I'll try to find time to review this weekend :wink:

codecholeric:
Okay, I managed to look into it :slightly_smiling_face: Thanks a lot for your support! I really like the approach of just extending the syntax a little bit without going overboard, but still making it way more powerful with this one little feature :smiley: 

I added some minor changes, because I think the check for nested groups was a little too weak. In particular, it would only find the nesting, if the brackets were directly following each other. But I could also do something like `..[in.(the).middle|other]..` and could bypass the check by that.

Looking into it for a while the following things came clear to me:
1) Nesting capturing groups into non-capturing groups would be really complicated to implement, because for `..I.think.(a)|and.(b).are.group.one` there would be two capturing groups, even though by alternation they could technically both be group 1. But this behavior would make it very complicated to do the correct package capturing
2) Nesting noncapturing alternations into other noncapturing alternations would complicate the transformation to regex a lot as well
3) While at the moment not explicitly checked nested capturing groups also make no sense for package matching
4) Nesting noncapturing groups into capturing groups might allow some ugly complicated stuff to write, but it's easy to implement and doesn't cause any trouble.

Based on that I tried to adjust it to forbid cases 1-3 and to make checking this easier also toplevel alternations. I also added a check that forbids noncapturing groups without alternations, because that just doesn't make sense :man_shrugging: 

I also added tests that mirror the problem in #685 and I think with this change it would now be possible to solve #685 just with the package identifier syntax.

Let me know what you think! Does the change make sense to you?

Pfoerd:
Thanks for your effort and for adding proper checks for the nested groups, you're totally right that they were too weak!

I'm a bit torn about the support for nesting noncapturing groups into capturing groups... I fully agree that this is a "low hanging fruit" from an implementation pov and can solve #685. 
On the other hand it makes the API of package matching more complicated and a bit ambiguous... From a user's pov one might ask "why can I do this kind of nested alternation but not the others?"... this introduces a certain inconsistency right? 

I'm wondering whether it could be better to only add simple top-level alternations without nesting options to the package identifier syntax and extend the APIs to add a full regex based package matching approach for complex usecases like the one in #685 (like suggested [here](https://github.com/TNG/ArchUnit/issues/662#issuecomment-964060058)). What do you think?

codecholeric:
Yeah, you might be right. Maybe I got pulled a little bit too much into "this is easy to implement and gives more power" versus "this is consistent and easy to understand". Let's forbid all nesting then and tackle the rest in a more powerful extended API :+1:
Fun fact, to solve the case in #685 you don't need it anyway :wink: Because the tests I wrote to cover these examples don't need nesting within the capturing groups... At least if I have understood it correctly :man_shrugging:

Pfoerd:
Thank you for your support again :) . Btw do you have any timeline when 1.0 will be released?

PR #899, opened 2022-06-23
Feature/issue 641 filtering env props

crizzis:


archunit-junit/junit5/engine/build.gradle
```diff
@@ -1,23 +1,29 @@
 plugins {
     id 'archunit.java-release-conventions'
+    id 'java-test-fixtures'
 }
 
 ext.moduleName = 'com.tngtech.archunit.junit5.engine'
 
-ext.minimumJavaVersion = JavaVersion.VERSION_1_8
+ext.minimumJavaVersion = JavaVersion.VERSION_1_9
```
codecholeric, line 8: We can't do this or we will break `archunit-junit5` support for all users out there that use Java 8!

archunit-junit/junit5/engine/build.gradle
```diff
@@ -1,23 +1,29 @@
...
 
 dependencies {
     api project(path: ':archunit')
     api project(path: ':archunit-junit5-api')
     api project(path: ':archunit-junit5-engine-api')
+
     implementation project(path: ':archunit-junit')
     dependency.addGuava { dependencyNotation, config -> implementation(dependencyNotation, config) }
     implementation dependency.slf4j
+    implementation dependency.junitPlatformLauncher
```
codecholeric, line 18: We should not add any dependencies on unrelated JUnit projects. Only depend on the platform. I think this would also be a strange dependency, since the Platform Launcher launches `TestEngines`, so a dependency back from `TestEngines` to the launcher feels wrong to me :man_shrugging:

archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/ArchUnitTestEngine.java
```diff
@@ -67,6 +72,22 @@
 public final class ArchUnitTestEngine extends HierarchicalTestEngine<ArchUnitEngineExecutionContext> {
     static final String UNIQUE_ID = "archunit";
 
+    private static final Collection<String> ABORTING_THROWABLE_NAMES = Arrays.asList(
+            "org.junit.internal.AssumptionViolatedException",
```
codecholeric, line 76: Why would we support JUnit 4 exceptions within the JUnit 5 support? :thinking: Besides that `.internal.` anyway sounds in a way that we should not handle it :man_shrugging:

archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/filtering/AbstractTestNameFilter.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private void replaceFilter(
+            EngineDiscoveryRequest discoveryRequest,
+            PostDiscoveryFilter postDiscoveryFilter,
+            PostDiscoveryFilter replacement) throws ReflectiveOperationException {
+        LauncherDiscoveryRequest request = (LauncherDiscoveryRequest) discoveryRequest;
+        List<PostDiscoveryFilter> filters = new ArrayList<>(request.getPostDiscoveryFilters());
+        filters.remove(postDiscoveryFilter);
+        filters.add(replacement);
+        ReflectionUtils.setField(discoveryRequest, "postDiscoveryFilters", filters);
```
codecholeric, line 89: This feels wrong to me :thinking: The JUnit Platform should provide extension points to filter without the need to break encapsulations of the platform in such extreme ways (even having to break the `final` modifier)

codecholeric:
Thanks for your hard work! I just had a quick look and added some comments for things I noticed at first sight. One important thing to keep in mind is that this is a library, not an application. I.e. we need to be extra conservative about dependencies in production code. And we can't freely adjust the required Java version, etc.

Also, can you explain why we need the tooling tests for this change already? I would imagine that this should be easily testable in a blackbox way by just extending e.g. the `ArchUnitTestEngineTest` class with tests that basically do

```
// before
System.setProperty(ARCHUNIT_INCLUDE_FILTER, "some*crazyFilter.*method")

// when -> test that discovery filters out non-included tests
```

archunit-junit/junit5/engine/build.gradle
crizzis: Ummm...

The getting started doc for ArchUnit says: 

> ArchUnit can be obtained from Maven Central.
> 
> Gradle
> dependencies {
>    testImplementation 'com.tngtech.archunit:archunit:0.23.1'
> }

and the `archunit` module already declares `minimumJavaVersion` to be 9, so I was under the impression ArchUnit already dropped support for Java 8

archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/filtering/AbstractTestNameFilter.java
crizzis: This `replaceFilter` will only be used for Gradle. I can remove it for now. 

I can explain the need for this workaround and how what we could do to get rid of it in the Gradle PR, would that be okay with you?

archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/ArchUnitTestEngine.java
crizzis: Those are all the flavors of assumption violated exceptions that can be thrown to indicate assumption violation (i.e. exceptions that should result in skip results, not failures). 

As to whether we should be handling the exception or not, please take a look at JUnit's [`OpenTest4JAndJUnit4AwareThrowableCollector`](https://github.com/junit-team/junit5/blob/main/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/support/OpenTest4JAndJUnit4AwareThrowableCollector.java). 

> Why would we support JUnit 4 exceptions within the JUnit 5 support?

Remember that you can run JUnit 4 ArchUnit tests using the `ArchUnitEngine` (NOT Vintage!)

archunit-junit/junit5/engine/build.gradle
crizzis: In that case, please have a look at `AbstractTestNameFilter.findPostDiscoveryFilter`. We need a dependency on Launcher to be able to cast the discovery request to `LauncherDiscoveryRequest`. 

If you insist on removing this dependency, please suggest a way to obtain the same result. We can, for instance, use reflection all the way through, I was just trying to maintain some level of compile-time safety.

crizzis:
> One important thing to keep in mind is that this is a library, not an application. I.e. we need to be extra conservative about dependencies in production code. And we can't freely adjust the required Java version, etc.

I really tried to keep new dependencies to a bare minimum, but please take a look at my responses - I suggested workarounds for some of the points you've raised. 

> Also, can you explain why we need the tooling tests for this change already? I would imagine that this should be easily testable in a blackbox way by just extending e.g. the `ArchUnitTestEngineTest` class with tests that basically do
> 
> ```
> // before
> System.setProperty(ARCHUNIT_INCLUDE_FILTER, "some*crazyFilter.*method")
> 
> // when -> test that discovery filters out non-included tests
> ```

This is really pending [`your decision`](https://github.com/TNG/ArchUnit/issues/641#issuecomment-1157756907). I agree that it would be super easy to just test it inside `ArchUnitTestEngineTest`. Though, if you decide for the new tests, then the same feature will be tested twice. If that's okay with you, and you feel a test inside `ArchUnitTestEngineTest` is needed anyway, please let me know.

archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/filtering/ArchUnitPropsTestSourceFilter.java
```diff
@@ -0,0 +1,137 @@
...
+            return Arrays.stream(patterns)
+                    .map(this::prepareRegex)
+                    .collect(Collectors.toList());
+        }
+
+        private Pattern prepareRegex(String pattern) {
+            String escaped = SPECIAL_REGEX_CHARS.matcher(pattern).replaceAll(ESCAPED_INPUT);
+            escaped = MULTIPLE_ESCAPED_ASTERISK.matcher(escaped).replaceAll(SINGLE_ESCAPED_ASTERISK_STRING);
+            escaped = SINGLE_ESCAPED_ASTERISK.matcher(escaped).replaceAll(MATCH_ALL_CHARACTERS);
+            return Pattern.compile("(?:^|\\.)" + escaped + "$", Pattern.CASE_INSENSITIVE | Pattern.UNICODE_CASE);
```
crizzis, line 124: In case the `:^|\\.` needs explanation, `^` is supposed to match FQ names, while `\\.` is supposed to match simple names (i.e. match the pattern starting from a dot, presumably the final dot in a package name)

archunit-junit/junit5/engine/src/testFixtures/java/com/example/SomeClass.java
```diff
@@ -0,0 +1,19 @@
+package com.example;
+
+import java.io.Serializable;
+
+import com.tngtech.archunit.junit.ArchTest;
+import com.tngtech.archunit.lang.ArchRule;
+
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.classes;
+
+public class SomeClass {
```
crizzis, line 10: Those files need to exist in order for `FieldSource.from` / `FieldSelector.selectField` to not throw `NoClassDefFoundError`s

archunit-junit/junit5/engine/build.gradle
```diff
@@ -62,7 +68,3 @@ def configureDependencies = { deps ->
     }
 }
 this.with project(':archunit-junit').configureJUnitArchive(configureDependencies)
-
-singlePackageExport {
-    exportedPackage = 'com.tngtech.archunit.junit.internal'
```
crizzis, line 67: @codecholeric Could you please suggest what I should do about this one?

If the other PRs are merged, there's going to be a lot more filtering stuff, so I kind of felt it would be clearer if those resided in a separate subpackage...

archunit-junit/junit5/engine/build.gradle
hankem: All previous releases (i.e. up to 0.23.1) actually support _Java 7;_ its support has only recently been dropped with #833. 😉

`minimumJavaVersion` is only set to `JavaVersion.VERSION_1_9` (or `JavaVersion.VERSION_16`) for individual Gradle tasks `compileJdk9mainJava`, `compileJdk9testJava`, `jdk9Test` (or `compileJdk16testJava`, `jdk16Test`, respectively).

crizzis: Duly noted. 

I've downgraded the version back to Java 8 (turns out Java 9 features are not yet used in this PR), but **please** consider adding this to the contributing guide as it's pretty confusing, especially given that some modules use `ext.minimumJavaVersion = JavaVersion.VERSION_1_9`

archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/ArchUnitTestEngine.java
crizzis: > JUnit 4 support has a different annotation @AnalyzeClasses than JUnit 5 support has.

Different *how*? They have the same name and package. 

This is exactly the configuration issue I was trying to tell you about. If you run your tests with both archunit and vintage engines, the tests are being run twice. They are eligible for discovery by both engines. Try debugging the tooling tests with both engines on, you will see both engines executing the test cases. 

> I don't think users should use JUnit 4 things within the scope of the ArchUnitTestEngine.

The question is do users know that? If they use the method flavor of archunit tests, they are allowed to use `assertThat`. Why wouldn't they be allowed to use `assumeThat`?

archunit-junit/junit5/engine/build.gradle
codecholeric: What I don't understand is why we would need any dependencies on other libraries or tooling at all to solve this problem. The first issue is simply: "Given a system property `archunit.junit.includePattern`, only execute tests that match the pattern", right?
Say I
1) Convert the pattern to a filter for `(ArchUnit)TestDescriptor` (we could e.g. analyse the `UniqueId`, which in the end is also just a concatenation like `.../[class:some.SomeClass]/[field:someField]` that matches the info we have in the pattern very closely; just not sure how stable this is; otherwise we could add a custom property to `(Abstract)ArchUnitTestDescriptor` or sth. like that)
2) Within `com.tngtech.archunit.junit.internal.ArchUnitTestEngine.discover(..)` once all the discoveries are through, right before I return the `result`, I add something like
```
result.accept(descriptor -> {
  if (!includeFilterMatches(descriptor)) {
    descriptor.removeFromHierarchy();
  }
});
```

Wouldn't that solve the problem? Tests I could write very easily as part of `ArchUnitTestEngineTest.Discovers`:

```
// GIVEN filter is set via system property
System.setProperty("archunit.junit.includePattern", "*SomeClass*");

// WHEN classes are discovered
testEngine.discover(/* discovery request that would normally include SomeClass */);

// THEN classes matching the system property filter are excluded
Set<String> displayNames = descriptor.getChildren().stream().map(TestDescriptor::getDisplayName).collect(toSet());
assertThat(displayNames).doesNotContain("SomeClass");
```

Maybe I'm missing something, but to me it looks like this could solve the issue without any need for additional test infrastructure or dependencies on other 3rd party libraries? :thinking:

archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/ArchUnitTestEngine.java
crizzis: > it is not supported to use archunit-junit4 and archunit-junit5 together on the classpath

Well, I mean... that changes things. A lot :p 

I was talking about the scenario where you have a project with JUnit 4 arch tests, with both archunit and junit-vintage engines enabled. In that case, each test gets picked up by junit-vintage (because of @RunWith(ArchunitRunner.class)) *and* by archunit (because of @AnalyzeClasses). You can't have both ArchUnitRunner *and* archunit engine on the classpath, though, while not simultaneously depending on both archunit-junit dependencies. So that makes sense to me now. 

Where exactly do the docs say that mixing archunit-junit4 and archunit-junit5 is not supported, btw? I read the user guide on junit support, but I couldn't find it anywhere...

In that case, I now realize I wrongly assumed support for test filtering was working with junit-vintage when actually it just happened to work because the archunit engine was the one that was running the tests :(

...which means there is no point in any support for junit 4 stuff in this or any of the other PRs. I will try to update the PRs accordingly next week. Thanks for your patience:)

archunit-junit/junit5/engine/build.gradle
crizzis: > 1. Convert the pattern to a filter for (ArchUnit)TestDescriptor (we could e.g. analyse the UniqueId, which in the end is also just a concatenation like .../[class:some.SomeClass]/[field:someField] that matches the info we have in the pattern very closely; just not sure how stable this is; otherwise we could add a custom property to (Abstract)ArchUnitTestDescriptor or sth. like that)
2. Within com.tngtech.archunit.junit.internal.ArchUnitTestEngine.discover(..) once all the discoveries are through, right before I return the result, I add something like ```result.accept(descriptor -> {
  if (!includeFilterMatches(descriptor)) {
    descriptor.removeFromHierarchy();
  }
});```

Well, um, yes, this is essentially what this PR does. The only difference is it filters child descriptors before they are added, and not afterwards. 

> Tests I could write very easily as part of ArchUnitTestEngineTest.Discovers

Sure, I've already done that. Have a look at the relevant tests. 

> this could solve the issue without any need for additional test infrastructure or dependencies on other 3rd party libraries

Yes, absolutely. This PR does not depend on 3rd party libraries to be able to perform the filtering. It depends on 3rd party libraries solely because: 

> I committed some of the code in advance as it's kind of hard to keep 3 PRs in line, especially when one defines an abstract base class for use with the other ones :(

I can try to isolate the changes needed in the other PRs, but this will require splitting some commits. It will also take a while. Are you OK with that?

crizzis:
> | EDIT# no. 2 Added relevant test to ArchUnitTestEngineTest; also, fixed a couple of bugs
> 
> If there are bugs in there it would be cool if you could create a PR only fixing that instead of mixing it. Then we could trace it way easier 🙂 Also bugs we should fix ASAP without waiting for a full feature to be done...

Sorry, I meant bugs **within the solution**, not existing bugs in ArchUnit

archunit-junit/junit5/engine/build.gradle
codecholeric: Sorry for the delay :grimacing: I was a little occupied to do that 1.0.0-rc1 release. Yes, please, if you don't mind! Make the first PR as small and simple as possible. Also, if we don't need all that test infrastructure code yet, it's also good, because then we can delay the decision further. The smaller and simpler the change, the easier it is to understand later on when we look back at the history. Also a small code change is easy to argue for, a big one needs to be pondered really carefully, if it brings the ROI on the maintenance costs it will cause.

codecholeric:
I implemented a simplified version in #1280. Since I have no capacity to work on this further, I'm gonna close this. Feel free to reopen if you want to pick this up or add further improvements.

Issue #475, opened 2020-11-23
General rule to test that class and its test are both at the same package


ars-java:
It will be convenient to have `GeneralCodingRules.TEST_SHOULD_HAVE_SAME_PACKAGE` rule to test that classes `Xxx` from `/src/` and `XxxTest` from `/test/` are both at the same package

hankem:
You can define a custom `@ArchTest` like in https://github.com/TNG/ArchUnit/issues/20#issuecomment-328225761.

ars-java:
Yes, I can. But I think that it's a useful common and reusable feature to add it to the library. It's more fundamental then restrictions for joda-time or JUL

codecholeric:
I agree, even though I wonder if the naming pattern should be customizable :thinking: 
Like `GeneralCodingRules.testsShouldResideInTheSamePackageIdentifyingTestsByName("$Test")`), and someone else could write `...IdentifyingTestsByName("$IntegrationTest")` or `...IdentifyingTestsByName("$Tests")` for example.
Anyway, if you have time I would be happy to support you with a PR for this feature :wink: Or if anybody else wants to contribute. I think it would be a good starter task :smiley: Otherwise I have to put it in my pipeline and implement it when I get to it, but looking at my task pipeline that could take a while :disappointed:

Rossalli:
Hi, @codecholeric! Could you please assign this to me? I would like to try help with this issue :)

Thanks!

mslowiak:
@Rossalli are you still working on that one? can I pick this up?

Rossalli:
Hi @mslowiak, I'm working on that :)

thatsIch:
@Rossalli how is your progress?

codecholeric:
Since the last comment from @Rossalli is from almost a year ago and there has been no reply to the last ping for over a month I would consider this issue open for grabs again...

mslowiak:
@codecholeric So, can I pick this up? :)

codecholeric:
Yes, feel free to pick it up! (Sorry if you're really still working on it @Rossalli, but without any life sign for so long I can't block the issue forever)

codecholeric:
@mslowiak I assigned you to it for transparency, hope that's okay :wink:

PR #908, opened 2022-07-11
Add rule checking test and implementation class are in same package

mslowiak:
This will add the library rule `GeneralCodingRules.testClassesShouldResideInTheSamePackageAsImplementation(..)` which will test that implementation and test class reside in the same package. The rule can e.g. detect mismatches like `com.myapp.correct.SomeClass` and `com.myapp.wrong.SomeClassTest`.

Resolves: #475

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/library/GeneralCodingRules.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/GeneralCodingRules.java
@@ -15,6 +15,10 @@
  */
 package com.tngtech.archunit.library;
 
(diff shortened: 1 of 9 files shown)


codecholeric:
Thanks a lot for your contribution!! :smile: Looks very good in general :+1: I just made some small adjustments, mainly

* adjust the message a little to be more consistent with other messages
* use common testing infrastructure to test the rule

I just adjusted it, since I want to release these days, so I cut down the round trip a little. I hope that's okay with you! :pray:

PR #909, opened 2022-07-12
Remove all current deprecations

codecholeric:
Since we are about to release a new major version we can finally throw out all existing deprecations.

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/AccessTarget.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/AccessTarget.java
@@ -19,7 +19,6 @@ import java.lang.annotation.Annotation;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.Set;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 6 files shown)


perlun:
If you're unlucky like me, you might have been using a really-old version like 0.14.1 where methods like `JavaClass.namesOf()` wasn't even deprecated yet. :grin: This was a bit confusing in the upgrade. Luckily. `git clone git@github.com:TNG/ArchUnit.git` and looking at the `git log -p ./archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java` output made it fairly simple to see what the method was replaced with.

(Hint: Use `Formatters.formatNamesOf()` or look at the diff to see the full details here)

It might even be worth suggesting for people still stuck on the `0.x` versions to **first** upgrade to https://github.com/TNG/ArchUnit/releases/tag/v0.23.1 before attempting the 1.0.x upgrade, since that will highlight all deprecations. That would probably have helped me. :thinking: I didn't realize it until now though.

----

(Btw, _big thanks_ for doing your very best to try and apply SemVer correctly! Can't tell you how much that means to me. :pray: Greatly appreciated!)

Issue #919, opened 2022-07-15
JavaPackage#getPackageDependenciesFromSelf returns dependencies from both self AND subpackages


grimsa:
In `JavaPacakge`:
- `getPackageDependenciesFromSelf` collects packages of classes returned by `getClassDependenciesFromSelf`
- `getClassDependenciesFromSelf` collects dependencies originating from classes returned by `getAllClasses`
- `getAllClasses` collects classes from both the current package AND all subpackages (descendants)

This caught me by surprise - I expected `getPackageDependenciesFromSelf` to only return the package dependencies originating from classes in the current package (i.e. excluding any subpackages).

At first glance it would seem that just replacing `getAllClasses` with `getClasses` in `getClassDependenciesFrom/ToSelf` would make it match my expectation.

If this is a bug, I'd be happy to open a PR with a fix.

codecholeric:
Thanks for raising this issue! The behavior actually was implemented like this intentionally :thinking: Because that would have been **my** expectation of the API :joy: I admit that the javadoc does not specify this :disappointed:
The reason for the current behavior is that when considering architecture we usually consider hierarchies, right? Packages model components where subpackages model subcomponents. So if I take my component and ask "what dependencies does it have", then it's the sum of all dependencies of the component plus its subcomponents. At least that was my thinking.
What is your use case for only considering the "flat" dependencies?
In any case, maybe we need both APIs? :thinking: Because dependencies of the flat content + subpackages is definitely a common use case for me :man_shrugging: Take for example you want to draw a diagram, then you would also take the package + all subpackages, no? :thinking:

grimsa:
I noticed this when trying to implement a "a package should not depend on its subpackages" rule (reasoning: https://javadevguy.wordpress.com/2017/12/18/happy-packaging/).

What I was trying to do was something like this:
```
    @ArchTest
    private static ArchRule packageRule = ArchRuleDefinition.all(packages)
            .that(DescribedPredicate.describe(
                    "our company packages",
                    (JavaPackage javaPackage) -> javaPackage.getName().startsWith("com.company")
            ))
            .should(new ArchCondition<>("not depend on their subpackages") {
                @Override
                public void check(JavaPackage javaPackage, ConditionEvents events) {
                    if (javaPackage.getClasses().isEmpty()) {
                        // if package has no classes, then it is not relevant
                        return;
                    }
                    Set<JavaPackage> allSubpackages = javaPackage.getAllSubpackages();
[...]

PR #922, opened 2022-07-20
improve naming around `JavaPackage` dependencies

grimsa:
This breaking change will clear up some semantic ambiguities around the methods of `JavaPackage` that deal with subpackages. In particular, we will now consider dependencies from/to a package only those where the origin/target class **directly** resides within the package. Subpackages are considered only those packages that reside **directly** within a package. Everything that considers all classes within the package and all subpackages recursively (i.e. subpackages of subpackages and so on) will now refer to the term "package tree".

Changes:
* Added `getClassDependencies{From/To}ThisPackage`
* Renamed `getClassDependencies{From/To}Self` to `getClassDependencies{From/To}ThisPackageTree`
* Added `getPackageDependencies{From/To}ThisPackage`
* Renamed `getPackageDependencies{From/To}Self` to `getPackageDependencies{From/To}ThisPackageTree`
* Renamed `getAllSubpackages` to `getSubpackagesInTree`
* Renamed `accept(..., ClassVisitor/PackageVisitor)` to `traversePackageTree(..., ClassVisitor/PackageVisitor)`

Resolves: #919

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaPackage.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaPackage.java
@@ -50,6 +50,40 @@ import static java.util.Collections.emptySet;
 import static java.util.Collections.singleton;
 import static java.util.stream.Collectors.toSet;
 
+/**
+ * Represents a package of Java classes as defined by the
+ * <a href="https://docs.oracle.com/javase/specs/jls/se17/html/jls-7.html">Java Language Specification</a>.
+ * I.e. a namespace/group for related classes, where each package can also contain further subpackages.
+ * Thus, packages define a hierarchical tree-like structure.<br>
+ * An example would be the package {@code java.lang} which contains {@code java.lang.Object}.<br>
+ * ArchUnit will consider the "classes of a package" to be the classes residing <b>directly</b>
+ * within the package. Furthermore, "subpackages" of a package are considered packages that are residing
+ * <b>directly</b> within this package. On the contrary, ArchUnit will call the hierarchical tree-like structure
+ * consisting of all packages that can be reached by traversing subpackages, subpackages of subpackages, etc.,
+ * as "package tree".
+ * <br><br>
+ * Take for example the classes
+ * <pre><code>
+ * com.example.TopLevel
+ * com.example.first.First
+ * com.example.first.nested.FirstNested
+ * com.example.first.nested.deeper_nested.FirstDeeperNested
+ * com.example.second.Second
+ * com.example.second.nested.SecondNested
+ * </code></pre>
+ * Then the package {@code com.example} would contain only the class {@code com.example.TopLevel}. It would also
+ * contain two subpackages {@code com.example.first} and {@code com.example.second} (but not {@code com.example.first.nested}
+ * as that is not <b>directly</b> contained within {@code com.example}).<br>
+ * The package tree of {@code com.example} would contain all packages (and classes within)
+ * <pre><code>
+ * {@code com.example}
+ * {@code com.example.first}
+ * {@code com.example.first.nested}
+ * {@code com.example.first.nested.deeper_nested}
+ * {@code com.example.second}
+ * {@code com.example.second.nested}
+ * </code></pre>
+ */
 public final class JavaPackage implements HasName, HasAnnotations<JavaPackage> {
     private final String name;
     private final String relativeName;
@@ -83,6 +117,11 @@ public final class JavaPackage implements HasName, HasAnnotations<JavaPackage> {
         return relativeName;
     }
 
+    /**
+     * @return The {@link JavaClass} representing the compiled {@code package-info.class} file of this {@link JavaPackage}
+     *         (for details refer to the <a href="https://docs.oracle.com/javase/specs/jls/se17/html/jls-7.html">Java Language Specification</a>).
+     *         Will throw an {@link IllegalArgumentException} if no {@code package-info} exists in this package.
+     */
     @PublicAPI(usage = ACCESS)
     public HasAnnotations<?> getPackageInfo() {
         Optional<? extends HasAnnotations<?>> packageInfo = tryGetPackageInfo();
@@ -92,11 +131,20 @@ public final class JavaPackage implements HasName, HasAnnotations<JavaPackage> {
         return packageInfo.get();
     }
 
+    /**
+     * @return The {@link JavaClass} representing the compiled {@code package-info.class} file of this {@link JavaPackage}
+     *         or {@link Optional#empty()} if no {@code package-info} exists in this package
+     *         (for details refer to the <a href="https://docs.oracle.com/javase/specs/jls/se17/html/jls-7.html">Java Language Specification</a>).
+     */
     @PublicAPI(usage = ACCESS)
     public Optional<? extends HasAnnotations<?>> tryGetPackageInfo() {
         return packageInfo;
     }
 
+    /**
+     * @return All annotations on the compiled {@link #getPackageInfo() package-info.class} file
+     *         (for details refer to the <a href="https://docs.oracle.com/javase/specs/jls/se17/html/jls-7.html">Java Language Specification</a>).
+     */
     @Override
     @PublicAPI(usage = ACCESS)
     public Set<? extends JavaAnnotation<JavaPackage>> getAnnotations() {
@@ -105,12 +153,29 @@ public final class JavaPackage implements HasName, HasAnnotations<JavaPackage> {
                 .orElse(emptySet());
     }
 
+    /**
+     * @return The {@link Annotation} of the given type on the {@link #getPackageInfo() package-info.class} of this package
+     *         (for details refer to the <a href="https://docs.oracle.com/javase/specs/jls/se17/html/jls-7.html">Java Language Specification</a>).
+     *         Will throw an {@link IllegalArgumentException} if either there is no {@code package-info}
+     *         or the {@code package-info} is not annotated with the respective annotation type.
+     * @param <A> The type of the {@link Annotation} to retrieve
+     * @see #tryGetAnnotationOfType(Class)
+     * @see #getAnnotationOfType(String)
+     */
     @Override
     @PublicAPI(usage = ACCESS)
     public <A extends Annotation> A getAnnotationOfType(Class<A> type) {
         return getAnnotationOfType(type.getName()).as(type);
     }
 
+    /**
+     * @return The {@link JavaAnnotation} matching the given type on the {@link #getPackageInfo() package-info.class} of this package
+     *         (for details refer to the <a href="https://docs.oracle.com/javase/specs/jls/se17/html/jls-7.html">Java Language Specification</a>).
+     *         Will throw an {@link IllegalArgumentException} if either there is no {@code package-info}
+     *         or the {@code package-info} is not annotated with the respective annotation type.
+     * @see #tryGetAnnotationOfType(String)
+     * @see #getAnnotationOfType(Class)
+     */
     @Override
     @PublicAPI(usage = ACCESS)
     public JavaAnnotation<JavaPackage> getAnnotationOfType(String typeName) {
@@ -121,6 +186,14 @@ public final class JavaPackage implements HasName, HasAnnotations<JavaPackage> {
         return annotation.get();
     }
 
+    /**
+     * @return The {@link Annotation} of the given type on the {@link #getPackageInfo() package-info.class}
+     *         of this package or {@link Optional#empty()} if either there is no {@code package-info}
+     *         or the {@code package-info} is not annotated with the respective annotation type.
+     * @param <A> The type of the {@link Annotation} to retrieve
+     * @see #getAnnotationOfType(Class)
+     * @see #tryGetAnnotationOfType(String)
+     */
     @Override
     @PublicAPI(usage = ACCESS)
     public <A extends Annotation> Optional<A> tryGetAnnotationOfType(Class<A> type) {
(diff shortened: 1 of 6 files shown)


archunit/src/test/java/com/tngtech/archunit/core/domain/JavaPackageTest.java
```diff
@@ -241,77 +241,151 @@ public void visits_packages() {
...
+        assertThat(examplePackage.getPackage("third").getPackageDependenciesToThisPackage())
+                .isEmpty();
 
-        assertThat(examplePackage.getPackage("second").getPackageDependenciesToSelf())
+        assertThat(examplePackage.getPackage("unrelated").getPackageDependenciesToThisPackage())
+                .isEmpty();
+    }
+
+    @Test
+    public void has_package_dependencies_to_this_package_and_subpackages() {
```
codecholeric, line 373: Except for `.getPackage("third")` this case seems to be identical with the one before. Should we add another class, or move a class within `"first"` to a subpackage to make the first assertion more clearly different?

codecholeric (changes requested):
Thanks a lot for your support :smile: Looks very good to me, just added a couple of minor comments you can look into :slightly_smiling_face:

codecholeric:
Also, maybe you could make the commit message a little more elaborate :slightly_smiling_face: Because "some issues" is quite fuzzy. What about "explicitly distinguish between direct and subpackage dependencies" or sth. like that as the header? (to keep it below the 70 chars) And then put some details in the body into which issue you ran and why you were confused and how we're solving it now? Just so when I see just this commit I have all the context from the message to understand what was the exact problem and how we decided to fix it (and why not any other way :grin:)

grimsa:
👍 I'll update the PR sometime over the next few days.

@codecholeric In the meantime - have you seen this https://github.com/TNG/ArchUnit/issues/919#issuecomment-1189998749? Maybe a larger-scale change would be desired.

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaPackageTest.java
```diff
@@ -241,77 +241,151 @@ public void visits_packages() {
...
 
     @Test
-    public void has_class_dependencies_to_other_packages() {
+    public void has_class_dependencies_from_this_package() {
         JavaPackage examplePackage = importPackage("packageexamples");
 
-        assertThatDependencies(examplePackage.getPackage("second").getClassDependenciesFromSelf())
+        assertThatDependencies(examplePackage.getPackage("second").getClassDependenciesFromThisPackage())
+                .contain(Second1.class, First2.class)
+                .doesNotContain(ClassDependingOnOtherSecondClass.class, Second1.class)
```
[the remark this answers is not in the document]
grimsa: I'm not sure I followed this.

Do you mean that if we added `second.Second2` that depends on `second.sub.SecondSub1`, this dependency should be returned from `getClassDependenciesFromThisPackage` but NOT from `getClassDependenciesFromThisPackageAndSubpackages`?
The current code would return it from both these methods.

codecholeric:
Hey, sorry, I was knocked out a little the recent weeks :grimacing: I answered at your comment. Given what you said, do you think it would make more sense to call the methods `get...ThisPackageTree` with a good Javadoc, instead of `get...ThisPackageAndSubpackages`? Since I agree that "subpackages" again could be understood as only the "direct" subpackages (it would be a little bit weird API to only consider one level of packages for such an API, but still not nice)

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaPackageTest.java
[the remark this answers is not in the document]
codecholeric: Hmm, I would have thought, that this example should not be contained in the dependencies, because from the view of `thisPackageAndSubpackages` this is a self-dependency, right? :thinking: I.e. why filter out dependencies where origin or target are in the root package, but not for subpackages? :thinking: I think that was an original oversight and should not be that way, I'm gonna fix it...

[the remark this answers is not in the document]
codecholeric: I fixed it and rebased your commit on top. So please check if that makes sense. But I think this was broken before, if I say "I consider all classes in the tree" for dependencies, then I should also treat self-references with respect to that set. Why would `mypkg.A -> mypkg.B` not be contained, but `mypkg.A -> mypkg.sub.B` would be? :man_shrugging: Anyway, let me know what you think about that...

codecholeric:
Since you seem a little overloaded to push this forward at the moment I created some fixups with some review suggestions (including the renaming from `ThisPackageAndAllSubpackages` to `ThisPackageTree`, since you have me convinced, that "subpackage" by definition would only be a direct child :wink:)
I also rebased onto the current `main` branch and added some preparatory refactorings, so maybe back up your local branch before you pull in this stuff :wink: 
If you don't find the time at all to look into this, let me know! (and no worries, I know how it is with additional time for OSS :wink:) I can also try to find some additional support to review :slightly_smiling_face: Otherwise, let me know what you think about these suggestions, if you think they make sense!
I'm just trying to push this, because I want to include it into `1.0.0`, and I feel slowly the window for review remarks of `1.0.0-rc1` can be closed :slightly_smiling_face:

grimsa:
Hey, thanks a lot for moving on this, and sorry for dropping the ball :)
I will check this today.

codecholeric:
No worries, as I said, I know how it is with this extra time for OSS :wink:

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaPackage.java
```diff
@@ -207,15 +302,15 @@ private void setParent(JavaPackage parent) {
...
+     * @return all classes directly contained in this package, no classes in subpackages (compare {@link #getAllClasses()})
      */
     @PublicAPI(usage = ACCESS)
     public Set<JavaClass> getClasses() {
         return classes;
     }
 
     /**
-     * @return all classes contained in this package or any sub-package (compare {@link #getClasses()})
+     * @return all classes contained in this package or any subpackage (compare {@link #getClasses()})
```
grimsa, line 313:
```suggestion
     * @return all classes contained in this package tree (compare {@link #getClasses()})
```

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaPackage.java
```diff
@@ -207,15 +302,15 @@ private void setParent(JavaPackage parent) {
     }
 
     /**
-     * @return all classes directly contained in this package, no classes in sub-packages (compare {@link #getAllClasses()})
+     * @return all classes directly contained in this package, no classes in subpackages (compare {@link #getAllClasses()})
```
grimsa, line 305:
```suggestion
     * @return all classes directly contained in this package, but not classes in the lower levels of the package tree (compare {@link #getAllClasses()})
```

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaPackage.java
```diff
@@ -227,32 +322,32 @@ public Set<JavaClass> getAllClasses() {
...
-     * @return all sub-packages including nested sub-packages contained in this package,
-     * e.g. {@code [java.lang, java.lang.annotation, java.util, java.util.concurrent, ...]} for package {@code java}
-     * (compare {@link #getSubpackages()})
+     * @return all subpackages contained in the package tree of this package. I.e. all subpackages, subpackages
+     *         of subpackages, and so on. For package {@code java} this would be
+     *         {@code [java.lang, java.lang.annotation, java.util, java.util.concurrent, ...]} (compare {@link #getSubpackages()})
      */
     @PublicAPI(usage = ACCESS)
-    public Set<JavaPackage> getAllSubpackages() {
+    public Set<JavaPackage> getSubpackageTree() {
```
grimsa, line 339: Naming - my first thought when I read `getSubpackageTree` is "this returns a tree-like data structure", while it actually returns a `Set`.
My second thought is "ah, right, that's okay :)"

A few alternative options I can think of (not sure better or not):
- `getSupbackagesInTree`
- `getSupbackagesInPackageTree`
- `getSubpackagesRecursively`

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaPackageTest.java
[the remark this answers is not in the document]
grimsa: Agreed, makes sense.

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaPackageTest.java
```diff
@@ -241,77 +242,156 @@ public void visits_packages() {
...
+
+        assertThatDependencies(examplePackage.getPackage("third").getClassDependenciesFromThisPackage())
+                .isEmpty();
+
+        assertThatDependencies(examplePackage.getPackage("unrelated").getClassDependenciesFromThisPackage())
+                .containOnly(AnyClass.class, Object.class);
+    }
+
+    @Test
+    public void has_class_dependencies_from_this_package_and_subpackages() {
```
grimsa, line 263:
```suggestion
    public void has_class_dependencies_from_this_package_tree() {
```

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaPackageTest.java
```diff
@@ -241,77 +242,156 @@ public void visits_packages() {
...
+        assertThatDependencies(examplePackage.getPackage("third").getClassDependenciesToThisPackage())
+                .isEmpty();
+
+        assertThatDependencies(examplePackage.getPackage("unrelated").getClassDependenciesToThisPackage())
+                .isEmpty();
     }
 
     @Test
-    public void has_class_dependencies_from_other_packages() {
+    public void has_class_dependencies_to_this_package_and_subpackages() {
```
grimsa, line 303:
```suggestion
    public void has_class_dependencies_to_this_package_tree() {
```

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaPackageTest.java
```diff
@@ -241,77 +242,156 @@ public void visits_packages() {
...
+        assertThat(examplePackage.getPackage("third").getPackageDependenciesFromThisPackage())
                 .isEmpty();
+
+        assertThatPackages(examplePackage.getPackage("unrelated").getPackageDependenciesFromThisPackage())
+                .containOnlyNames("java.lang");
     }
 
     @Test
-    public void has_package_dependencies_to_other_packages() {
+    public void has_package_dependencies_from_this_package_and_subpackages() {
```
grimsa, line 342:
```suggestion
    public void has_package_dependencies_from_this_package_tree() {
```

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaPackageTest.java
```diff
@@ -241,77 +242,156 @@ public void visits_packages() {
...
+        assertThat(examplePackage.getPackage("first").getPackageDependenciesToThisPackage())
                 .containsOnly(
                         examplePackage.getPackage("second"),
                         examplePackage.getPackage("second.sub"),
                         examplePackage.getPackage("third.sub"));
 
-        assertThat(examplePackage.getPackage("third").getPackageDependenciesToSelf())
-                .containsOnly(examplePackage.getPackage("second.sub"));
+        assertThat(examplePackage.getPackage("second").getPackageDependenciesToThisPackage())
+                .doesNotContain(examplePackage.getPackage("second"));
```
grimsa, line 369:
```suggestion
        assertThat(examplePackage.getPackage("second").getPackageDependenciesToThisPackage())
                .containsOnly(
                        examplePackage.getPackage("first"),
                        examplePackage.getPackage("second.sub"));
```
This would probably be a stronger assertion

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaPackageTest.java
```diff
@@ -241,77 +242,156 @@ public void visits_packages() {
...
+
+        assertThat(examplePackage.getPackage("third").getPackageDependenciesToThisPackage())
+                .isEmpty();
+
+        assertThat(examplePackage.getPackage("unrelated").getPackageDependenciesToThisPackage())
+                .isEmpty();
+    }
+
+    @Test
+    public void has_package_dependencies_to_this_package_and_subpackages() {
```
grimsa, line 379:
```suggestion
    public void has_package_dependencies_to_this_package_tree() {
```

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaPackageTest.java
```diff
@@ -241,77 +242,156 @@ public void visits_packages() {
...
 
-        assertThat(examplePackage.getPackage("second").getPackageDependenciesToSelf())
+        assertThat(examplePackage.getPackage("first").getPackageDependenciesToThisPackageTree())
+                .containsOnly(
+                        examplePackage.getPackage("second"),
+                        examplePackage.getPackage("second.sub"),
+                        examplePackage.getPackage("third.sub"));
+
+        assertThat(examplePackage.getPackage("second").getPackageDependenciesToThisPackageTree())
                 .doesNotContain(examplePackage.getPackage("second"));
```
grimsa, line 389:
```suggestion
                .containsOnly(examplePackage.getPackage("first"));
```
Stronger assertion

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaPackageTest.java
grimsa: Now that you have fixed the bug mentioned here: https://github.com/TNG/ArchUnit/pull/922/files#r956754244 the assertion for `second` is also different, so probably it's okay as-is.

grimsa (commented):
The changes you made look excellent.
Added just a few minor comments.

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaPackage.java
codecholeric: Oh yeah, now that you say it I completely agree! I'm sold on `getSubpackagesInTree` :slightly_smiling_face: Because "subpackages" already sets the context, I think just using "tree" is fine then (I don't think there is any ambiguity; what other "tree" would I get my subpackages from :man_shrugging:)
I mean, likely `getSubpackagesRecursively` is the most precise one, but given that we now set a strong connotation with "package tree" I think it makes sense to still use "tree" here...

codecholeric:
Thanks a lot for your thorough review!! :smiley: And for finding all the places in test code where I overlooked replacing "package and subpackages" by "package tree" :see_no_evil:

codecholeric (approved):

Issue #182, opened 2019-06-16
Allow `@AnalyzeClasses` to be used as a meta annotation


bobtiernay-okta:
Having `@AnalyzeClasses` allowed to be a meta annotation would allow one to encapsulate repeating options across tests:

```java
@Target(ElementType.TYPE)
@Retention(RetentionPolicy.RUNTIME)
@AnalyzeClasses(packages = {PACKAGE_X, PACKAGE_Y}, importOptions = {DoNotIncludeTests.class, DoNotIncludeJars.class, DoNotIncludeArchives.class})
public @interface AnalyzeStandardClasses
```

For similar features in Spring, checkout https://docs.spring.io/spring/docs/current/spring-framework-reference/testing.html#integration-testing-annotations-meta.

codecholeric:
Sounds like a reasonable addition :smiley: Would be worth it to ponder about the implications if several `@AnalyzeClasses` come together then. I.e.

```
@AnalyzeUi
@AnalyzeService
class MyGeneralArchRules {
    @ArchTest
    static final ArchRule firstRule = // ...
}
```

Because up to now this scenario was impossible by design. I guess since all the members of `@AnalyzeClasses` are varargs anyway, it would be easy to just aggregate all `@AnalyzeClasses` within the hierarchy and add all the varargs together. This means addition of locations to import and intersection of `ImportOptions` though...
Also the use of two different cache modes would be a question to solve, who would win if one `@AnalyzeClasses` wants to clear the cache after the run and one wants to keep it? The safe thing would be to use the shorter one I guess.

I'll see if anyone wants to pitch in a PR here since I have to work on some other issues first.

Maiklins:
Is this something that can still be picked up? Then I would give it a go.

codecholeric:
Yes, I don't think anybody is actively working on this, so if you want to give it a shot I'm happy to support you where necessary :slightly_smiling_face:

mathze:
As it seems that @Maiklins didn't found the time to take over, I would be happy to give it a try.
Would it be ok for you @codecholeric?

mathze:
Didn't got any response so I assume no one is working at this. So I've created #1300

Maiklins:
@mathze I did not find the time to start on that one. So go ahead. :)

mathze:
Note: My PR only allows the plain use as meta annotation. Atm. it does not respect merging nor inheritance from super classes.
[...]

PR #932, opened 2022-08-02
Make AnalyzeClasses inherited

zsoltkatona:
I'd propose this change so a common configuration for the AnalyzeClasses annotation can be set in a common superclass for ArchUnit test classes within a project:
```
@AnalyzeClasses(packages = "com.example.service", importOptions = {DoNotIncludeTests.class, DoNotIncludeSomething.class})
@Target(ElementType.TYPE)
@Retention(RetentionPolicy.RUNTIME)
@Inherited
public @interface ArchRulesTest {
    ...
}

@ArchRulesTest
public class SecurityArchitecturePolicies {
    ...
}

@ArchRulesTest
public class MappingFrameworkPolicies {
    ...
}

@ArchRulesTest
public class LayeredArchitecturePolicies {
    ...
}
```
I haven't actually tested this, so this is just a recommendation - if you'd like me to fork the repo, create tests, etc. please let me know and I'll see when I'll be able to get to it.

### DIFF
--- a/archunit-junit/junit5/api/src/main/java/com/tngtech/archunit/junit/AnalyzeClasses.java
+++ b/archunit-junit/junit5/api/src/main/java/com/tngtech/archunit/junit/AnalyzeClasses.java
@@ -17,6 +17,7 @@ package com.tngtech.archunit.junit;
 
 import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
+import java.lang.annotation.Inherited;
 
 import com.tngtech.archunit.PublicAPI;
 import com.tngtech.archunit.core.domain.JavaClasses;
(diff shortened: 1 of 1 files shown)


codecholeric:
Just to clarify, are you proposing the same as suggested here: https://github.com/TNG/ArchUnit/issues/182 ?

Please be careful when using the word "inherit" :wink: Because it can also easily be understood in an OO way that you talk about object inheritance. You are talking about using `@AnalyzeClasses` as a meta-annotation if I understand you correctly?

codecholeric:
And unfortunately adding `@Inherited` will AFAIS not really contribute to this? Or maybe you can explain why you think that `@Inherited` (given the [respective docs](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/annotation/Inherited.html)) would have any influence on this?

codecholeric:
Gonna close this since there has been no activity for a long time. Feel free to reopen if you want to revisit this!

PR #954, opened 2022-09-10
make `importPackages("")` and `importClasspath(..)` consistent

codecholeric:
So far `ClassFileImporter.importPackages("")` would only return classes that returns, but not classes derived from the set classpath. The reason is that this corner case has been overlooked in `NormalizedResourceName.startsWith(..)`. I.e. if the prefix is empty, then the resource would only be included if it starts with a `/`, but resources in archives do not start with a `/`.


hankem (approved):
Looks reasonable, thanks!

Issue #923, opened 2022-07-20
FreezingArchRule breaks when executed from IDE and maven


mustaphazorgati:
**Environment**:
* JDK11

**Expected behavior**:
When the store files are generated, the tests should all pass independent of the test execution (here: IntelliJ or maven)

**Actual behavior**:
* When the store files were generated through IntelliJ, the tests only pass when they are executed through IntelliJ. Executing the tests through maven will result in test failures.
* When the store files were generated through maven, the tests only pass when they are executed through maven. Executing the tests through IntelliJ will result in test failures.

**How to reproduce**:
* Check out this PR: https://github.com/Taskana/taskana/pull/1971 where we've enabled our architecture tests (store files were generated through IntelliJ)
* Execute the ArchitectureTest class through IntelliJ. All tests should pass.
* Execute the ArchitectureTest class through maven. Some tests should fail. 
Execution command (from project root directory): 
```sh
#build dependencies
./mvnw clean install -pl :taskana-core-test -am -DskipTests
#execute tests from ArchitectureTest class
./mvnw clean verify -pl :taskana-core-test -Dtest="ArchitectureTest"
```
**Further notes**:

As of right now we've disabled the tests, since this behavior is blocking us. 
You can find our test class here: https://github.com/Taskana/taskana/blob/master/lib/taskana-core-test/src/test/java/acceptance/ArchitectureTest.java
The tests packagesShouldBeFreeOfCyclicDependencies and classesShouldBeFreeOfCyclicDependencies will fail.
We've already set the following properties to an unreasonable high amount in order to make sure that all errors are detected.
```
cycles.maxNumberToDetect=100000
cycles.maxNumberOfDependenciesPerEdge=10000
```

During our initial debugging we've found out that the line numbers differ between the execution through maven and IntelliJ. Furthermore we've seen that IntelliJ and maven name lambda's differently, which is a second difference in the store files.

![Screenshot from 2022-07-18 16-18-53](https://user-images.githubusercontent.com/15628173/179991914-e5f0cea6-f189-4a97-be72-6b47c88ca5c7.png)
![Screenshot from 2022-07-18 16-18-30](https://user-images.githubusercontent.com/15628173/179991922-5f259f6a-2443-46d3-bc87-7e0cb0668ca7.png)

codecholeric:
[...]

PR #957, opened 2022-09-16
fix bug where origins of synthetic method calls override each other

codecholeric:
Calls to lambda methods `lambda$abc$123` seem to have unique origins with current JDK versions. I.e. even two identical consecutive lambda invocations will lead to two separate synthetic methods. Unfortunately, this is not the case for synthetic `access$123` methods (to circumvent private modifiers), which lead to a bug where multiple origins of such a call would override each other. 
This lead e.g. to the behavior in https://github.com/TNG/ArchUnit/issues/923, where the application of AspectJ introduced an `access$123` method that was called from multiple places. But, only one such call would randomly be created during the import, which lead to an unpredictable behavior of which cycles between outer and inner classes would be detected. 
We now record multiple access records for any synthetic call and trace back all those records instead of just the last one. This way all such synthetic accesses should be imported correctly now.


hankem (approved):
LGTM!

PR #958, opened 2022-09-18
Consolidate `ClassFileImporter.importClasspath(..)` with other APIs

codecholeric:
All other import methods behave consistently by taking into account all the `ImportOptions` that have been added via `withImportOption(..)` and then only controlling where to look for class files by the final `import...()` method. Only `importClasspath()` transparently adds an `ImportOption` to exclude archives from the import. This creates a surprising API, also that the only way to really import the whole classpath then is to use `importClasspath(emptySet())` with an empty set of `ImportOptions` passed to the `importClasspath(..)` method.
We now consolidate these APIs to make `importClasspath()` behave exactly like the other methods. We remove the version `importClasspath(Collection<ImportOption> options)` as there is no advantage now over using `withImportOptions(options).importClasspath()` like with all other import methods. Right now is the best moment to do this breaking change, since we are about to release version `1.0.0`.

### DIFF
--- a/docs/userguide/006_The_Core_API.adoc
+++ b/docs/userguide/006_The_Core_API.adoc
@@ -26,7 +26,7 @@ specified as URLs or as JAR files.
 
 Furthermore specific locations can be filtered out, if they are contained in the source of classes,
 but should not be imported. A typical use case would be to ignore test classes, when the classpath
-is imported. This can be achieved by specifying `ImportOptions`:
+is imported. This can be achieved by specifying `ImportOption`﻿s:
 
 [source,java,options="nowrap"]
 ----
@@ -47,7 +47,7 @@ A `Location` is principally an URI, i.e. ArchUnit considers sources as File or J
 
 For the two common cases to skip importing JAR files and to skip importing test files
 (for typical setups, like a Maven or Gradle build),
-there already exist predefined `ImportOptions`:
+there already exist predefined `ImportOption`﻿s:
 
 [source,java,options="nowrap"]
 ----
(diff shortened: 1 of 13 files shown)


docs/userguide/006_The_Core_API.adoc
```diff
@@ -284,7 +284,7 @@ object from the Reflection API. For example:
 
 [source,java,options="nowrap"]
 ----
-JavaClasses classes = new ClassFileImporter().importClasspath(new ImportOptions());
+JavaClasses classes = new ClassFileImporter().importClasspath();
```
hankem, line 287: L29 and L50, as well as `009_JUnit_Support.adoc`'s L107, still refer to
> `ImportOptions`

which could IMO be changed to
> `ImportOption`s

hankem (approved):
The old version of `new ClassFileImporter().importClasspath()` was indeed strange. Thanks for making this consistent!

docs/userguide/006_The_Core_API.adoc
codecholeric: Hmm, yes, I was pondering about it. Thought it also looks weird to have the 's' outside (and I think in other Javadocs, etc., I sometimes put the whole plural into the code as well), but maybe it's better to change it :thinking:

hankem: A JavaDoc link is different because it unambigiously targets the `ImportOption` class.

But in the documentation, you don't have a link. When I see "`ImportOptions`" there, I'd assume that there is a class (or some other code construct) `ImportOptions` and be confused (if not even misled when I find the package-private class `ImportOptions`). I'd use "`ImportOption`s" to separate which part is code (`ImportOption`) and which part is grammar ("-s").

PR #959, opened 2022-09-18
move package `library.plantuml` -> `library.plantuml.rules`

codecholeric:
... to free up the namespace for further possible enhancements around PlantUML. Our upcoming `1.0.0` release is the perfect point in time to do this breaking change. This will allow us to also add some infrastructure to create diagrams in the future within the `library.plantuml` package and make it clear that all existing code has the context of creating rules from PlantUML.

### DIFF
--- a/archunit-integration-test/src/test/java/com/tngtech/archunit/PublicAPIRules.java
+++ b/archunit-integration-test/src/test/java/com/tngtech/archunit/PublicAPIRules.java
@@ -22,6 +22,7 @@ import com.tngtech.archunit.lang.ArchRule;
 import com.tngtech.archunit.lang.CompositeArchRule;
 import com.tngtech.archunit.lang.ConditionEvents;
 import com.tngtech.archunit.lang.SimpleConditionEvent;
+import com.tngtech.archunit.lang.conditions.ArchPredicates;
 import com.tngtech.archunit.lang.syntax.ArchRuleDefinition;
 
 import static com.google.common.collect.Iterables.getLast;
@@ -42,8 +43,9 @@ import static com.tngtech.archunit.core.domain.properties.CanBeAnnotated.Predica
 import static com.tngtech.archunit.core.domain.properties.HasModifiers.Predicates.modifier;
 import static com.tngtech.archunit.core.domain.properties.HasName.Utils.namesOf;
 import static com.tngtech.archunit.lang.SimpleConditionEvent.violated;
+import static com.tngtech.archunit.lang.conditions.ArchConditions.beAnnotatedWith;
+import static com.tngtech.archunit.lang.conditions.ArchConditions.have;
 import static com.tngtech.archunit.lang.conditions.ArchPredicates.are;
-import static com.tngtech.archunit.lang.conditions.ArchPredicates.have;
 import static com.tngtech.archunit.lang.conditions.ArchPredicates.is;
 import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.classes;
 import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.codeUnits;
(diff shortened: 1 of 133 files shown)


archunit/src/main/java/com/tngtech/archunit/library/plantuml/rules/PlantUmlArchCondition.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.tngtech.archunit.library.plantuml;
+package com.tngtech.archunit.library.plantuml.rules;
```
hankem, line 17: 1. Should `class PlantUmlArchCondition` itself be declared as `@PublicAPI`?
2. Is `..rules` really a good package name for a _condition_?

hankem (approved):

archunit/src/main/java/com/tngtech/archunit/library/plantuml/rules/PlantUmlArchCondition.java
codecholeric: 1. Honestly, I never established a clear convention for this :thinking: It implicitly is, because it has `@PublicAPI` members, but maybe it would be better from a user's perspective to make the class itself also `@PublicAPI` :thinking: But then it should maybe be done for all classes, because I doubt it's the only candidate :man_shrugging: 
2. My intention was to create one bucket for everything related to creating rules from PlantUML. I didn't really want to orient the package name so much on what we have right now (right now we only have a condition), but more on the vision for this package (everything about PlantUML rules). My vision is to put something in there to also create PlantUML diagrams, and then the distinction between "diagram_creation"/"rendering"(whatever we'll call it) and "rules" would make sense to me, as a rough splitting into these two different semantic parts.

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTest.java
```diff
@@ -2163,12 +2157,6 @@ public boolean matches(JavaMember value) {
         };
     }
 
-    private static JavaClass fakeClassWithPackage(String pkg) {
-        JavaClass javaClass = mock(JavaClass.class);
-        when(javaClass.getPackageName()).thenReturn(pkg);
-        return javaClass;
-    }
-
```
hankem, line 2171: Did you intend to add this cleanup to the following commit?
> mark classes that are part of the public API as `@PublicAPI`

archunit-integration-test/src/test/java/com/tngtech/archunit/PublicAPIRules.java
```diff
@@ -66,6 +69,19 @@ public class PublicAPIRules {
...
 
+    @ArchTest
+    public static final ArchRule public_classes_of_public_API_should_be_annotated_with_PublicAPI =
+            classes()
+                    .that(haveMemberThatBelongsToPublicApi())
+                    .should(beAnnotatedWith(PublicAPI.class).<JavaClass>forSubtype()
+                            .or(haveSupertype(annotatedWith(PublicAPI.class)).and(ArchConditions.not(have(modifier(PUBLIC))))));
+
+    private static ArchCondition<JavaClass> haveSupertype(DescribedPredicate<? super JavaClass> predicate) {
+        return ArchConditions.have(DescribedPredicate.describe(
```
hankem, line 80: You could use the static import of `ArchConditions.have`.

hankem (approved):
Thanks for explicitly declaring `@PublicAPI` at the class level! 💙

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaCodeUnitReference.java
```diff
@@ -23,15 +23,17 @@
...
+@PublicAPI(usage = ACCESS)
 public abstract class JavaCodeUnitReference<T extends CodeUnitReferenceTarget> extends JavaCodeUnitAccess<T> {
     JavaCodeUnitReference(JavaAccessBuilder<T, ?> builder) {
         super(builder);
     }
 
     /**
      * Predefined {@link DescribedPredicate predicates} targeting {@link JavaCodeUnitReference}.
-     * Further predicates to be used with {@link JavaCodeUnitReference} can be found at {@link JavaCodeUnitAccess.Predicates}.
+     * Further, predicates to be used with {@link JavaCodeUnitReference} can be found at {@link JavaCodeUnitAccess.Predicates}.
```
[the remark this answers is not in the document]
codecholeric: Oh yeah :see_no_evil: IntelliJ auto-"correct" :grimacing:

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTest.java
codecholeric: Yes, because making `JavaClass` `final` was the reason to do this. But I guess I could also split it out (or motivate it in the commit message).

codecholeric: I split it into a separate commit so it's easier to understand the motivation :slightly_smiling_face:

Issue #960, opened 2022-09-20
Classes missing from PlantUML diagram should not cause exception


codecholeric:
At the moment, when using the PlantUML feature to create `ArchRules` from PlantUML component diagrams every class must be contained in the diagram or an `IllegalStateException` is thrown. The reason behind this is that we want to prevent users from missing classes or creating tests that in fact only cover a little part of the code base while getting a false sense of security.

This has one severe problem though: In legacy code bases there might be many classes that **should** be contained by the diagram, but are not at the moment. Getting an `IllegalStateException` thrown prevents the use of `FreezingArchRule` to freeze this legacy code state and improve it piece by piece. The only way would be to either create a legacy bucket in the diagram, which is tedious, or to exclude all the classes from the rule creating a big black hole where classes are uncovered by the rule and can further degenerate.

We should change classes not contained in the diagram from an `IllegalStateException` to a regular rule violation to improve this.

PR #964, opened 2022-09-25
Replace PlantUML rule exceptions by violations

codecholeric:
So far, when using `PlantUmlArchCondition`, if a class is contained in no component or multiple components an exception is thrown. This makes it impossible to freeze the current state with `FreezingArchRule`, because the test will simply crash hard.
We now replace these exceptions by regular rule violations. That way they can be treated like any other architecture violation and can also be frozen with `FreezingArchRule`.

Resolves: #960


archunit/src/main/java/com/tngtech/archunit/library/plantuml/rules/PlantUmlArchCondition.java
```diff
@@ -159,20 +162,38 @@ public PlantUmlArchCondition ignoreDependencies(DescribedPredicate<? super Depen
...
+            events.add(violated(javaClass, String.format("Class %s is not contained in any component", javaClass.getName())));
+            return;
+        }
+
+        Set<PlantUmlComponent> components = javaClassDiagramAssociation.getAssociatedComponents(javaClass);
+        if (components.size() > 1) {
+            events.add(violated(javaClass, String.format(
+                    "Class %s may not be contained in more than one component, but is contained in %s",
+                    javaClass.getName(),
+                    components.stream().map(it -> it.getComponentName().asString()).sorted().collect(toList()))));
```
hankem, line 180: Do we really want to rely on the `toString()` of the `List` built by `Collectors.toList()`?
I'd probably build the `String` explicitly:
```suggestion
            String prefix = String.format(
                    "Class %s may not be contained in more than one component, but is contained in [",
                    javaClass.getName());
            String message = components.stream().map(it -> it.getComponentName().asString()).sorted()
                    .collect(joining(", ", prefix, "]"));
            events.add(violated(javaClass, message));
```

archunit/src/test/java/com/tngtech/archunit/library/plantuml/rules/JavaClassDiagramAssociationTest.java
```diff
@@ -57,19 +56,6 @@ public void get_target_package_identifiers_of_class() {
...
     }
 
-    @Test
-    public void rejects_class_not_contained_in_any_component() {
-        JavaClassDiagramAssociation javaClassDiagramAssociation = createAssociation(TestDiagram.in(temporaryFolder)
-                .component("SomeComponent").withStereoTypes("..someStereotype.")
-                .write());
-        JavaClass classNotContained = importClassWithContext(Object.class);
-
-        thrown.expect(IllegalStateException.class);
```
hankem, line 67: We could keep the test case, but document the new behavior that a `NoSuchElementException` is thrown.

archunit/src/test/java/com/tngtech/archunit/library/plantuml/rules/JavaClassDiagramAssociationTest.java
```diff
@@ -82,22 +68,6 @@ public void reports_if_class_is_contained_in_any_component() {
...
 
-    @Test
-    public void class_resides_in_multiple_packages() {
-        JavaClassDiagramAssociation javaClassDiagramAssociation = createAssociation(TestDiagram.in(temporaryFolder)
-                .component("A").withStereoTypes("..foopackage..")
-                .component("B").withStereoTypes("..barpackage")
-                .write());
-        JavaClass classContainedInTwoComponents = importClassWithContext(ClassInFooAndBarPackage.class);
-
-        thrown.expect(ComponentIntersectionException.class);
```
hankem, line 93: We could keep the test case, but document the new behavior that a `IllegalArgumentException` is thrown.

hankem (approved):

archunit/src/main/java/com/tngtech/archunit/library/plantuml/rules/PlantUmlArchCondition.java
codecholeric: Hmm, I consciously did this because
a) I think the JDK's `Collection.toString()` methods have been very stable and consistent ... forever. I.e. I would trust the JDK to not suddenly change such a central API.
b) we have tests that run on each JDK version, so even if this would change we would notice it
But, on the other hand I agree that it's more explicit to read, if we do it ourselves :man_shrugging: So in the end no super strong opinion, I'll just go with your comment then :wink: (and on the other hand, the JDK did suddenly change `Annotation.toString()` in some version :joy:)

PR #965, opened 2022-09-26
Some minor code improvements

hankem:
Contributes some random code improvements for minor _"issues"_ detected by the IDE:
* make some fields final
* make some inner classes static
* simplify code: remove redundant interface definition/cast/;, !anyMatch -> noneMatch
* remove useless throws declarations
* update Mockito and use mockito-junit-jupiter's MockitoExtension
* replace deprecated `com.google.common.io.Files` API
* replace `assertThat(x).containsOnlyElementsOf` with `.hasSameElementsAs`
* replace deprecated JUnit 4 `ExpectedException`


build.gradle
```diff
@@ -60,7 +60,8 @@ ext {
             junitPlatformEngine : [group: 'org.junit.platform', name: 'junit-platform-engine', version: '1.8.2'],
             hamcrest            : [group: 'org.hamcrest', name: 'hamcrest-core', version: '1.3'],
             junit_dataprovider  : [group: 'com.tngtech.java', name: 'junit-dataprovider', version: '1.11.0'],
-            mockito             : [group: 'org.mockito', name: 'mockito-core', version: '4.4.0'],
+            mockito             : [group: 'org.mockito', name: 'mockito-core', version: '4.6.1'],
+            mockito_junit5      : [group: 'org.mockito', name: 'mockito-junit-jupiter', version: '4.6.1'],
```
hankem, line 64: FYI: A version ≥`4.7.0` broke the build with
> Execution failed for task ':archunit:compileJdk9testJava'.
> java.time.DateTimeException: Invalid value for MonthOfYear (valid values 1 - 12): 0

(which sounds like [a JDK bug](https://bugs.openjdk.org/browse/JDK-8184940))

archunit-junit/junit5/engine/src/test/java/com/tngtech/archunit/junit/internal/ArchUnitTestEngineTest.java
```diff
@@ -116,6 +118,7 @@
 import static org.mockito.Mockito.when;
 
 @ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.WARN)
```
hankem, line 121: As we only have one single usage of `MockitoExtension`, I wonder whether we shouldn't just call `MockitoAnnotations.initMocks(testInstance);` directly (and drop the `mockito-junit-jupiter` dependency again).

codecholeric: Hmm, yeah, sounds fair :thinking: Seems overkill to have an extra dependency for that and also not really saving anything in this case :grimacing:

codecholeric: On second thought, `initMocks(..)` is meanwhile deprecated, you should now call `openMocks(..)` and hold the returned `AutoCloseable` to close it in the tear down method :zany_face:
Once we migrate the other code from JUnit 4 to JUnit 5 we'll also need this in more places, so let's just keep the dependency for now...

codecholeric (approved):

codecholeric:
Thanks a lot :star_struck:

PR #968, opened 2022-10-03
Rename `JavaPackage.getAllClasses()` -> `getClassesInPackageTree()`

codecholeric:
In all other places we have replaced the fuzzy word "all" by "in package tree" to clarify the semantics. The Javadoc already hinted at the fact that this method name should be consolidated. Now with release 1.0.0 we can break the API and don't need to deprecate the old method. I decided to call the method `getClassesInPackageTree()`, even though the similar method `getSubpackagesInTree()` does not add the infix "package", because I think for the former it is worth it to clarify the context, while the latter already talks about packages, so "tree" is already clear from the context.

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaPackage.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaPackage.java
@@ -303,7 +303,7 @@ public final class JavaPackage implements HasName, HasAnnotations<JavaPackage> {
     }
 
     /**
-     * @return all classes directly contained in this package, but not classes in the lower levels of the package tree (compare {@link #getAllClasses()})
+     * @return all classes directly contained in this package, but not classes in the lower levels of the package tree (compare {@link #getClassesInPackageTree()})
      */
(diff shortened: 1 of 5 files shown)


archunit/src/main/java/com/tngtech/archunit/core/domain/JavaPackage.java
```diff
@@ -303,21 +303,22 @@ private void setParent(JavaPackage parent) {
...
+     * @return all classes directly contained in this package, but not classes in the lower levels of the package tree (compare {@link #getClassesInPackageTree()})
      */
     @PublicAPI(usage = ACCESS)
     public Set<JavaClass> getClasses() {
         return classes;
     }
 
     /**
-     * @return all classes contained in this package tree (compare {@link #getClasses()})
+     * @return all classes contained in this {@link JavaPackage package tree}. I.e. all classes in this package,
```
hankem, line 314:
```suggestion
     * @return all classes contained in this {@link JavaPackage package tree}, i.e. all classes in this package,
```

hankem (approved):

PR #970, opened 2022-10-03
add general coding rules for assertion messages and deprecated APIs

### LINKED ISSUE #963: Can ArchUnit check that assertions include a message?
I ran into a situation today where our production code hit an assertion without a message, resulting in an `AssertionError` being thrown e.g.:

```java
String text = "Hello";
assert text.isEmpty();
```

Clearly this would have been easier to diagnose if the assertion had a n associated message e.g.:

```java
String text = "Hello";
assert text.isEmpty() : "Text should have been empty: " + text;
```

A couple of us immediately thought "We should create an ArchUnit test to require that all assertions statements include a message", but I can't see how this would be expressed using ArchUnit. 

Is this even possible in ArchUnit?

Or is this not really considered architectural and I should be looking somewhere like SpotBugs instead?

hankem:
This PR adds two more `GeneralCodingRules`:
* `ASSERTIONS_SHOULD_HAVE_DETAIL_MESSAGE` resolves #963
* `DEPRECATED_API_SHOULD_NOT_BE_USED` checks that no class accesses members (i.e. calls methods or constructors, or accesses fields) that are either `@Deprecated` themselves or declared in `@Deprecated` classes, or references `@Deprecated` class objects.

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/library/GeneralCodingRules.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/GeneralCodingRules.java
@@ -480,4 +480,22 @@ public final class GeneralCodingRules {
             }
         };
     }
+
+    /**
+     * A rule that checks that all {@link AssertionError AssertionErrors} (e.g. from the {@code assert} keyword) have a detail message.
+     * <p>
+     * Example:
+     * <pre>{@code
+     * assert x > 0;  // violation
+     * throw new AssertionError();  // violation
+     *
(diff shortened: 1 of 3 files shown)


codecholeric (approved):
Looks good, thanks a lot!! :slightly_smiling_face: I'm gonna merge it even though the JDK 10 test fails, because as discussed I think this is some problem with the test environment. The same test runs fine on my local Ubuntu machine with the provisioned JDK :unamused:

PR #978, opened 2022-10-07
Fix userguide regarding DescribedPredicate

thmuch:
DescribedPredicate had an `apply` method that was renamed to `test` in 1.0.

This PR renames the only mention of DescribedPredicate#apply() in the user guide.

Signed-off-by: Thomas Much <github@snailshell.de>


hankem:
Thank you so much for spotting this and for the fix!

Issue #298, opened 2020-01-22
Fail to detect cycles when the depending variable is `final` or in interface


ceachother:
I'm at `0.13.0`.

When I have 2 class:
```
public class ClassA {
    public static final String TEST = "test";
    public void test() {
        System.out.println(ClassB.TEST);
    }
}

public class ClassB {
    public static final String TEST = "test";
    public void test() {
        System.out.println(ClassA.TEST);
    }
}
```
The`slices().matching("pkg.(**)..").should().beFreeOfCycles();` won't report any violations.

Is this a bug or you guys made the decision intentionally?

hankem:
ArchUnit currently does not detect such a dependency because it is not (so directly) present in the bytecode: a `static final` primitive or `String` is just inlined in the constant pool of the class.

When I change your example constants to `testA` / `testB` to make them distinct,
`javap -c- p -v ClassA.class` shows 
<details><summary>this bytecode</summary>

```
public class ClassA
  minor version: 0
  major version: 55
  flags: (0x0021) ACC_PUBLIC, ACC_SUPER
  this_class: #6                          // ClassA
  super_class: #7                         // java/lang/Object
  interfaces: 0, fields: 1, methods: 2, attributes: 1
Constant pool:
   #1 = Methodref          #7.#19         // java/lang/Object."<init>":()V
   #2 = Fieldref           #20.#21        // java/lang/System.out:Ljava/io/PrintStream;
   #3 = Class              #22            // ClassB
   #4 = String             #23            // testB
   #5 = Methodref          #24.#25        // java/io/PrintStream.println:(Ljava/lang/String;)V
   #6 = Class              #26            // ClassA
   #7 = Class              #27            // java/lang/Object
   #8 = Utf8               TEST
   #9 = Utf8               Ljava/lang/String;
  #10 = Utf8               ConstantValue
  #11 = String             #28            // testA
  #12 = Utf8               <init>
  #13 = Utf8               ()V
  #14 = Utf8               Code
  #15 = Utf8               LineNumberTable
  #16 = Utf8               test
  #17 = Utf8               SourceFile
  #18 = Utf8               ClassA.java
  #19 = NameAndType        #12:#13        // "<init>":()V
  #20 = Class              #29            // java/lang/System
  #21 = NameAndType        #30:#31        // out:Ljava/io/PrintStream;
  #22 = Utf8               ClassB
  #23 = Utf8               testB
  #24 = Class              #32            // java/io/PrintStream
[...]

PR #999, opened 2022-11-08
recognize empty records

### LINKED ISSUE #998: Java 14 records with empty constructors violate beRecords() rule
Given that I have an arch unit test like this:

```
package org.example;

import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.classes;

import com.tngtech.archunit.core.importer.ClassFileImporter;
import org.junit.jupiter.api.Test;

public class RecordTest
{

  @Test
  public void recordTest() {
    var javaClasses = new ClassFileImporter().importPackages("org.example");

    classes().that()
        .haveNameMatching("(.*)Record")
        .should()
        .beRecords()
        .because("they should be simple data structures")
        .check(javaClasses);
  }
}

```

And a record like this in my project:

```
package org.example;

public record EmptyRecord()
{
}
```

When I run the arch unit test, then a violation is reported, even though `EmptyRecord` is a valid (but useless) Java 14 record:

```
Architecture Violation [Priority: MEDIUM] - Rule 'classes that have name matching '(.*)Record' should be records, because they should be simple data structures' was violated (1 times):
Class <org.example.EmptyRecord> is no record in (EmptyRecord.java:0)
java.lang.AssertionError: Architecture Violation [Priority: MEDIUM] - Rule 'classes that have name matching '(.*)Record' should be records, because they should be simple data structures' was violated (1 times):
Class <org.example.EmptyRecord> is no record in (EmptyRecord.java:0)
	at com.tngtech.archunit.lang.ArchRule$Assertions.assertNoViolation(ArchRule.java:94)
	at com.tngtech.archunit.lang.ArchRule$Assertions.check(ArchRule.java:86)
	at com.tngtech.archunit.lang.ArchRule$Factory$SimpleArchRule.check(ArchRule.java:165)
	at org.example.RecordTest.recordTest(RecordTest.java:20)
```

When the record is extended with a parameter, then the violation is resolved:

```
package org.example;

public record EmptyRecord(String someValue)
{
}
```

A full example to reproduce the issue can be found [here](https://github.com/SoylentBob/arch-unit-record-bug).

hankem:
As `ClassVisitor#visitRecordComponent` is not invoked for empty records (without components), we better rely on the access flags provided to `ClassVisitor#visit`.

resolves #998

### DIFF
--- a/archunit-integration-test/src/test/java/com/tngtech/archunit/ImporterRules.java
+++ b/archunit-integration-test/src/test/java/com/tngtech/archunit/ImporterRules.java
@@ -9,8 +9,10 @@ import com.tngtech.archunit.junit.ArchTest;
 import com.tngtech.archunit.lang.ArchRule;
 
 import static com.tngtech.archunit.ArchUnitArchitectureTest.THIRDPARTY_PACKAGE_IDENTIFIER;
+import static com.tngtech.archunit.base.DescribedPredicate.doNot;
 import static com.tngtech.archunit.base.DescribedPredicate.not;
 import static com.tngtech.archunit.core.domain.JavaClass.Predicates.belongToAnyOf;
+import static com.tngtech.archunit.core.domain.JavaClass.Predicates.resideOutsideOfPackage;
 import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.noClasses;
 
 public class ImporterRules {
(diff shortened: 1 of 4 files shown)


codecholeric (approved):
Thanks a lot for tackling this :smiley: I actually realized that I originally coded `JavaModifier` a little bit shady :see_no_evil: Because the idea was, that domain in general does not depend on ASM at all (I made one conscious exception for `JavaClassDescriptor` at the time, for convenience reasons, maybe I should also revise that at some point :thinking:). Anyway, I think the dependency on the ASM opcodes crept into `JavaModifier` by accident :grimacing: Thus, I added one little commit adding an additional arch test and deprecating the public API in `JavaModifier` that takes "ASM access" flags (I think this actually makes no sense at all in hindsight). Does this look reasonable to you?
Other than that all your changes make sense from my point of view, given the current state :slightly_smiling_face: It would be nice if we would only need to check `(asmAccess & Opcodes.ACC_RECORD) != 0` once, instead of two separate places, but I see no easy way to do this with the current state and without introducing and additional `JavaModifier.RECORD` which likely makes no sense :shrug: So maybe we can just revise this when we remove the deprecated public API in the future and move the handling of ASM access flags into the import context again.

hankem:
Yes, I agree: `JavaModifier.getModifiersForClass(int asmAccess)` is an unexpected public API (but I doubt that users could make use of it).
I had also briefly thought about `JavaModifier.RECORD`, but as this is [just an ASM specific access flag](https://gitlab.ow2.org/asm/asm/-/blob/ASM_9_4/asm/src/main/java/org/objectweb/asm/Opcodes.java#L328-333), I also decided against it.

TL;DR: 👍

archunit-integration-test/src/test/java/com/tngtech/archunit/ImporterRules.java
```diff
@@ -3,22 +3,39 @@
...
+    public static final ArchRule asm_is_only_used_in_importer_or_JavaClassDescriptor =
+            noClasses()
+                    .that(
+                            resideOutsideOfPackage("..core.importer..")
+                                    .and(doNot(belongToAnyOf(
+                                            JavaClassDescriptor.class,
+                                            // whenever we get the chance to break the public API
+                                            // we should likely remove the dependencies from JavaModifier to ASM.
+                                            // Those crept in by accident at some point because the design was convenient.
+                                            JavaModifier.class)))
```
hankem, line 36: We might not even have to exclude `JavaModifier` as long as we only import static `int` constants, cf. https://github.com/TNG/ArchUnit/issues/298#issuecomment-577038056.

codecholeric: Okay, as agreed I'm only gonna add the comment there, but not add `JavaModifier` to avoid further dependencies slipping in :slightly_smiling_face:

PR #1009, opened 2022-11-27
Reduce allocations in NormalizedResourceName

dreis2211:
Hi,

I've been doing some profiling of some test suites lately that use ArchUnit. While the majority of allocations are done in the processing of classes (unsurprisingly), there are also some caused by `NormalizedResourceName` (mostly from actually normalizing the name) that seem avoidable.

I've put together a little benchmark comparison of the current approach (old), one with precompiled patterns and the approach proposed in this PR (new).
```
Benchmark                                                          (input)  Mode  Cnt     Score     Error   Units
MyBenchmark.testNew                                                com/foo  avgt   10     8,530 ±   0,115   ns/op
MyBenchmark.testNew:·gc.alloc.rate.norm                            com/foo  avgt   10    ≈ 10⁻⁶              B/op

MyBenchmark.testNew                                              /com/foo/  avgt   10    35,633 ±   0,537   ns/op
MyBenchmark.testNew:·gc.alloc.rate.norm                          /com/foo/  avgt   10    96,000 ±   0,001    B/op

MyBenchmark.testOld                                                com/foo  avgt   10   573,092 ±   9,175   ns/op
MyBenchmark.testOld:·gc.alloc.rate.norm                            com/foo  avgt   10  1752,000 ±   0,001    B/op

MyBenchmark.testOld                                              /com/foo/  avgt   10   580,875 ±  18,804   ns/op
MyBenchmark.testOld:·gc.alloc.rate.norm                          /com/foo/  avgt   10  1784,000 ±  38,247    B/op

MyBenchmark.testPrecompiledPattern                                 com/foo  avgt   10   305,113 ±   5,875   ns/op
MyBenchmark.testPrecompiledPattern:·gc.alloc.rate.norm             com/foo  avgt   10   720,000 ±   0,001    B/op

MyBenchmark.testPrecompiledPattern                               /com/foo/  avgt   10   330,982 ±  19,807   ns/op
MyBenchmark.testPrecompiledPattern:·gc.alloc.rate.norm           /com/foo/  avgt   10   752,000 ±  38,247    B/op
```

As you can see, the provided approach in this PR avoids allocations completely in case no stripping of trailing and leading slashes needs to be done, but is also considerably faster when it is done.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/importer/NormalizedResourceName.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/importer/NormalizedResourceName.java
@@ -21,13 +21,36 @@ class NormalizedResourceName {
     private final String resourceName;
 
     private NormalizedResourceName(String resourceName) {
-        this.resourceName = resourceName
-                // normalize Windows backslashes
-                .replace('\\', '/')
-                // remove leading slashes
-                .replaceAll("^/*", "")
-                // remove trailing slashes
-                .replaceAll("/*$", "");
+        this.resourceName = normalizeName(resourceName);
+    }
+
+    private String normalizeName(String resourceName) {
+        // normalize Windows backslashes
+        resourceName = resourceName.replace('\\', '/');
+        // remove leading slashes
+        resourceName = removeLeadingSlashes(resourceName);
+        // remove trailing slashes
+        return removeTrailingSlashes(resourceName);
+    }
+
+    private String removeLeadingSlashes(String s) {
+        int index;
+        for (index = 0; index < s.length(); index++) {
+            if (s.charAt(index) != '/') {
+                break;
+            }
+        }
+        return s.substring(index);
+    }
+
+    private String removeTrailingSlashes(String s) {
+        int index;
+        for (index = s.length() - 1; index >= 0; index--) {
+            if (s.charAt(index) != '/') {
+                break;
+            }
+        }
+        return s.substring(0, index + 1);
     }
 
     static NormalizedResourceName from(String resourceName) {
@@ -44,7 +67,7 @@ class NormalizedResourceName {
 
     private boolean isAncestorPath(NormalizedResourceName prefix) {
         return resourceName.startsWith(prefix.resourceName) &&
-                resourceName.substring(prefix.resourceName.length()).startsWith("/");
+                resourceName.startsWith("/", prefix.resourceName.length());
     }
 
     /**


codecholeric:
Thanks a lot for your contribution and profiling this! :smile: I'm always happy about things that speed up ArchUnit. Can you maybe elaborate a little how exactly you profiled this? Because, while your changes make sense to me, that they would be faster (after all Regex matching is a quite expensive operation in comparison), when I profile it with a real use case (e.g. just importing the whole classpath which takes over 20 seconds on my machine) I don't notice any real difference :thinking:
I'm just wondering if this might be premature optimization that doesn't really move the needle, because the bottleneck is actually in a different place :thinking: Do you have any profiling of a real import where this makes a difference for you?

dreis2211:
@codecholeric I've fired up async-profiler against my test-suites.

This is definitely on the micro-optimization end. In our test-suites ~0,5-1% of allocations coming from this. (Notice the lower right corner)

image

I don't know if you notice the tiny purple highlights that are sprinkled through the flamegraph. E.g. one of them looks like this when zoomed in:

image

It's not about bottleneck optimizations or speed-ups here, but rather to avoid allocations that can be simply avoided.

codecholeric:
But it also has a trade-off, right? Before I can read the code in 3 lines, it's pretty concise. After I have these verbose functions that take quite long to read. I'm sure there are many micro optimizations that could get a couple of ms at the cost of readability, but I'm not convinced this makes sense then from the point of maintainability :thinking: The JVM is just pretty good at optimizing code, and keeping the more concise, high-level construct also can benefit from future optimizations when the JVM evolves :man_shrugging: 
So, I think if we increase the code size and decrease maintainability, then we should have a real use case benchmark where we see the benefit :thinking:
We could maybe go with the change of `isAncestorPath`, but now, looking at it again and following the same argumentation, I would maybe just rewrite it as

```
private boolean isAncestorPath(NormalizedResourceName prefix) {
  return resourceName.startsWith(prefix.resourceName + "/");
}
```

:joy: Again, more allocations, but profiling the import I don't see a difference and I think this actually reads easier :man_shrugging:

dreis2211:
I don't see how the new code is unreadable, really. One could think of extracting the two methods into a small utility class and then the code looks even better and nobody needs to understand the regexes like before.
```
       // normalize Windows backslashes
        resourceName = resourceName.replace('\\', '/');
        // remove leading slashes
        resourceName = StringUtils.removeLeadingSlashes(resourceName);
        // remove trailing slashes
        return StringUtils.removeTrailingSlashes(resourceName);
```

And the new code isn't really preventing JVM optimizations. Stuff like loop unrolling for example is what could still be done here for example. The methods are also small enough to be inlined. But in the end, you maintain it. 

Our usage of ArchUnit is pretty limited. From a gut feeling of 1000 tests we have maybe 10 ArchUnit tests. Given that we already reach 0,5-1% with the very limited use, I contributed this. Again - it's not about speedups, it's about reducing allocations where possible to reduce GC impact (that is then consuming CPU cycles again...)

But again: your library, your call.

codecholeric:
Sorry that I am so slow here :see_no_evil: I thought about it for quite a while and I'm still not convinced. I mean, yes, you can move the functions out into some utils, but that doesn't mean it's then code we don't have to maintain anymore :man_shrugging: I also don't want to say it's "unreadable", really, it's just *more* to read if you want to know what's going on (given that you set regex as a precondition, but that is just the case for ArchUnit anyway, because it's used in many places)

If we start going down this road we can likely find many places where we can multiply our lines of code to do some low-level constructs instead of higher level constructs and then optimize some "abstract" metric (like allocation). But I don't feel it moves the needle. If it should move the needle for the library we need something tangible. Like "for use case x this makes the import 10% faster" or "this benchmark shows how it reduces the necessary heap size for a typical import". For me those are the arguments that then count and that I've followed so far.
So, if we can't find a measurable impact I would not go for it.

But even if we should not go for it I really appreciate the discussion and making these things explicit. Because then we did measure, we know pro's and con's and if we should revisit we already have this knowledge :slightly_smiling_face:

dreis2211:
Thanks for your time.
> But I don't feel it moves the needle. If it should move the needle for the library we need something tangible. Like "for use case x this makes the import 10% faster" or "this benchmark shows how it reduces the necessary heap size for a typical import". For me those are the arguments that then count and that I've followed so far.

See my remark about this
> Our usage of ArchUnit is pretty limited. From a gut feeling of 1000 tests we have maybe 10 ArchUnit tests. Given that we already reach 0,5-1% with the very limited use, I contributed this

I don't know how I can make this more "tangible" for you. But if you feel this is not worthwhile and problematic for the maintenance of the project, please close the PR. This didn't seem like a problematic PR at all and I have a hard time understanding your scepticism on the matter, but in the end you're the maintainer of it. Not me.

Cheers

Issue #949, opened 2022-08-26
Ability to include/exclude Gradle testFixtures via importOptions


britter:
The `com.tngtech.archunit.core.importer.ImportOption` class currently only provides ways to include/exclude (among others) tests. With Gradle's `java-test-fixtures` plugin it's possible to have another source set called `testFixtures` and a corresponding classes directory containing test fixtures. In my case I want to ignore them from my architecture test, but it's currently not possible. Also there does not seem to be an easy way to extend filters with custom logic.

hankem:
The **good news** is that it's actually quite easy to use custom import options: With 
```java
class DoNotIncludeTestFixtures implements ImportOption {
    @Override
    public boolean includes(Location location) {
        return !isTestFixture(location);
    }
}
```
you can use
```java
@AnalyzeClasses(importOptions = DoNotIncludeTestFixtures.class // ...
```
or
```java
    static final JavaClasses classes = new ClassFileImporter()
            .withImportOption(new DoNotIncludeTestFixtures())
            .import//...
```

The **bad news** is that the local `testFixture` source set seems to get compiled to `$buildDir/classes/java/test/` just as the `test` source set, so you cannot distinguish `testFixture` and `test` classes unless you manage to change one output directory.

britter:
@hankem thank you for the pointer!

> The *bad news* is that the local `testFixture` source set seems to get compiled to `$buildDir/classes/java/test/` just as the test source set, so you cannot distinguish `testFixture` and `test` classes unless you manage to change one output directory.

That's not what I'm seeing. If not special configuration is applied, Gradle separates the output directories of source sets. So testFixtures will end up in `$buildDir/classes/java/testFixtures`.

However while debugging I realized that Gradle will put the test fixture jar on the test runtime classpath, not the source set output folder. So I needed to filter a little bit differently. Here's my full ImportOptions implementation that filters both test classes and test fixtures classes:

```
public class DotNotIncludeTestsOrTestFixtures implements ImportOption {

    private static final Pattern TESTS_PATTERN = Pattern.compile(".*/build/classes/([^/]+/)?test/.*");

    @Override
    public boolean includes(Location location) {
        return !isTests(location) && !isTestFixtures(location);
    }

    private static boolean isTestFixtures(Location location) {
        return location.isJar() && location.contains("test-fixtures.jar");
    }

    private static boolean isTests(Location location) {
        return location.matches(TESTS_PATTERN);
    }
}
```

[...]

PR #1010, opened 2022-11-28
Add `ImportOption` `DoNotIncludeGradleTestFixtures`

codecholeric:
Gradle provides the `java-test-fixtures` plugin out of the box, which will add a separate source set for creating test fixtures (i.e. test support code). This source set can see the main sources and be seen by the test sources. Depending on how the tests are run they might be put onto the classpath as folder path (containing the compiled sources) or JAR archive. Thus, we have to exclude both, the folder where Gradle compiles the files to by convention and archives that follow the Gradle convention to end in `-test-fixtures.jar`. However, this can of course only be a heuristic. Custom modules that mimic the convention will e.g. also be excluded. But in these cases users have to customize their import options then, instead of using the predefined one.

Resolves: #949

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
@@ -56,6 +56,17 @@ public interface ImportOption {
                 return onlyIncludeTests.includes(location);
             }
         },
+        /**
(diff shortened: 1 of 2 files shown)


britter:
@codecholeric I can confirm that this code works in my project by copying the `DoNotIncludeGradleTestFixtures` to my code base. Thanks for looking into this!

hankem (approved):
LGTM

PR #1040, opened 2023-01-06
More easily detect overridden methods & calls to super

### LINKED ISSUE #982: More easily detect overridden methods & calls to super
On a recent project we're heavily using the visitor pattern, and would like to ensure that people who override visitor methods call the super implementation. Seemed to me like a reasonable thing to check with ArchUnit, for fast feedback. Yet when implementing this I found there's some challenges with the current APIs, which might benefit from explicitly named constructs in the DSL.

Here's a reproduction sample of the domain:
```java
class Tree {
	UUID id;
	List<Tree> nodes;
}

class TreeVisitor {
	public Tree visit(Tree tree) {
		// Also visit child nodes
		for (Tree child : tree.nodes) {
			visit(child);
		}
		return tree;
	}
}

class GoodVisitor extends TreeVisitor {
	@Override
	public Tree visit(Tree tree) {
		return super.visit(tree);
	}
}

class IndifferentVisitor extends TreeVisitor {
    // Does not override, should be OK
}

class BadVisitor extends TreeVisitor {
	@Override
	public Tree visit(Tree tree) {
		return tree; // Lacks call to super.visit(tree), and thus traversal of child nodes
	}
}
```

I'm able for the simplest case to test this using the below constructs:
```java
import java.util.List;
import java.util.Set;
import java.util.UUID;
import java.util.function.Predicate;

import com.tngtech.archunit.base.DescribedPredicate;
import com.tngtech.archunit.core.domain.*;
import com.tngtech.archunit.core.domain.AccessTarget.CodeUnitCallTarget;
import com.tngtech.archunit.core.importer.ClassFileImporter;
import com.tngtech.archunit.junit.AnalyzeClasses;
import com.tngtech.archunit.lang.ArchCondition;
import com.tngtech.archunit.lang.ArchRule;
import com.tngtech.archunit.lang.ConditionEvents;
import com.tngtech.archunit.lang.SimpleConditionEvent;
import org.junit.jupiter.api.Test;

import static com.tngtech.archunit.core.domain.properties.HasName.Utils.namesOf;
import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.methods;
import static org.assertj.core.api.Assertions.assertThat;
import static org.junit.jupiter.api.Assertions.assertThrows;

@AnalyzeClasses
class ArchitectureTest {

//	@ArchTest
	public static final ArchRule visitor_methods_should_call_super = methods()
			.that().areDeclaredInClassesThat().areAssignableTo(TreeVisitor.class)
			.and().haveNameStartingWith("visit")
			.and(new OverridesSuperClassMethod())
			.should(new CallSuperClassMethod())
			.as("Overridden visit methods should call super class method")
			.because("visitor nagivation requires calls to super class methods");

	@Test
	void verify_visitor() {
		JavaClasses javaClasses = new ClassFileImporter().importPackagesOf(TreeVisitor.class);
		AssertionError error = assertThrows(AssertionError.class,
				() -> visitor_methods_should_call_super.check(javaClasses));
		assertThat(error.getMessage()).startsWith(
				"""
						Architecture Violation [Priority: MEDIUM] - Rule 'Overridden visit methods should call super class method, because visitor nagivation requires calls to super class methods' was violated (1 times):
						Method <com.github.timtebeek.BadVisitor.visit(com.github.timtebeek.Tree)> does not call super class method at (ArchitectureTest.java:""");
	}
}

class OverridesSuperClassMethod extends DescribedPredicate<JavaMethod> {
	OverridesSuperClassMethod() {
		super("overrides super class method");
	}

	@Override
	public boolean test(JavaMethod inputMethod) {
		return inputMethod.getOwner().getAllRawSuperclasses().stream()
				.flatMap(c -> c.getMethods().stream())
				.anyMatch(hasMatchingNameAndParameters(inputMethod));
	}

	static Predicate<JavaMethod> hasMatchingNameAndParameters(JavaMethod inputMethod) {
		return superMethod -> superMethod.getName().equals(inputMethod.getName())
				&& superMethod.getRawParameterTypes().size() == inputMethod.getRawParameterTypes().size()
				&& (superMethod.getDescriptor().equals(inputMethod.getDescriptor())
						|| namesOf(superMethod.getRawParameterTypes())
								.containsAll(namesOf(inputMethod.getRawParameterTypes())));
	}
}

class CallSuperClassMethod extends ArchCondition<JavaMethod> {
	CallSuperClassMethod() {
		super("call super class method");
	}

	@Override
	public void check(JavaMethod inputMethod, ConditionEvents events) {
		String ownerFullName = inputMethod.getOwner().getFullName();
		List<JavaClass> inputParameters = inputMethod.getRawParameterTypes();
		Set<JavaMethodCall> callsFromSelf = inputMethod.getMethodCallsFromSelf();
		boolean satisfied = callsFromSelf.stream()
				.map(JavaCall::getTarget)
				.filter(target -> !ownerFullName.equals(target.getOwner().getFullName()))
				.filter(target -> target.getName().equals(inputMethod.getName()))
				.map(CodeUnitCallTarget::getRawParameterTypes)
				.anyMatch(targetTypes -> targetTypes.size() == inputParameters.size());
		String message = String.format("%s does not call super class method at %s",
				inputMethod.getDescription(), inputMethod.getSourceCodeLocation());
		events.add(new SimpleConditionEvent(inputMethod, satisfied, message));
	}
}
```

Yet the `OverridesSuperClassMethod` & `CallSuperClassMethod` conditions are awkward and fragile to implement.

Worse still, the conditions as they are right now break down when we add subclasses into the mix:
```java
class JavaTree extends Tree {
}

class GoodJavaVisitor extends TreeVisitor {
	@Override
	public JavaTree visit(Tree tree) {
		return (JavaTree) super.visit(tree);
	}
}

class BadJavaVisitor extends TreeVisitor {
	@Override
	public JavaTree visit(Tree tree) {
		return (JavaTree) tree;
	}
}
```
```
Architecture Violation [Priority: MEDIUM] - Rule 'Overridden visit methods should call super class method, because visitor nagivation requires calls to super class methods' was violated (4 times):
Method <com.github.timtebeek.BadJavaVisitor.visit(com.github.timtebeek.Tree)> does not call super class method at (ArchitectureTest.java:1)
Method <com.github.timtebeek.BadJavaVisitor.visit(com.github.timtebeek.Tree)> does not call super class method at (ArchitectureTest.java:136)
Method <com.github.timtebeek.BadVisitor.visit(com.github.timtebeek.Tree)> does not call super class method at (ArchitectureTest.java:119)
Method <com.github.timtebeek.GoodJavaVisitor.visit(com.github.timtebeek.Tree)> does not call super class method at (ArchitectureTest.java:1)
```
Not immediately clear to me why this would add a violation for `GoodJavaVisitor` or two violations for `BadJavaVisitor`.

Would it be possible to implement detection of overridden methods & calls to super methods properly once in ArchUnit, and make that available through the API? That way more people can add similar checks more easily.

java-coding-prodigy:
Fixes #982

Currently added a `JavaMethod#isOverridden` method for now.


archunit/src/main/java/com/tngtech/archunit/core/domain/JavaMethod.java
```diff
@@ -125,6 +126,12 @@ public String getDescription() {
         return "Method <" + getFullName() + ">";
     }
 
+    @PublicAPI(usage = ACCESS)
+    public boolean isOverridden() {
+        Method method = methodSupplier.get();
+        return Arrays.asList(method.getDeclaringClass().getSuperclass().getDeclaredMethods()).contains(method);
```
hankem, line 132: It would be nice to avoid the (expensive) reflection API, but more importantly: Does this even work for you? Please provide a unit test that proves the correctnes of your code.

I'd use something like this:
```java
public class JavaMethodTest {
    @Test
    public void isOverridden() {
        class Base {
            void method1() {}
            void method1(int x) {}
        }
        class Child extends Base {
            void method1() {}
            void method2() {}
        }
        JavaClass childClass = new ClassFileImporter().importClass(Child.class);
        assertThat(childClass.getMethod("method1").isOverridden()).isTrue();
        assertThat(childClass.getMethod("method1", int.class).isOverridden()).isFalse();
        assertThat(childClass.getMethod("method2").isOverridden()).isFalse();
}
```
Your implementation however gives me `childClass.getMethod("method1").isOverridden() == false`.

hankem (changes requested):

java-coding-prodigy:
I couldn't figure out a direct way to get the inherited methods of the super class I tried to use `.getMethods()` but that only returns the public methods.
If we have something like a class extending a class which also extends another class, and the lowest class has a method that the highest class has, but the middle class does not, then `.isOverridden()` returns false.
I don't know whether our API or the reflection API has an easy way to fix this or not

hankem (changes requested):
You may have seen that the PR's `DCO` check fails. Can you please squash your changes and sign it off according to the [DCO](https://github.com/TNG/ArchUnit/blob/main/DCO)? 

It's also preferable to rebase to the current `main` instead of merging it in order to create a simple linear history.

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaMethodTest.java
```diff
@@ -0,0 +1,39 @@
...
+            void method2() {
+            }
+        }
+
+        JavaClass childClass = new ClassFileImporter().importClass(Child.class);
+        JavaClass baseClass = new ClassFileImporter().importClass(Base.class);
+        JavaMethod childMethod1 = childClass.getMethod("method1");
+        JavaMethod baseMethod1 = baseClass.getMethod("method1");
+        assertNotEquals(childMethod1, baseMethod1);
+        assertEquals(childMethod1.getName(), baseMethod1.getName());
```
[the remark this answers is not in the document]
java-coding-prodigy: Oops, I used that assertion while I was debugging why my check failed and forgot to remove it, will do so in the next commit.

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaMethod.java
```diff
@@ -15,20 +15,20 @@
...
 import com.tngtech.archunit.base.Suppliers;
 import com.tngtech.archunit.core.importer.DomainBuilders;
 
+import java.lang.reflect.Method;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
```
hankem, line 31: Can we please adhere to the import order specified in [`CONTRIBUTING.md`](https://github.com/TNG/ArchUnit/blob/main/CONTRIBUTING.md), i.e. keep the `java.*` imports at the top?

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaMethod.java
```diff
@@ -125,6 +125,14 @@ public String getDescription() {
         return "Method <" + getFullName() + ">";
     }
 
+    @PublicAPI(usage = ACCESS)
+    public boolean isOverridden() {
+        return getOwner().getAllRawSuperclasses().stream()
```
hankem, line 130: If we want to conisder overridden methods from interfaces as well, we can replace this stream with
```java
Streams.concat(getOwner().getAllRawSuperclasses().stream(), getOwner().getAllRawInterfaces().stream())
```

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaMethod.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2022 TNG Technology Consulting GmbH
+ * Copyright 2014-2023 TNG Technology Consulting GmbH
```
hankem, line 2: I guess we should isolate this _"Happy new year, code base!"_ change into a separate commit (that also fixes all java files, which happens automatically when building the project).

hankem (commented):
Nice; I think you're getting close! 👍

About the [DCO](https://github.com/TNG/ArchUnit/blob/main/DCO): I would do 
* `git rebase -i origin/main`
* replace `pick` with `f` for all commits after the first one and do
* `git commit --amend -s` to create the sign-off, followed by
* `git push -f`

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaMethod.java
java-coding-prodigy: Yeah for some reason my IntelliJ or some other plugin automatically made this change for all the classes with this documentation, I forgot to undo the changes on this file itself but did them for all other classes. 😅

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaMethod.java
java-coding-prodigy: Cool, I'll do so in `JavaMethodTest` as well

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaMethod.java
```diff
@@ -125,6 +126,14 @@ public String getDescription() {
         return "Method <" + getFullName() + ">";
     }
 
+    @PublicAPI(usage = ACCESS)
+    public boolean isOverridden() {
+        return Stream.concat(getOwner().getAllRawSuperclasses().stream(), getOwner().getAllRawInterfaces().stream())
```
hankem, line 131: Can we pleae add a small test to cover the feature that overriden methods from interfaces are detected?

hankem (commented):
Could you please sign-off your commits (most easily done after rebasing & squashing all into one), as suggested before in https://github.com/TNG/ArchUnit/pull/1040#pullrequestreview-1239752587?
Without the DCO check passing, GitHub won't let us merge the PR.

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaMethodTest.java
```diff
@@ -0,0 +1,58 @@
...
+        assertThat(baseClass.getMethod("method1", int.class).isOverridden()).isFalse();
+        assertThat(childClass.getMethod("method1").isOverridden()).isTrue();
+        assertThat(childClass.getMethod("method2").isOverridden()).isFalse();
+        assertThat(grandChildClass.getMethod("method1").isOverridden()).isTrue();
+        assertThat(grandChildClass.getMethod("method1", int.class).isOverridden()).isTrue();
+        assertThat(grandChildClass.getMethod("method2").isOverridden()).isTrue();
+        assertThat(grandChildClass.getMethod("method3").isOverridden()).isFalse();
+        //TODO add testing for methods with generic parameters
+    }
+}
```
hankem, line 58: We should also have a test for the use case of #1047 .

java-coding-prodigy: I believe that is already in place, currently we check if the parameters and name of the method are equal, and that implies compatibility of return type.
If that is not the case, then the user code will not compile in the first place

hankem: I thought that we should document that this works (which it curently does, but we also want to avoid any future regression):
```java
    @Test
    public void more_specific_return_type_is_supported() {
        class Parent {
            CharSequence method() { return "base"; }
        }
        class Child extends Parent {
            @Override
            String method() { return "overridden"; }
        }
        JavaClass childClass = new ClassFileImporter().importClass(Child.class);
        JavaMethod method = childClass.getMethod("method");
        assertThat(method.isOverridden()).isTrue();
    }
```

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaMethodTest.java
```diff
@@ -0,0 +1,58 @@
...
+        JavaClass grandChildClass = importer.importClass(GrandChild.class);
+        assertThat(baseClass.getMethod("method1").isOverridden()).isFalse();
+        assertThat(baseClass.getMethod("method1", int.class).isOverridden()).isFalse();
+        assertThat(childClass.getMethod("method1").isOverridden()).isTrue();
+        assertThat(childClass.getMethod("method2").isOverridden()).isFalse();
+        assertThat(grandChildClass.getMethod("method1").isOverridden()).isTrue();
+        assertThat(grandChildClass.getMethod("method1", int.class).isOverridden()).isTrue();
+        assertThat(grandChildClass.getMethod("method2").isOverridden()).isTrue();
+        assertThat(grandChildClass.getMethod("method3").isOverridden()).isFalse();
+        //TODO add testing for methods with generic parameters
```
hankem, line 56: It seems that we indeed don't recognize all overriden generic methods:
```java
    @Test
    public void overridden_generic_methods_are_supported() {
        class Parent<T extends Number> {
            void method(T t) { }
        }
        class Child extends Parent<Integer> {
            @Override
            void method(Integer t) { }
        }
        JavaClass childClass = new ClassFileImporter().importClass(Child.class);
        JavaMethod method = childClass.getMethod("method", Integer.class);
        assertThat(method.isOverridden()).isTrue();  // Expecting value to be true but was false 
    }
```

java-coding-prodigy: Hmmm, this might complicate the code now.
I'll see what I can do.

java-coding-prodigy:
So far I think I know what we can do to fix the issue.
We need to get the supperclasses and interfaces of the method's class but they cannot be the raw ones as they don't have type arguments.
So looks like we can keep on calling a .`supperclass()` again and again unti we reach `java.lang.Object` as well as use `.interfaces`
However this solves the current problem at hand, it may not be bug-free.

Also ignore `JavaMethodTest.java` that is just me playing around with the objects while debugging.

codecholeric:
@java-coding-prodigy are you still working on this? Do you need any support?

java-coding-prodigy:
> @java-coding-prodigy are you still working on this? Do you need any support?

I was working on it some time ago and I could continue now. However I had run into some issues earlier.
I'll see what I can do.

java-coding-prodigy:
Fixed the generic issue, however code is too ugly and slow, we should try to enhance it.
I have only done this for classes, but I presume it would be similar for interfaces too.

java-coding-prodigy:
I'm thinking of adding a `JavaMethod#overridenFrom(JavaType` and make `isOverriden` call that method. What do you think @hankem  ?

Issue #1049, opened 2023-01-23
Comment Regarding Generic Type Possibly No Longer Relevant


raphaelNguyen:
https://github.com/TNG/ArchUnit/blob/2c183aa14b416ab20bde2f9ad0023c23a9c37f5a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaField.java#L53-L61

When I called `getType()` on a field declared with generic types, `List<String> field`, I get back a `JavaType` object that is also an instance of `JavaParameterizedType` (actual class is `ImportedParameterizedType`) and describe the generic class sufficiently.

Is that what this comment is referring to, and if so, is this comment still relevant?

I was looking for a way to find the actual parameter type of a generic type field, found this comment which seems to indicate that it was not yet implemented but a bit of practical experimentation (screenshots below) seems to show that it does. Now I'm a bit confused.

Thank you for your time.

![image](https://user-images.githubusercontent.com/42530075/214173527-c4e97e88-df25-4e95-a55d-070a6050d29b.png)
![image](https://user-images.githubusercontent.com/42530075/214173385-e419057c-9f77-44e5-9297-3806896be396.png)

codecholeric:
Really sorry for the confusion, you're right, the comment is outdated! I've added a fix into this PR [d4fc2b5](https://github.com/TNG/ArchUnit/pull/1050/commits/d4fc2b5340082919dcdaacee90648262adabf3bb) (as it's just some minor docs change).

Issue #1038, opened 2023-01-05
Revise log levels


odrotbohm:
Importing class files currently writes a lot of `DEBUG` output to the console, as most loggers default to `DEBUG` log level. Especially in Spring Boot applications, that usually configure their integration test log levels via `application.properties` adding a configuration file dedicated to the logging library causes the Boot log configuration to back off, as that now expects *all* logging to be configured in that file. Ideally, I'd like to see ArchUnit to not produce any output, except logging explicitly configured to.

It would be nice if the following steps were taken:
* Lower the log levels in `PluginLoader` for the Java version detection to `TRACE`.
* Lower the log levels in `UrlSource`, `ArchConfiguration` and `JavaClassProcessor` to `TRACE`.

codecholeric:
Yes, I got you :joy: Should probably have done this a long time ago just to avoid the discussion :man_shrugging: I just always had the feeling that "taking `archunit.properties` from path x" is a different quality than "importing method foo". And setting everything to `TRACE` will effectively not allow to distinguish by level anymore. But I guess it's always possible to just to a category based configuration :man_shrugging: I'm gonna adjust it...

odrotbohm:
Thanks! 🫶

PR #1050, opened 2023-01-25
adjust all log-levels besides WARN to TRACE

codecholeric:
Over the years there have been continuous issues by people not configuring their logging system appropriately and then being spammed by masses of DEBUG logs. In particular, because ArchUnit hooks into SLF4J and as it seems e.g. Spring Boot sets the log-level to DEBUG if the configuration is missing. While I principally think people should just correctly configure their logging infrastructure I've also been worn down over time ;-) So, I will reduce all non-WARN levels to TRACE now and accept that it will not be possible anymore to distinguish extremely spammy output (like "found access from x to y") from more central output (like "dependency resolution configuration") by log-level.

Resolves: #1038
Resolves: #1049

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaType.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaType.java
@@ -15,12 +15,29 @@
  */
 package com.tngtech.archunit.core.domain;
 
+import java.lang.reflect.Type;
+
 import com.tngtech.archunit.PublicAPI;
 import com.tngtech.archunit.base.ChainableFunction;
 import com.tngtech.archunit.core.domain.properties.HasName;
 
 import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
 
+/**
+ * Represents a general Java type. This can e.g. be a class like {@code java.lang.String}, a parameterized type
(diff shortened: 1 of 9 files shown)


archunit/src/main/java/com/tngtech/archunit/core/domain/JavaType.java
```diff
@@ -15,12 +15,29 @@
...
 import com.tngtech.archunit.base.ChainableFunction;
 import com.tngtech.archunit.core.domain.properties.HasName;
 
 import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
 
+/**
+ * Represents a general Java type. This can e.g. be a class like {@code java.lang.String}, a parameterized type
+ * like {@code List<String>} or a type variable like {@code T}.<br>
+ * Besides having a {@link HasName#getName() name} and offering the possibility to being converted to an
+ * {@link #toErasure() erasure} (which is then always a raw class object) {@link JavaType} doesn't offer
```
hankem, line 30:
```suggestion
* {@link #toErasure() erasure} (which is then always {@link JavaClass a raw class object}) {@link JavaType} doesn't offer
```

hankem (approved):

PR #1052, opened 2023-01-26
Bump dependencies

hankem:
This PR drops the tests against non-LTS versions (which have aready reached their [end-of-life](https://endoflife.date/java)) and combines the dependency upgrades from PRs #1022, #1024, ~#1042~ #1053, #1044 (fixing [CVE-2023-22796](https://github.com/TNG/ArchUnit/security/dependabot/11)), #1048, & #1054.


codecholeric (approved):
Thanks a lot! Looks good and makes sense to reduce the CI duration by kicking out these tests against EOL versions. From what I saw online (and what IMHO makes sense) there are really down to no users at all on versions like JDK 9 or 10 (not like JDK 7 in older times that still had quite some users when it was already EOL for a while)

PR #1078, opened 2023-03-03
Add modules syntax to rules API

codecholeric:
This PR will add a convenient fluent rules API to partition `JavaClasses` into (architecture) modules and assert conditions on this module structure. The basic syntax has the form

```
modules()
  .definedBy... // use one of the various methods to decide how to partition classes into modules
  .should()... // evaluate some predefined or custom condition
```

While some concepts are quite similar to the `Slices` API I've noticed over the years that most people don't see the possibilities to use the `Slices` API to check modularization properties. Thus, I decided to not make the existing API more powerful, but instead create this new API which offers most features of the `Slices` API (e.g. create from package identifiers, check for cycles) but also more powerful / concise possibilities to introduce modularization into a code base (e.g. by an annotated `package-info`) and carry meta-information like allowed dependencies from such an annotated object into the modules rule.


hankem (commented):
Sorry, I didn't make it through all commits yet. I so far have just a few suggestions for minor renamings, but I thought I'd submit those already.

archunit/src/test/java/com/tngtech/archunit/library/modules/syntax/GivenModulesTest.java
```diff
@@ -0,0 +1,82 @@
...
+    @Rule
+    public final ArchConfigurationRule archConfigurationRule = new ArchConfigurationRule();
+
+    // note that description logic is already covered by RandomModulesSyntaxTest
+
+    @Test
+    public void allows_restricting_modules() {
+        archConfigurationRule.setFailOnEmptyShould(false);
+
+        assertThatRule(modulesByClassName().should(alwaysBeViolations())).checking(importClasses(Object.class, String.class))
```
hankem, line 31:
```suggestion
        assertThatRule(modulesByClassName().should(alwaysBeViolations()))
                .checking(importClasses(Object.class, String.class))
```

archunit-example/example-junit4/src/test/java/com/tngtech/archunit/exampletest/junit4/ModulesTest.java
```diff
@@ -0,0 +1,174 @@
...
+                    .findFirst()
+                    .map(annotatedClassInPackage -> ArchModule.Identifier.from(annotatedClassInPackage.getAnnotationOfType(AppModule.class).name()));
+
+            return identifierInCurrentPackage.orElseGet(identifierInParentPackageOf(javaPackage));
+        }
+
+        private Supplier<ArchModule.Identifier> identifierInParentPackageOf(JavaPackage javaPackage) {
+            return () -> javaPackage.getParent().isPresent()
+                    ? getIdentifierOfPackage(javaPackage.getParent().get())
+                    : ArchModule.Identifier.ignore();
```
hankem, line 171: Not relevant at all, but this could be:
```suggestion
            return () -> javaPackage.getParent()
                    .map(parent -> getIdentifierOfPackage(parent))
                    .orElseGet(ArchModule.Identifier::ignore);
```

archunit-integration-test/src/test/java/com/tngtech/archunit/PublicAPIRules.java
```diff
@@ -113,14 +117,31 @@ public class PublicAPIRules {
     public static final ArchRule only_entry_point_and_syntax_interfaces_should_be_public =
             classes()
                     .that().resideInAPackage("..syntax..")
-                    .and().haveNameNotMatching(".*" + ArchRuleDefinition.class.getSimpleName() + ".*")
+                    .and().haveNameNotMatching(".*RuleDefinition.*")
```
hankem, line 120: I wonder whether `ModuleRuleDefinition.modules()` couldn't live in `ArchRuleDefinition` as well. After all, module rules are `ArchRule`s, too. For disconverability, it would certainly help to have a consistent entry point – especially since the rule structure is so similiar.

hankem: If you prefer to keep `ModuleRuleDefinition`, this rule's description should be changed according to

archunit/src/main/java/com/tngtech/archunit/library/cycle_detection/Cycle.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.tngtech.archunit.PublicAPI;
+
+import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
+
+/**
+ * A cycle formed by the referenced {@code EDGEs}. A cycle in this context always refers to a "simple" cycle,
+ * i.e. the list of edges is not empty, the {@link Edge#getOrigin() origin} of the first {@link Edge} is equal
+ * to the {@link Edge#getTarget() target} of the last {@link Edge} and every node contained in the cycle
+ * is passed exactly once.
```
hankem, line 28: Maybe
```suggestion
 * is contained exactly once.
```
?

archunit/src/main/java/com/tngtech/archunit/library/cycle_detection/CycleDetector.java
```diff
@@ -0,0 +1,61 @@
...
+     * </p>
+     * Then both cycles, red and blue, would be reported.<br><br>
+     * For performance reasons the {@link CycleDetector} has a maximum number of cycles to detect.
+     * Once this limit is reached the algorithm will terminate and all cycles so far will be reported.
+     * For further information please refer to {@link Cycles#maxNumberOfCyclesReached()}.<br><br>
+     * Note that the passed edges must only reference the passed nodes as their {@link Edge#getOrigin() origin} and
```
hankem, line 45: Maybe
```suggestion
     * Note that the given edges must only reference the given nodes as their {@link Edge#getOrigin() origin} and
```
?

archunit/src/test/java/com/tngtech/archunit/library/cycle_detection/CyclesAssertion.java
```diff
@@ -0,0 +1,54 @@
...
+        return this;
+    }
+
+    CyclesAssertion containsOnly(Cycle<?>... cycles) {
+        hasSize(cycles.length);
+
+        Set<Set<List<?>>> thisOriginsAndTargets = actual.stream().map(it -> toOriginsAndTargets(it.getEdges())).collect(toSet());
+        Set<Set<List<?>>> otherOriginsAndTargets = Arrays.stream(cycles).map(it -> toOriginsAndTargets(it.getEdges())).collect(toSet());
+        assertThat(thisOriginsAndTargets).isSubsetOf(otherOriginsAndTargets);
+        assertThat(otherOriginsAndTargets).isSubsetOf(thisOriginsAndTargets);
```
hankem, line 36: Couldn't this be simplified?
```suggestion
        assertThat(thisOriginsAndTargets).isEqualTo(otherOriginsAndTargets);
```

archunit/src/main/java/com/tngtech/archunit/library/cycle_detection/rules/CycleArchCondition.java
```diff
@@ -0,0 +1,382 @@
...
+        void record(Cycle<ComponentDependency<COMPONENT>> cycle, ConditionEvents events) {
+            events.add(newEvent(cycle));
+        }
+
+        private ConditionEvent newEvent(Cycle<ComponentDependency<COMPONENT>> cycle) {
+            Map<String, ComponentDependency<COMPONENT>> descriptionsToEdges = sortEdgesByDescription(cycle);
+            String description = createDescription(descriptionsToEdges.keySet(), CYCLE_DETECTED_SECTION_INTRO.length());
+            String details = createDetails(descriptionsToEdges);
+            return new SimpleConditionEvent(cycle,
+                    false,
```
hankem, line 244: We could use `violated`.

archunit/src/main/java/com/tngtech/archunit/library/cycle_detection/rules/CycleArchCondition.java
```diff
@@ -0,0 +1,382 @@
...
+        // with respect to comparing the description lexicographically
+        @SuppressWarnings("OptionalGetWithoutIsPresent")
+        private ComponentDependency<COMPONENT> findStartEdge(Cycle<ComponentDependency<COMPONENT>> cycle) {
+            return cycle.getEdges().stream().min(comparing(input -> getDescriptionOfComponent.apply(input.getOrigin()))).get();
+        }
+
+        private String createDescription(Collection<String> edgeDescriptions, int indent) {
+            List<String> descriptions = new ArrayList<>(edgeDescriptions);
+            descriptions.add(descriptions.get(0));
+            return Joiner.on(" -> " + lineSeparator() + Strings.repeat(" ", indent)).join(descriptions);
```
hankem, line 271: `lineSeparator() + Strings.repeat(" ", CYCLE_DETECTED_SECTION_INTRO.length())` could be a static constant (`CYCLE_SEPARATOR`). Currently, it is re-calculated for every cycle.

I assume that you wanted to have both usages of `CYCLE_DETECTED_SECTION_INTRO` next to each other in `newEvent`, but IMO it's not obvious how `CYCLE_DETECTED_SECTION_INTRO.length()` is used in `createDescription` anyways.

archunit/src/main/java/com/tngtech/archunit/library/modules/syntax/ModulesShouldConjunction.java
```diff
@@ -0,0 +1,45 @@
...
+public interface ModulesShouldConjunction<DESCRIPTOR extends ArchModule.Descriptor> {
+
+    /**
+     * Like {@link #andShould(ArchCondition)} but offers a fluent API to pick the condition to join.
+     */
+    @PublicAPI(usage = ACCESS)
+    ModulesShould<DESCRIPTOR> andShould();
+
+    /**
+     * Joins another condition to this rule with {@code and} semantics. That is, all classes under test
```
hankem, line 36: Maybe
```suggestion
     * Joins another condition to this rule with {@code and} semantics. That is, all modules under test
```

archunit/src/main/java/com/tngtech/archunit/library/modules/syntax/ModulesByAnnotationShould.java
```diff
@@ -0,0 +1,109 @@
...
+     * and the annotated root classes
+     * <pre><code>
+     * &#64;MyModule(name = "Module One", allowedDependencies = {"Module Two", "Module Three"})
+     * interface ModuleOneDescriptor {}
+     *
+     * &#64;MyModule(name = "Module Two", allowedDependencies = {"Module Three"})
+     * interface ModuleTwoDescriptor {}
+     *
+     * &#64;MyModule(name = "Module Three")
+     * interface ModuleTwoDescriptor {}
```
hankem, line 56:
```suggestion
     * interface ModuleThreeDescriptor {}
```

archunit/src/test/java/com/tngtech/archunit/library/modules/syntax/ModulesShouldTest.java
```diff
@@ -0,0 +1,302 @@
...
+
+    @Test
+    public void respectTheirAllowedDependenciesDeclaredIn_rejects_property_of_wrong_type() {
+        assertThatThrownBy(
+                () -> modules().definedByAnnotation(TestModule.class)
+                        .should().respectTheirAllowedDependenciesDeclaredIn("name", consideringOnlyDependenciesBetweenModules())
+                        .evaluate(new ClassFileImporter().importPackagesOf(ClassInModule1.class, InternalClassInModule2.class))
+        )
+                .isInstanceOf(IllegalArgumentException.class)
+                .hasMessageContaining(String.format("Property @%s.name() must be of type %s[]", TestModule.class.getSimpleName(), String.class.getSimpleName()));
```
hankem, line 109: I would either use
```suggestion
                .hasMessageContaining(String.format("Property @%s.name() must be of type %s", TestModule.class.getSimpleName(), String[].class.getSimpleName()));
```
or directly inline it
```suggestion
                .hasMessageContaining(String.format("Property @%s.name() must be of type String[]", TestModule.class.getSimpleName()));
```

archunit/src/main/java/com/tngtech/archunit/library/modules/syntax/ModulesShould.java
```diff
@@ -0,0 +1,77 @@
...
+     * Creates a rule to check that the {@link ArchModule}s under consideration don't have any forbidden
+     * {@link ArchModule#getModuleDependenciesFromSelf() module dependencies} according to the passed
+     * {@code allowedDependencyPredicate}. It is possible to adjust which {@link Dependency class dependencies}
+     * the rule considers relevant by the passed {@link ModuleDependencyScope dependencyScope}.
+     *
+     * @param allowedDependencyPredicate Decides which {@link ModuleDependency module dependencies} are allowed.
+     *                                   If the {@link DescribedPredicate} returns {@code true} the dependency is allowed,
+     *                                   otherwise it is forbidden.
+     * @param dependencyScope            Allows to adjust which {@link Dependency dependencies} are considered relevant by the rule
+     * @return A {@link ArchRule} to be checked against a set of {@link JavaClasses}
```
hankem, line 43:
```suggestion
     * @return An {@link ArchRule} to be checked against a set of {@link JavaClasses}
```
(There are two other occurences in `ModulesShould` and two more in `ModulesByAnnotationShould`.)

archunit/src/main/java/com/tngtech/archunit/library/modules/syntax/ModulesByAnnotationShould.java
```diff
@@ -0,0 +1,109 @@
...
+import com.tngtech.archunit.library.modules.ArchModule;
+
+import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
+
+@PublicAPI(usage = ACCESS)
+public interface ModulesByAnnotationShould<ANNOTATION extends Annotation> extends ModulesShould<AnnotationDescriptor<ANNOTATION>> {
+
+    /**
+     * Like {@link #respectTheirAllowedDependencies(DescribedPredicate, ModuleDependencyScope)}, but the allowed dependencies will be automatically
+     * derived from the {@link ANNOTATION} property named {@code annotationPropertyName}. This property *must* be of type {@code String[]}
```
hankem, line 36: Why is `ANNOTATION` a `@link`? (I would have expected `@code`. Isn't it a parameter just like `annotationPropertyName`?)

archunit/src/test/java/com/tngtech/archunit/library/modules/syntax/ModulesShouldTest.java
hankem: Same in `onlyDependOnEachOtherThroughPackagesDeclaredIn_rejects_property_of_wrong_type`.

docs/userguide/008_The_Library_API.adoc
```diff
@@ -197,6 +196,69 @@ cycles.maxNumberToDetect=50
...
 
+
+=== Modularization Rules
+
+[NOTE]
+Note: ArchUnit doesn't strive to be a "competition" for module systems like the
+Java Platform Module System. Such systems have advantages like checks at compile time
+versus test time as ArchUnit does. So, if another module system works well in your
+environment, there is no need to switch over. But ArchUnit can bring JPMS-like features
+to older code bases, etc. Java 8 projects, or environments where the JPMS is for some
```
hankem, line 207: Should this be
```suggestion
to older code bases, e.g. Java 8 projects, or environments where the JPMS is for some
```
?

docs/userguide/008_The_Library_API.adoc
```diff
@@ -197,6 +196,69 @@ cycles.maxNumberToDetect=50
...
+----
+@AppModule(
+  name = "Module One",
+  allowedDependencies = {"Module Two", "Module Three"},
+  exposedPackages = {"..module_one.api.."}
+)
+package com.myapp.example.module_one;
+----
+
+We can then derive a general "self-growing" rule using the following API:
```
hankem, line 238: What means _"self-growing"_?

archunit/src/main/java/com/tngtech/archunit/base/ArchUnitException.java
```diff
@@ -48,6 +48,10 @@ public static class ReflectionException extends ArchUnitException {
         public ReflectionException(Throwable cause) {
             super(cause);
         }
+
+        public ReflectionException(String message, Throwable cause) {
```
hankem, line 52: Unused?

archunit-integration-test/src/test/java/com/tngtech/archunit/PublicAPIRules.java
hankem: Okay, I see that `ArchRuleDefinition` lives in `lang.syntax`, whereas `ModuleRuleDefinition` should probably live in some `library` package.

archunit/src/main/java/com/tngtech/archunit/library/modules/syntax/ModulesByAnnotationShould.java
```diff
@@ -0,0 +1,109 @@
...
+     * |Module Two| ------> |Module Three|
+     * `----------'         `------------'
+     * </code></pre>
+     *
+     * @param annotationPropertyName The name of the property declared within {@link ANNOTATION} that declares allowed dependencies to other {@link ArchModule}s by name
+     * @param dependencyScope        Allows to adjust which {@link Dependency dependencies} are considered relevant by the rule
+     * @return A {@link ArchRule} to be checked against a set of {@link JavaClasses}
+     */
+    @PublicAPI(usage = ACCESS)
+    ModulesByAnnotationRule<ANNOTATION> respectTheirAllowedDependenciesDeclaredIn(String annotationPropertyName, ModuleDependencyScope dependencyScope);
```
hankem, line 76: I wonder whether we should offer this API in a typesafe way:
```suggestion
    ModulesByAnnotationRule<ANNOTATION> respectTheirAllowedDependenciesDeclaredIn(Function<ANNOTATION, String[]> allowedDependencies, ModuleDependencyScope dependencyScope);
```
(`String[]` could also be `Iterable<String>` or something like that...)

If we don't insist on the rule description _"... should respect their allowed dependencies declared in 'allowedDependencies'"_, users of this API could easily replace the magic literal `"allowedDependencies"` with the method reference `AppModule::allowedDependencies`.

Same for `onlyDependOnEachOtherThroughPackagesDeclaredIn`, obviously.

archunit/src/main/java/com/tngtech/archunit/library/cycle_detection/rules/CycleArchCondition.java
```diff
@@ -0,0 +1,382 @@
...
+        }
+
+        private String createDetails(Map<String, ComponentDependency<COMPONENT>> descriptionsToEdges) {
+            List<String> details = new ArrayList<>();
+            int componentIndex = 0;
+            for (Map.Entry<String, ComponentDependency<COMPONENT>> edgeWithDescription : descriptionsToEdges.entrySet()) {
+                ++componentIndex;
+                details.add(String.format("  %d. Dependencies of %s", componentIndex, edgeWithDescription.getKey()));
+                details.addAll(dependenciesDescription(edgeWithDescription.getValue()));
+            }
```
hankem, line 281: I personally prefer `descriptionsToEdges.forEach` over because you can give better names than `edgeWithDescription.getKey()` and `edgeWithDescription.getValue()` – even though a primitive `int` doesn't work as index anymore:
```suggestion
            AtomicInteger componentIndex = new AtomicInteger(0);
            descriptionsToEdges.forEach((description, dependencies) -> {
                details.add(String.format("  %d. Dependencies of %s", componentIndex.incrementAndGet(), description));
                details.addAll(dependenciesDescription(dependencies));
            });            
```

archunit/src/main/java/com/tngtech/archunit/library/cycle_detection/rules/CycleArchCondition.java
```diff
@@ -0,0 +1,382 @@
...
+                    ? allDependencies.stream().limit(maxDependencies).collect(toList())
+                    : allDependencies;
+
+            List<String> result = dependenciesToDisplay.stream()
+                    .map(dependency -> DEPENDENCY_DETAILS_INDENT + "- " + dependency.getDescription())
+                    .collect(toCollection(ArrayList::new));
+            if (tooManyDependenciesToDisplay) {
+                result.add(DEPENDENCY_DETAILS_INDENT + String.format("(%d further dependencies have been omitted...)",
+                        allDependencies.size() - dependenciesToDisplay.size()));
+            }
```
hankem, line 300: Do you think that `Stream.limit` is expensive?
I'd drop the extra complexity to optimize the calculation of `dependenciesToDisplay` and just do:
```suggestion
            List<String> result = allDependencies.stream()
                    .limit(maxDependencies)
                    .map(dependency -> DEPENDENCY_DETAILS_INDENT + "- " + dependency.getDescription())
                    .collect(toCollection(ArrayList::new));
            if (tooManyDependenciesToDisplay) {
                result.add(DEPENDENCY_DETAILS_INDENT + String.format("(%d further dependencies have been omitted...)",
                        allDependencies.size() - maxDependencies));
            }
```

archunit/src/main/java/com/tngtech/archunit/library/modules/syntax/ModulesByAnnotationShould.java
```diff
@@ -0,0 +1,109 @@
...
+
+@PublicAPI(usage = ACCESS)
+public interface ModulesByAnnotationShould<ANNOTATION extends Annotation> extends ModulesShould<AnnotationDescriptor<ANNOTATION>> {
+
+    /**
+     * Like {@link #respectTheirAllowedDependencies(DescribedPredicate, ModuleDependencyScope)}, but the allowed dependencies will be automatically
+     * derived from the {@link ANNOTATION} property named {@code annotationPropertyName}. This property *must* be of type {@code String[]}
+     * and contain the {@link ArchModule#getName() names} of the {@link ArchModule}s to which access is allowed.
+     * <br><br>
+     * For example, given the annotation
```
hankem, line 39: > given the annotation `@MyModule`

I feel that we should emphasize more that users have to define their own annotation, and that this modules-by-annotation API is not limited to an ArchUnit-annotation (which most likely wouldn't be available for `src/main` – even though this is not a requirement; the modules could be defined in `src/test`, too).

archunit/src/main/java/com/tngtech/archunit/library/modules/ArchModules.java
```diff
@@ -0,0 +1,556 @@
...
+
+/**
+ * A collection of {@link ArchModule "architectural modules"}. This class provides a convenient API to partition the {@link JavaClass classes}
+ * of a code base into (cohesive) modules and assert properties of these modules. E.g. their dependencies to each other or dependencies
+ * not contained in any of the constructed modules.<br>
+ * This class provides several entry points to create {@link ArchModule modules} from a set of {@link JavaClass classes} in a convenient way:<br>
+ * <ul>
+ *     <li>{@link #defineBy(IdentifierAssociation)} - the most generic API allowing a lot of flexibility</li>
+ *     <li>{@link #defineByPackages(String)} - an API similar to {@link Slices#matching(String)}</li>
+ * </ul>
```
hankem, line 63: Should `{@link #defineByRootClasses(Predicate)}`, `{@link #defineByAnnotation(Class)}`, `{@link #defineByAnnotation(Class, Function)}` also be mentioned here?

archunit/src/main/java/com/tngtech/archunit/library/modules/ArchModules.java
```diff
@@ -0,0 +1,556 @@
...
+
+    private ArchModules(Set<ArchModule<DESCRIPTOR>> modules) {
+        this.modulesByIdentifier = modules.stream().collect(toMap(ArchModule::getIdentifier, identity()));
+        this.modulesByName = modules.stream().collect(toMap(ArchModule::getName, identity()));
+
+        SetMultimap<ArchModule.Identifier, ModuleDependency<DESCRIPTOR>> moduleDependenciesByOrigin = HashMultimap.create();
+        modules.forEach(it -> moduleDependenciesByOrigin.putAll(it.getIdentifier(), createModuleDependencies(it, modules)));
+
+        SetMultimap<ArchModule.Identifier, ModuleDependency<DESCRIPTOR>> moduleDependenciesByTarget = HashMultimap.create();
+        moduleDependenciesByOrigin.values().forEach(it -> moduleDependenciesByTarget.put(it.getTarget().getIdentifier(), it));
```
hankem, line 78: I would usually prefer a semantic variable name instead of `it`.
For `modules.forEach(it`, it's maybe easy to see that `it` is a `module`, but I find it more difficult for `moduleDependenciesByOrigin.values()`, so I'd personally go for `moduleDependency` here.

archunit/src/main/java/com/tngtech/archunit/library/modules/ArchModule.java
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    /**
+     * An {@link Identifier} of an {@link ArchModule}. An {@link Identifier} is basically an ordered list of string parts that
+     * uniquely identifies an {@link ArchModule}, i.e. two {@link ArchModule modules} are equal, if and only if
+     * their identifier is equal (i.e. all the textual parts of the identifier match in order).
+     */
+    @PublicAPI(usage = ACCESS)
+    public static final class Identifier implements Iterable<String> {
+        private final List<String> parts;
```
hankem, line 195: I've probably missed something, but my impression was that the Identifier API with multiple parts might be more powerful than actually needed. Wouldn't a `String` usually be good enough?

hankem (approved):
I'm really sorry that it took me so long to look through this PR!
With `149 files changed, 6422 insertions(+), 987 deletions(-)`, it has introduced quite a few new APIs, but I don't think that I'll find any issues from further looking at it.

I've left lots of minor comments, which are mostly irrelevant (I just wrote down whenever I noticed something). The only major suggestion is [the one about declaring allowed dependencies and exposed packages](https://github.com/TNG/ArchUnit/pull/1078/#discussion_r1163370234). Furthermore, [I'm not sure that the flexibility of multi-part `ArchModule.Identifier`s is needed](https://github.com/TNG/ArchUnit/pull/1078/files#r1166095517), but it doesn't hurt (much 😉).

codecholeric:
Thanks a lot and yeah, I know the PR is quite massive :grimacing: IMHO we mainly need to make sure we don't release any "stupid" public API that we're then sitting on for a long time :wink: Everything else we can evolve later :man_shrugging: 
I'm gonna check out your comments!

archunit/src/main/java/com/tngtech/archunit/library/modules/ArchModule.java
codecholeric: This comes a little from the package parsing. Suppose I'm parsing hierarchical packages like
```
my.app.order.wares
my.app.order.billing
my.app.customer.management
my.app.customer.contact
```
with `my.app.(*).(*)..`. Then the structured identifier would make it quite easy to derive a name like `Module Order - Submodule Wares` from this. It would of course also be possible to parse a single string identifier again, but this API seemed more natural from this viewpoint :man_shrugging: 

Does this make sense to you or do you think I've just given too much emphasis to a corner case, complicating the typical use case? :thinking:

archunit-integration-test/src/test/java/com/tngtech/archunit/PublicAPIRules.java
codecholeric: Yes, it is also consistent with `SliceRuleDefinition`. If we would add it to `ArchRuleDefinition` it would maybe make discoverability easier, but it would also violate the intended layer structure (where `lang` shouldn't depend on `library`).
If we would want something central like that then `ArchRuleDefinition` should not reside within `lang` but on a higher hierarchy layer. But as I said, we have also `Architectures` and `SliceRuleDefinition` as entry points to define `ArchRule`s within `library`, so at the moment it's at least not inconsistent...

codecholeric: I thought the rule text is still fair as it is, because everything of the modules API also resides in a package `..syntax..` and that's the only thing the test checks :man_shrugging:

archunit/src/main/java/com/tngtech/archunit/library/modules/ModuleDependency.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    private static Set<Dependency> filterDependenciesBetween(ArchModule<?> origin, ArchModule<?> target) {
+        ImmutableSet.Builder<Dependency> result = ImmutableSet.builder();
+        for (Dependency dependency : origin.getClassDependenciesFromSelf()) {
+            if (target.contains(dependency.getTargetClass())) {
+                result.add(dependency);
+            }
+        }
+        return result.build();
```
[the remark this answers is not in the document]
codecholeric: Or even

```suggestion
        return origin.getClassDependenciesFromSelf().stream()
	                        .filter(dependency -> target.contains(dependency.getTargetClass()))
	                        .collect(toImmutableSet());
```

:slightly_smiling_face:

archunit/src/main/java/com/tngtech/archunit/library/modules/syntax/ModulesByAnnotationShould.java
codecholeric: The reason for me was that `annotationPropertyName` is a locally scoped thing (and I can't even link to it if I want), but the scope of `ANNOTATION` is wider :man_shrugging: Making it a link makes it immediately clear for me, that I mean the class type parameter :man_shrugging:

archunit/src/main/java/com/tngtech/archunit/library/modules/syntax/ModulesByAnnotationShould.java
codecholeric: Maybe

```suggestion
     * For example, given the user-defined annotation
```

archunit/src/main/java/com/tngtech/archunit/library/modules/syntax/ModulesByAnnotationShould.java
codecholeric: :see_no_evil:

archunit/src/main/java/com/tngtech/archunit/library/modules/syntax/ModulesByAnnotationShould.java
codecholeric: I had actually thought the same thing and even already implemented that change, but then as you say, it came down to the rule text. I wanted to see this in the rule text somehow, to be consistent with other places and to also immediately see from the rule text how this is defined. Then I switched it to a `DescribedFunction`. But by the end with all the bloat I felt the convenience was actually gone :man_shrugging: So it came down to rule text vs type safety and I finally thought, that these pure architecture annotations likely don't change often and if, then I would expect some rule could fail and check it anyway :man_shrugging: That was my thought process at least :thinking: Not having it in the rule text at all also didn't feel good to me :thinking:

archunit/src/test/java/com/tngtech/archunit/library/cycle_detection/CyclesAssertion.java
codecholeric: Hmm, I think this might be a refactoring leftover, I also think it should work this way :+1: :slightly_smiling_face: At least if I kept anything from my math studies :joy:

archunit/src/main/java/com/tngtech/archunit/library/modules/ArchModules.java
```diff
@@ -0,0 +1,556 @@
...
+ * </ul>
+ */
+@PublicAPI(usage = ACCESS)
+public final class ArchModules<DESCRIPTOR extends ArchModule.Descriptor> extends ForwardingCollection<ArchModule<DESCRIPTOR>> {
+    private final Map<Identifier, ArchModule<DESCRIPTOR>> modulesByIdentifier;
+    private final Map<String, ArchModule<DESCRIPTOR>> modulesByName;
+
+    private ArchModules(Set<ArchModule<DESCRIPTOR>> modules) {
+        this.modulesByIdentifier = modules.stream().collect(toMap(ArchModule::getIdentifier, identity()));
+        this.modulesByName = modules.stream().collect(toMap(ArchModule::getName, identity()));
```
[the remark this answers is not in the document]
codecholeric: Yeah, I had thought maybe that together with the value would be good enough, but probably you're right and we should present them in a better way to the user :thinking:

archunit-integration-test/src/test/java/com/tngtech/archunit/PublicAPIRules.java
hankem: My suggestion is to change
```diff
     public static final ArchRule only_entry_point_and_syntax_interfaces_should_be_public =
             classes()
                     .that().resideInAPackage("..syntax..")
                     .and().haveNameNotMatching(".*RuleDefinition.*")
                     .and().areNotInterfaces()
                     .and().areNotAnnotatedWith(Internal.class)
                     .and(are(not(onlyUsedAsPublicApiParameter())))
                     .should().notBePublic()
-                    .as(String.format(
-                            "Only %s and interfaces within the ArchUnit syntax (..syntax..) should be public",
-                            ArchRuleDefinition.class.getSimpleName()));
+                    .as("Only RuleDefinitions and interfaces within the ArchUnit syntax (..syntax..) should be public");
```
to match the change of commit `add modules API to syntax`:
```diff
     public static final ArchRule only_entry_point_and_syntax_interfaces_should_be_public =
             classes()
                     .that().resideInAPackage("..syntax..")
-                    .and().haveNameNotMatching(".*" + ArchRuleDefinition.class.getSimpleName() + ".*")
+                    .and().haveNameNotMatching(".*RuleDefinition.*")
                     .and().areNotInterfaces()
                     .and().areNotAnnotatedWith(Internal.class)
                     .should().notBePublic()
                     .as(String.format(
                             "Only %s and interfaces within the ArchUnit syntax (..syntax..) should be public",
                             ArchRuleDefinition.class.getSimpleName()));
```

archunit/src/main/java/com/tngtech/archunit/library/modules/syntax/ModulesByAnnotationShould.java
hankem: Hehehe, I see your point. 
I had assumed that a custom `AppModule` annotation wouldn't have multiple ambiguous properties, such that the rule description _"modules defined by annotation `@AppModule` should respect their allowed dependencies"_ might be good enough for a rule
```java
        modules()
                .definedByAnnotation(AppModule.class)
                .should().respectTheirAllowedDependenciesDeclaredIn(AppModule::allowedDependencies, // ...
```
so I would have prefered the version that has less degrees of freedom (allowing for usage errors) and simpler code, but it's just a thought. You're definitely right that it doesn't matter once the rule is written, as it will almost never change.

archunit/src/main/java/com/tngtech/archunit/library/modules/ArchModule.java
hankem: The only _"complications"_ are that users have to choose from a richer API, and that maintainers have to, well, maintain it. 😜

archunit/src/test/java/com/tngtech/archunit/library/modules/ArchModulesTest.java
```diff
@@ -0,0 +1,511 @@
...
+
+    private String getExamplePackage(String subpackageName) {
+        return getClass().getPackage().getName() + ".testexamples." + subpackageName;
+    }
+
+    private static ModuleDependenciesAssertion assertThatModuleDependencies(Collection<? extends ModuleDependency<?>> dependencies) {
+        return new ModuleDependenciesAssertion(dependencies);
+    }
+
+    static class ModuleDependenciesAssertion extends
```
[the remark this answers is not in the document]
codecholeric: I guess you mean for discoverability in the future? But I'm not completely convinced. Because often I feel we ponder a lot about the production code to be modular and encapsulated, then in the test scope we create some monolithic entangled mess where one place knows all the contexts again :thinking: That ship might already have sailed with `testutil.assertion`, but I still have my doubts that this specific library stuff should end up there :thinking: Likely a convention where modularized assertions somehow mimic the production structure might be better :thinking: I think as long as `testutil.assertion` only tackles core domain objects it would be okay. That we've mingled assertions for core and lang is already a little questionable, but also mingling in super specific library assertions ... I don't know :man_shrugging:

archunit/src/main/java/com/tngtech/archunit/library/modules/ArchModules.java
```diff
@@ -0,0 +1,556 @@
...
+                    identifierToRootClass.put(identifier, rootClass);
+                });
+            }
+
+            private boolean packagesOverlap(String firstPackageName, String secondPackageName) {
+                return packageContains(firstPackageName, secondPackageName) || packageContains(secondPackageName, firstPackageName);
+            }
+
+            private boolean packageContains(String parentPackage, String childPackage) {
+                return childPackage.equals(parentPackage) || childPackage.startsWith(parentPackage + ".");
```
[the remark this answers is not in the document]
codecholeric: I know what you mean, but since for one side we really just have the package name I wouldn't know how it should solve this case completely? I mean, we could check `rootClass.getPackage().containsPackageWithName(pkg)`, but how would the other way round look like if `pkg` is just a string? And since `pkg` is derived from the module declaration (and thus is no `JavaPackage` I don't know how a method on `JavaPackage` could solve this then?

archunit/src/test/java/com/tngtech/archunit/library/modules/ArchModulesTest.java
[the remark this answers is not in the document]
hankem: Okay, convinced! 😉

archunit/src/main/java/com/tngtech/archunit/library/modules/ArchModules.java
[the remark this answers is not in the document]
codecholeric: Actually, for identifier it doesn't matter. Because the code around ensures there is no way (we're always creating this by grouping classes by identifier, thus 2 modules with the same identifier can't be created).
For name I'm gonna add some validation though...

archunit/src/main/java/com/tngtech/archunit/library/modules/ArchModule.java
codecholeric: I hereby accept the risk :wink: (you can hold me accountable in a year or so when we're bitten by this :stuck_out_tongue_winking_eye:)

archunit/src/main/java/com/tngtech/archunit/library/modules/ArchModules.java
codecholeric: Yes, I wanted to extend the docs in every further commit :see_no_evil:

archunit/src/main/java/com/tngtech/archunit/library/modules/ArchModules.java
codecholeric: For me it depends, if it's very simple I like `it`, but I agree that in this case where we refer to it multiple times this threshold is likely crossed :wink:

archunit/src/main/java/com/tngtech/archunit/library/modules/syntax/ModulesByAnnotationShould.java
codecholeric: I would keep it like this for now for consistency. Because I think a method `respectTheirAllowedDependenciesDeclaredIn` that only adds `respect their allowed dependencies` to the rule text also has some further complications (e.g. automated testing assumes that the method names really match the textual description).
I we see people complaining about this we can still ponder about adding a safer API...

archunit/src/main/java/com/tngtech/archunit/library/modules/ArchModules.java
```diff
@@ -0,0 +1,556 @@
...
+         */
+        @PublicAPI(usage = ACCESS)
+        public Creator deriveNameFromPattern(final String namingPattern) {
+            return new Creator(identifierAssociation, identifier -> {
+                String result = namingPattern;
+                for (int i = 1; i <= identifier.getNumberOfParts(); i++) {
+                    result = result
+                            .replace("$" + i, identifier.getPart(i))
+                            .replace("${" + i + "}", identifier.getPart(i))
+                            .replace("$@", joinIdentifier(identifier));
```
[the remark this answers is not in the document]
codecholeric: Yeah, true :man_facepalming:

archunit/src/main/java/com/tngtech/archunit/library/modules/ArchModules.java
```diff
@@ -0,0 +1,556 @@
...
+        @PublicAPI(usage = ACCESS)
+        public <D extends ArchModule.Descriptor> WithGenericDescriptor<D> describeBy(DescriptorCreator<D> descriptorCreator) {
+            return new WithGenericDescriptor<>(identifierAssociation, descriptorCreator);
+        }
+
+        /**
+         * @see WithGenericDescriptor#modularize(JavaClasses)
+         */
+        @PublicAPI(usage = ACCESS)
+        public ArchModules<?> modularize(JavaClasses classes) {
```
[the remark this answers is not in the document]
codecholeric: As discussed, "modularize" for the action of creating modules from a set of elements seems "good enough" for now and the imperative form `partition(classes)` would otherwise also look somewhat strange in the base API :wink:

archunit/src/main/java/com/tngtech/archunit/library/modules/ArchModules.java
```diff
@@ -0,0 +1,556 @@
...
+     */
+    @PublicAPI(usage = ACCESS)
+    public Optional<ArchModule<DESCRIPTOR>> tryGetByIdentifier(String... identifier) {
+        return Optional.ofNullable(modulesByIdentifier.get(Identifier.from(identifier)));
+    }
+
+    /**
+     * @param name The name of an {@link ArchModule}
+     * @return A contained {@link ArchModule} with the passed {@link ArchModule#getName() name}.
+     *         This method will throw an exception if no matching {@link ArchModule} is contained.
```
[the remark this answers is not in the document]
hankem: Hm, I thought that I had spotted a copy & error, but I don't see it anymore. Maybe I was confused (more than usually, I mean).

archunit/src/main/java/com/tngtech/archunit/library/modules/ArchModules.java
```diff
@@ -0,0 +1,556 @@
...
+         *
+         * @param namingPattern A string naming pattern deriving the {@link ArchModule} {@link ArchModule#getName() name} from
+         *                      the {@link ArchModule.Identifier}
+         * @return A fluent API to further customize how to create {@link ArchModules}
+         */
+        @PublicAPI(usage = ACCESS)
+        public Creator deriveNameFromPattern(final String namingPattern) {
+            return new Creator(identifierAssociation, identifier -> {
+                String result = namingPattern;
+                for (int i = 1; i <= identifier.getNumberOfParts(); i++) {
```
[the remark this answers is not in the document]
codecholeric: Yes, true in theory, but in this case I would wait for somebody to really complain about this before calling to action :joy: Because in this very specific context I can hardly imagine it... And if somebody would really encounter this they would have the more generic API to fall back to to solve their issue.

archunit/src/main/java/com/tngtech/archunit/library/modules/ArchModules.java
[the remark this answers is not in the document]
codecholeric: As discussed, looks like a :ghost: so I'm gonna resolve it.

archunit/src/main/java/com/tngtech/archunit/library/modules/ArchModules.java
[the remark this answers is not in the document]
codecholeric: As discussed, we're gonna keep this for now. If we should improve `JavaPackage` later on to cover this case we can still revisit this.

archunit/src/main/java/com/tngtech/archunit/library/cycle_detection/rules/CycleArchCondition.java
codecholeric: Hmm, yeah, the idea was to see the connection at the call site :thinking: But you're right, defining constants next to each other is likely also good enough...

archunit/src/main/java/com/tngtech/archunit/library/modules/ArchModules.java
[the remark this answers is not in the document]
hankem: I finally had a chance to take a closer look in my IDE (I find it super difficult to see it on GitHub).

I had initially seen eb23cb9206ace196a2e954038329010e913b28e1:
```java
    /**
     * @param identifier The (textual) parts of an {@link ArchModule.Identifier}.
     * @return The contained {@link ArchModule} having an {@link ArchModule.Identifier} comprised of the passed {@code identifier} parts.
     *         This method will throw an exception if no matching {@link ArchModule} is contained.
     */
    @PublicAPI(usage = ACCESS)
    public Optional<ArchModule> tryGetByIdentifier(String... identifier) {
        return Optional.ofNullable(modulesByIdentifier.get(Identifier.from(identifier)));
    }
```
It has been fixed with d441ee1580a11594101d0e83a9ad77893c330814.

docs/userguide/008_The_Library_API.adoc
codecholeric: Mhmm, what I meant was that the rule automatically works for new components without the need to be adjusted. Take the PlantUML rule as a contrary example, every time you add a new module you have to adjust the diagram, i.e. the rule :man_shrugging: 
What I mean is that this can be set up once and teams can follow the convention without having to adjust the arch test itself...

codecholeric: I just omitted the part, because I think it's evident enough from the following explanation and doesn't really contribute much to the docs :+1:

archunit/src/main/java/com/tngtech/archunit/library/cycle_detection/rules/CycleArchCondition.java
codecholeric: Yeah, sounds reasonable, I didn't look over the code again because I just moved code from one place to the other in this refactoring. But sure, let's take the opportunity to simplify this a little :slightly_smiling_face:

archunit/src/main/java/com/tngtech/archunit/library/modules/ModuleDependency.java
```diff
@@ -0,0 +1,126 @@
...
+        Set<Dependency> classDependencies = filterDependenciesBetween(origin, target);
+        return !classDependencies.isEmpty()
+                ? Optional.of(new ModuleDependency<>(origin, target, classDependencies))
+                : Optional.empty();
+    }
+
+    private static Set<Dependency> filterDependenciesBetween(ArchModule<?> origin, ArchModule<?> target) {
+        return origin.getClassDependenciesFromSelf().stream()
+	                        .filter(dependency -> target.contains(dependency.getTargetClass()))
+	                        .collect(toImmutableSet());
```
hankem, line 124: Just FYI: these lines somehow got an inconsistent indentation: 😉
```suggestion
                .filter(dependency -> target.contains(dependency.getTargetClass()))
                .collect(toImmutableSet());
```

hankem (approved):

Issue #1069, opened 2023-02-23
TryCatchBlock in lambda will not be collected


demonbug:
the second assertion will fail
```
    @Test
    public void test_trycatch_contract() {
        JavaClasses classes = new ClassFileImporter().importClasses(ClassWithTryCatchInLambda.class);

        List<JavaMethodCall> calls = classes.stream()
            .flatMap(cls -> cls.getMethodCallsFromSelf().stream())
            .filter(call -> "parseDouble".equals(call.getName()))
            .collect(Collectors.toList());
        Assertions.assertFalse(calls.isEmpty());

        List<TryCatchBlock> tryBlocks =
            calls.stream().flatMap(call -> call.getContainingTryBlocks().stream()).collect(Collectors.toList());
        Assertions.assertFalse(tryBlocks.isEmpty());
    }

    private class ClassWithTryCatchInLambda {
        public void testCall() {
            testFunctionInterface(text -> {
                try {
                    Double.parseDouble(text);
                } catch (NumberFormatException e) {
                    e.printStackTrace();
                }
            });
        }

        private void testFunctionInterface(Consumer<String> func) {
            func.accept("123456");
        }
    }
```

codecholeric:
Thanks for reporting this! Yes, this seems to be a bug, because the synthetic methods created for lambdas are treated differently. We have to remap the try-blocks from those to the original methods. I'm gonna try to fix this (unless somebody else wants to jump in :wink:)

PR #1096, opened 2023-04-02
Fix missing try-catch-blocks in lambdas

codecholeric:
With #847 we have changed the way lambdas are handled. Before ArchUnit didn't have any special handling for lambdas which lead to synthetic methods (like `lambda$foo$123`) being imported like regular methods and connections like `call() -> doSomething()` in the following example would be lost:

```
void call() {
  inLambda(() -> doSomething());
}
```

Afterwards we started to eliminate these synthetic methods and attached the call directly to the origin method. However, try-catch-blocks are also associated with methods as their owners and here we forgot to attach them to the correct declaring method. Thus, by throwing out synthetic lambda methods from the import we would completely lose try-catch-blocks declared within lambdas.

We now fix this by resolving the correct origin/declaring method for any try-catch-block, even if it's declared within a lambda.

Resolves: #1069


codecholeric:
Mhmm, looking at the failed CI it probably makes sense to check out the memory footprint :grimacing: I checked CPU performance and that seems okay, but maybe the additional caching of the access resolution really drives up the needed heap :thinking: Just need to figure out if this is because we were close to the limit and "those additional 50MB now broke it", or if it's like "now you need double the heap" :grimacing:

archunit/src/test/java/com/tngtech/archunit/testutil/Assertions.java
```diff
@@ -228,8 +227,4 @@ public static CodeUnitAccessAssertion assertThatCall(JavaConstructorCall call) {
     public static TryCatchBlockAssertion assertThatTryCatchBlock(TryCatchBlock tryCatchBlock) {
         return new TryCatchBlockAssertion(tryCatchBlock);
     }
-
-    public static TryCatchBlocksAssertion assertThatTryCatchBlocks(Set<TryCatchBlock> tryCatchBlocks) {
```
hankem, line 232: This change is currently in the commit `make processing of raw try-catch-blocks consistent with other imports`, but I think that it would better fit in `remove obsolete class `TryCatchBlocksAssertion`.

hankem (approved):
Looks good, thank you!

Issue #857, opened 2022-04-28
[How to] Detect reference for field in methods


omuomugin:
I've trying to write tests like below

```
ArchRuleDefinition.fields()
  .that().inSomeCondition()
  .should()
  .notBeAccessedFromMethods().inSomeCondition()
```

or 

```
ArchRuleDefinition.methods()
  .that().inSomeCondition()
  .should()
  .notBeAccessFields().inSomeCondition()
```

Is there any kind of support for detecting dependency of fields in methods.

codecholeric:
I think this is unfortunately still missing from the fluent API, even though it would be a nice addition! For the moment you will have to resort to add this yourself, e.g. like

```
@ArchTest
public static final ArchRule example =
  fields().that().inSomeCondition()
    .should(notBeAccessedByMethodsThat(inSomeConditionPredicate));

private static ArchCondition<JavaField> notBeAccessedByMethodsThat(DescribedPredicate<JavaMethod> predicate) {
  return new ArchCondition<JavaField>("not be accessed by methods that " + predicate.getDescription()) {
  @Override
  public void check(JavaField field, ConditionEvents events) {
    field.getAccessesToSelf().stream()
      .filter(access ->
        access.getOrigin() instanceof JavaMethod
          && predicate.test((JavaMethod) access.getOrigin())
      )
      .forEach(violation ->
        events.add(SimpleConditionEvent.violated(field, violation.getDescription()))
      );
  }
  };
}
```

I'll keep the issue open, maybe somebody wants to contribute this to the fluent API (or I'll do it once I get around to it, but no promises there about the timeline, since there are a lot of other issues I have to tackle first :grimacing:)

omuomugin:
Thanks for your help.
I succeeded to handle it just like the example you gave me.

One little point to point out was `predicate.test` was actually `predicate.apply` (just in case for other people not to be confused)

However adding these kinds of support to fluent API will help myself so I can give it a try.

I'm thinking of API like below.

```
fields()
.that()
.inSomeCondition()
.should()
.notBeAccessedByMethods()
.that()
.inSomeCondition()
```

Was there any alternative that you were thinking of?

codecholeric:
Oh sorry, I was playing around on the branch https://github.com/TNG/ArchUnit/pull/833 :grimacing: Once that is merged (for ArchUnit 1.0.0rc1 :wink:) the method will be `predicate.test(..)`...
Anyway, glad you could solve it :smiley:

[...]

PR #1105, opened 2023-04-28
add support to detect accesses of fields by specific methods

leonardhusmann:
This PR adds a method to the public API of `FieldsShould` to check if a field has been accessed by methods matching a given predicate. This is checked by a new `ArchCondition`.

I've diverged a little bit from the original suggestion of this issue: I chose to stay consistent with the current API and return an `ArchCondition<JavaField>` to allow method chaining, e.g.

```
fields().that().inSomeCondition()
    .should().notBeAccessedByMethodsThat(...)
    .andShould().notBeFinal()
    .andShould()....
```

What do you think? Any feedback is highly appreciated :)

Resolves: #857

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/lang/conditions/ArchConditions.java
+++ b/archunit/src/main/java/com/tngtech/archunit/lang/conditions/ArchConditions.java
@@ -1274,6 +1274,21 @@ public final class ArchConditions {
                 origin.is(matching(JavaConstructor.class, predicate)), GET_CALLS_OF_SELF);
     }
 
+    @PublicAPI(usage = ACCESS)
+    public static ArchCondition<JavaField> beAccessedByMethodsThat(DescribedPredicate<? super JavaMethod> predicate) {
+        return new ArchCondition<JavaField>("be accessed by methods that " + predicate.getDescription()) {
+            @Override
+            public void check(JavaField field, ConditionEvents events) {
+                field.getAccessesToSelf().stream()
+                        .filter(access -> access.getOrigin() instanceof JavaMethod)
+                        .forEach(access -> {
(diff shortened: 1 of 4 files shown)


codecholeric:
Thanks a lot for adding this! 😃 And sorry that it took me so long to look at it! First of all, what do you mean by

| return an `ArchCondition<JavaField>` to allow method chaining

? I don't see how returning that would allow chaining as in the example? But in any case, what you did looks consistent, so thanks 🙂 Only thing I noticed is that for all the other methods we have a positive and negative form and I realized that there might be use cases where users want to combine this with a `noFields` start of a rule. So I added the inverted version (and adjusted the test, because I think that was a little too much mingling, i.e. it deserves it's own test). What do you think?

Independently of that, retrospectively I should have added something smarter like a generic `not()` function to the API 🙈  Instead of "duplicating" every method. But it's already a little inconsistent between `classes` and `fields` 😒  So maybe at some point we will release a `2.0.0` and clean up / consolidate the rules API again...

codecholeric:
Anyway, what do you think about the review changes in my commit? If everything looks good to you I'd squash everything into one commit...

leonardhusmann:
Hi, thank your for your feedback :smiley: 

> First of all, what do you mean by

This is meant with regards to the initial suggestion in the [issue](https://github.com/TNG/ArchUnit/issues/857) from April '22. What I meant by this is that the arguments (`DescribedPredicate`) are passed as parameters and the function returns the `ArchCondition`. Just as you did it in your first answer in the thread of the issue :smiley:.

Apart from that: thank your for the review changes, they make a lot of sense :) So everything looks good to me :+1:

codecholeric:
Cool, then thanks a lot again, squashed, rebased and now merged 😃

PR #1114, opened 2023-05-26
Add (not)`haveFullyQualifiedNameAnyOf` to API

perlun:
Use case: we wanted to do some whitelisting of certain classes for a particular check, deprecating all usage of a given Java package (`org.joda.time`). We couldn't use existing predicates for this because:

* Existing checks like `areAssignableFrom` and `areAssignableTo` only operate on a single class.
* Also, `assignableFrom` is unsuitable for one of our checks since that would implicitly whitelist the base class.
* There is also `belongToAnyOf` which would be more correct, but it has to be combined with `areNotNestedClasses` in that case to avoid running checks on nested classes. (We run both a "normal" test for the whitelisted classes to allow them to use the deprecated package, but also an "inverted" test to ensure classes don't remain whitelisted when they no longer use the deprecated code. I don't remember if this worked or not, but either way, it feels more complex to have to jump through hoops to do an "exact" check for multiple whitelisted classes)

Hence, the suggested new additions to the API in this PR.

### DIFF
--- a/archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/ShouldOnlyByClassesThatTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/ShouldOnlyByClassesThatTest.java
@@ -86,6 +86,32 @@ public class ShouldOnlyByClassesThatTest {
 
     @Test
     @UseDataProvider("should_only_be_by_rule_starts")
+    public void haveFullyQualifiedNameAnyOf(ClassesThat<ClassesShouldConjunction> classesShouldOnlyBeBy) {
+        Set<JavaClass> classes = filterClassesAppearingInFailureReport(
+                classesShouldOnlyBeBy.haveFullyQualifiedNameAnyOf(Foo.class.getName()))
+                .on(ClassAccessedByFoo.class, Foo.class,
+                        ClassAccessedByBar.class, Bar.class,
+                        ClassAccessedByBaz.class, Baz.class);
+
+        assertThatTypes(classes).matchInAnyOrder(
+                ClassAccessedByBar.class, Bar.class,
+                ClassAccessedByBaz.class, Baz.class);
+    }
+
+    @Test
+    @UseDataProvider("should_only_be_by_rule_starts")
+    public void doNotHaveFullyQualifiedNameAnyOf(ClassesThat<ClassesShouldConjunction> classesShouldOnlyBeBy) {
+        Set<JavaClass> classes = filterClassesAppearingInFailureReport(
+                classesShouldOnlyBeBy.doNotHaveFullyQualifiedNameAnyOf(Foo.class.getName()))
+                .on(ClassAccessedByFoo.class, Foo.class,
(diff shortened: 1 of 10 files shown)


archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/ShouldOnlyByClassesThatTest.java
```diff
@@ -84,6 +84,32 @@ public void doNotHaveFullyQualifiedName(ClassesThat<ClassesShouldConjunction> cl
...
+    public void doNotHaveFullyQualifiedNameAnyOf(ClassesThat<ClassesShouldConjunction> classesShouldOnlyBeBy) {
+        Set<JavaClass> classes = filterClassesAppearingInFailureReport(
+                classesShouldOnlyBeBy.doNotHaveFullyQualifiedNameAnyOf(Foo.class.getName()))
+                .on(ClassAccessedByFoo.class, Foo.class,
+                        ClassAccessedByBar.class, Bar.class,
+                        ClassAccessedByBaz.class, Baz.class);
+
+        assertThatTypes(classes).matchInAnyOrder(ClassAccessedByFoo.class, Foo.class);
+    }
+
```
perlun, line 112: I added tests to the best of my knowledge, and as `ClassesThatTestsExistTest` instructed me. Please let me know if there are any tests missing, I'm not sure this is enough. (feel free to just fix them yourself if it's easier that way :+1:)

The tests added are trivial since I could just copy existing single-`haveFullyQualifiedName` tests.

perlun:
@hankem @codecholeric Please let me know what to do to push this forward.

perlun:
@hankem @codecholeric Ping, please help me push this towards completion. :relaxed: :bow:

hankem:
I'm really sorry! These are crazy times... it's nothing with your PR, but I was occupied with several other projects that didn't allow me to follow up more with ArchUnit than a bare minimum. I know, it's a bad excuse, but I can only say that you're not forgotten.

codecholeric:
Sorry for the late answer! There's just a couple of things coming to my mind.
First, you could already achieve the same by just chaining the existing method with an `or()`, right? :thinking: (or by using the regex method chaining quoted fully qualified names with a `|` :joy:) Does the varargs method provide sufficient value over that?
Second, if we would add this method, should it then maybe also take `Collection`? Is the point of the new method to be used programmatically, passing on a set of class names? Then maybe the `Collection` interface is even nicer than varargs which forces the use of an array?
Third, if we would add this method, should we then deprecate `haveFullyQualifiedName(..)`? Since the new method would be a superset anyway...
Fourth, wouldn't `haveAnyFullyQualifiedNameOf(a, b, c)` be more consistent with the existing API?. Because in other places ArchUnit's rule API usually sounds like a "human spoken sentence" (at least to me), like I would explain in a discussion "classes that reside in a package foo should be annotated with bar". I can't picture myself saying "should have fully qualified name any of [foo, bar]", so to me it sounds inconsistent with the pattern that the other methods follow :shrug:

hankem:
Closing due to inactivity. Feel free to reopen!

Issue #723, opened 2021-11-21
Support an easy way to obtain all involved `JavaClasses` of a generic signature


codecholeric:
There should be an easy way to obtain all `JavaClasses` that are involved in composing a generic type signature.

E.g.

```
class SomeClass {
  Map<? super String, ? extends Serializable> method() {..}
}
```

Then there should be some API to derive the set of classes `[Map, String, Serializable]` from `method.getReturnType()`. We already have some logic like this to calculate the dependencies of a class, since all these classes are also dependencies of the generic return type.

Open question: How should this method be named and where should it live? Should it be something like

```
JavaType.getRawTypesInvolvedInSignature()
```

? What would be a good name for this?

crizzis:
Hi! I'm thinking of giving it a go, could you describe the use case, though?

I was thinking that perhaps it would be somewhat more useful to be able to write sth like: 

```
methods().withReturnTypeThat().hasRawType(Map.class).and().hasNthGenericParameterThat(0).hasUpperBound(String.class)
```

WDYT?

For this to work, I believe we would need a new `TypeThat` DSL (perhaps with a bridge of some sort to `ClassThat`, but I'm not sure). 

Also, since collections are going to be common use cases, probably some shorthand methods like `isCollectionOfTypeThat()`, `isMapOfTypeThat()`, `isCollectionOfRawType(String)` would be in order.

Airblader:
> could you describe the intended use case, though?

This idea originated from Apache Flink where we have annotations for API stability, and we want a rule that makes sure that all types involved in a stable API are marked stable themselves. So we want to express something like

```
Methods
  that are declared in classes annotated with @Stable
  have parameter types that are annotated with @Stable
  and have a return type that is annotated with @Stable
  and have generic type arguments that are annotated with @Stable
```

crizzis:
@Airblader This is a very specific use case, though, and I'm not sure `JavaType.getRawTypesInvolvedInSignature()` is the right approach. `List<MyStableType>` would give you `[List.class, MyStableType.class]`; would you then expect `List` to be `@Stable`? 

Also, `? super MyStableType` is not necessarily `@Stable` itself. 

[...]

PR #1116, opened 2023-05-30
Allow to retrieve involved raw types of `JavaType` and `JavaMember`

leonardhusmann:
This PR adds two new methods: `JavaType.getAllInvolvedRawTypes()` and `JavaCodeUnit.getAllInvolvedRawTypes()`.
`JavaType.getAllInvolvedRawTypes()` returns all involved types in this `JavaType` (e.g. return type of a method)
i.e. for a method `public Map<? extends String, ? super Map<Serializable, Object>> method(List<String> input) {...}`
the methods should return the following:
- `method.getReturnType().getAllInvolvedRawTypes()` would return `Set.of(Map.class, String.class, Serializable.class, Object.class)`
- `method.getAllInvolvedRawTypes()` would return `Set.of(Map.class, String.class, Serializable.class, Object.class, List.class)`

WDYT? Feedback is highly appreciated :)

Resolves: #723

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
@@ -656,6 +656,11 @@ public final class JavaClass
         return this;
     }
 
+    @Override
+    public Set<JavaClass> getAllInvolvedRawTypes() {
+        return ImmutableSet.of(getBaseComponentType());
+    }
+
     @PublicAPI(usage = ACCESS)
     public Optional<JavaClass> getRawSuperclass() {
         return superclass.getRaw();
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaCodeUnit.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaCodeUnit.java
@@ -20,6 +20,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Stream;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
@@ -43,6 +44,7 @@ import static com.google.common.collect.Sets.union;
 import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
 import static com.tngtech.archunit.core.domain.Formatters.formatMethod;
 import static com.tngtech.archunit.core.domain.properties.HasName.Utils.namesOf;
+import static java.util.stream.Collectors.toSet;
 
 /**
  * Represents a unit of code containing accesses to other units of code. A unit of code can be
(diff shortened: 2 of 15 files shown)


codecholeric:
Thanks a lot for your contribution! 😃 The code looks quite clean and works 🙂 
To save some time I simply adjusted the code directly and created two commits from it.

Here are the bigger things I changed:

* I renamed `JavaCodeUnit.getAllTypesInvolvedInSignature()` to simply `JavaCodeUnit.getAllInvolvedRawTypes()`, because IMHO there is no other way how the "raw types involved in a method" could be understood anyway.
  Instead I added some more detailed Javadoc to explain what exactly it means
* I pulled up `JavaCodeUnit.getAllInvolvedRawTypes()` to `JavaMember.getAllInvolvedRawTypes()`, because I think it also makes sense for `JavaField` (there the signature is basically simply the type) and then it's more unified
* I made `JavaClass.getAllInvolvedRawTypes()` return the base component type instead of the class itself, because if the `JavaClass` represents an array I think we are interested in the component type and not the array type.
  This method is mainly about finding dependencies and the array type is usually not the dependency we're interested in, but the component type. I might be wrong there, maybe it's array type and component type, but I'd wait for
  a concrete issue before adjusting it :grin:

But I'm gonna try to give you some feedback on the stuff I noticed so you can hopefully also take some stuff away 🙂 

First thing I noticed was the commit structure. For ArchUnit we try to create self-contained logical units.
At least to be ready to merge, the history shouldn't look like "this is what I did in which order", but "these are logical, cohesive and self-contained steps to reach the goal".
These steps should be as small as possible, but still contain a logical unit. For example, a commit "add tests" i.g. doesn't make any sense IMHO.
Because it means the commits before were incomplete. For me the tests are an inseparable requirement for the implementation.
You should be able to check out any commit and the code base in this state should be complete, reasonable and fully tested. 

Another thing I noticed was that Javadoc was missing. We don't create Javadoc in a cargo cult way, but every non-trivial method should get it.
And "involved raw types" isn't self-explanatory enough from my point of view to be understood i.g. without any Javadoc.
BTW: Public methods also need an explicit annotation `@PublicAPI` or the tests will fail 😉 

Regarding the tests, there are a lot of custom assertions that make your life simpler. We never use `JUnit` `assertEquals`, but only AssertJ, since it's nicer to read and offers autocomplete. But for all domain objects where we have recurring assertions we have custom matchers, e.g. `assertThatTypes(..)` which then directly allows to assert class objects.
I also thought that in some places the test cases could have been a little more complex, so I adjusted this.
You could also try test driven development for these nice self-contained cases, because for some code I saw that test cases were missing.
E.g. you implemented stuff in `JavaGenericArrayType`, but I didn't see any test with a generic array. If you would do this test driven,  then such a thing couldn't happen 🤷‍♂️ 
Also, in tests you don't need to import every single type involved, because ArchUnit has an automatic resolution process that will import transitive stuff as needed. So it's always worth a shot if `importClass(Single.class)` is maybe already good enough for the test case.

Last but not least I saw quite some unrelated code reformatting. IMHO you shouldn't touch unrelated code like this, definitely not in the same commit as "real" changes, cause it just makes the review harder (and reasoning about the change later on as well).
If you think some formatting is suboptimal we can discuss it and potentially adjust it, but it should at least be a completely separate commit then with a message explicitly describing that this is a pure formatting change.
Nevertheless, I also saw that a fresh import of the IntelliJ code style meanwhile sets hard wrap by default to 150 chars 🙈 
This was different in earlier times, so I adjusted the code formatter for IntelliJ in a pre-commit to fix the hard wrap to something so big that we effectively just decide about the line breaks.
Not sure if this is i.g. the best thing, or if we should have a more restrictive hard wrap, but so far I also don't remember this to caues any problems 🤷‍♂️ 

BTW: We also don't use any branch prefix convention 😁 E.g. in the PR title...

codecholeric (approved):

leonardhusmann:
Wow! thanks a lot for this extensive feedback :smile::pray::pray:

PR #1131, opened 2023-07-04
Let `JarFileLocation` work with custom `ClassLoader` URIs

odrotbohm:
IMPORTANT: I did not add any new test cases to verify the scenario described in the following, as I was unable to get the project imported into my IDE. The existing tests still succeed so that I feel confident I didn't break anything. Furthermore, a downstream sample project producing the described scenario works with the fix applied and previously failed.

Some ClassLoaders that work with repackaged JAR files return custom resource URIs to indicate custom class loading locations. For example, the ClassLoader in a packaged Spring Boot's returns the following URI for source package named example: jar:file:/Path/to/my.jar!/BOOT-INF/classes!/example/. Note the second "!/" to indicate a classpath root.

Prior to this commit, JarFileLocation was splitting paths to a resource at the first "!/" assuming the remainder of the string would depict the actual resource path. That remainder potentially containing a further "!/" would prevent the JAR entry matching in FromJar.classFilesBeneath(…) as the entries themselves do not contain the exclamation mark.

This commit changes the treatment of the URI in JarFileLocation to rather use the *last* "!/" as splitting point so that the remainder is a proper path within the ClassLoader and the matching in FromJar.classFilesBeneath(…) works properly.

### DIFF
--- a/archunit-3rd-party-test/build.gradle
+++ b/archunit-3rd-party-test/build.gradle
@@ -1,0 +1,16 @@
+plugins {
+    id 'archunit.java-conventions'
+}
+
+ext.moduleName = 'com.tngtech.archunit.thirdpartytest'
+
+dependencies {
+    testImplementation project(path: ':archunit', configuration: 'shadow')
+    testImplementation project(path: ':archunit', configuration: 'tests')
+    testImplementation dependency.springBootLoader
+    dependency.addGuava { dependencyNotation, config -> testImplementation(dependencyNotation, config) }
+    testImplementation dependency.log4j_slf4j
+    testImplementation dependency.junit4
+    testImplementation dependency.junit_dataprovider
+    testImplementation dependency.assertj
+}
--- a/archunit-3rd-party-test/src/test/java/com/tngtech/archunit/core/importer/SpringLocationsTest.java
+++ b/archunit-3rd-party-test/src/test/java/com/tngtech/archunit/core/importer/SpringLocationsTest.java
@@ -1,0 +1,96 @@
+package com.tngtech.archunit.core.importer;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URL;
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.function.Function;
+import java.util.jar.JarFile;
+import java.util.stream.Stream;
+
+import com.tngtech.archunit.core.importer.testexamples.SomeEnum;
+import com.tngtech.archunit.testutil.SystemPropertiesRule;
+import com.tngtech.java.junit.dataprovider.DataProvider;
+import com.tngtech.java.junit.dataprovider.DataProviderRunner;
+import com.tngtech.java.junit.dataprovider.UseDataProvider;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.springframework.boot.loader.LaunchedURLClassLoader;
+import org.springframework.boot.loader.archive.Archive;
+import org.springframework.boot.loader.archive.JarFileArchive;
+
+import static com.google.common.collect.Iterators.getOnlyElement;
+import static com.google.common.collect.MoreCollectors.onlyElement;
+import static com.google.common.collect.Streams.stream;
+import static com.google.common.io.ByteStreams.toByteArray;
+import static com.tngtech.archunit.core.importer.LocationTest.classFileEntry;
+import static com.tngtech.archunit.core.importer.LocationTest.urlOfClass;
+import static com.tngtech.archunit.core.importer.LocationsTest.unchecked;
+import static com.tngtech.java.junit.dataprovider.DataProviders.testForEach;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(DataProviderRunner.class)
+public class SpringLocationsTest {
+    /**
+     * Spring Boot configures some system properties that we want to reset afterward (e.g. custom URL stream handler)
+     */
+    @Rule
+    public final SystemPropertiesRule systemPropertiesRule = new SystemPropertiesRule();
+
+    @DataProvider
+    public static Object[][] springBootJars() {
+        Function<Function<TestJarFile, TestJarFile>, TestJarFile> createSpringBootJar = setUpJarFile -> setUpJarFile.apply(new TestJarFile())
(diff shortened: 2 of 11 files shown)


odrotbohm:
@codecholeric @hankem – This was originally reported against Spring Modulith in spring-projects/spring-modulith#221. The fix would be crucial to get our runtime support working. I'd appreciate any estimations on a release including that fix (potentially rather a 1.0.2 than a 1.1?) as we find ourselves blocked from releasing Spring Modulith 1.0 RC1.

codecholeric:
Hey @odrotbohm, again, sorry that it took me so long right now :see_no_evil: My problem was that I couldn't really overview the whole impact, because I never anticipated multiple separators `!/` within the `ClassLoader` URLs. Thus, I wanted to be careful that I really understand this and the impact.
To understand it I decided to add some real "3rd party tests" for this, using the Spring Boot `ClassLoader`, but then my journey down the rabbit hole began :grimacing: I didn't know you could customize URL handling that heavily (and transparently), and once I saw that there was a custom Spring `JarFile`, `JarURLConnection`, `URLStreamHandler` and so on I started to ponder if this should even go into core ArchUnit, or if it would be a cleaner approach to define some extension point that Modulith could implement to add this custom URL handling in any way that makes sense. I'm still not completely sure about this, but I figured that using the last separator `!/` for now doesn't destroy anything and if it fixes the problem at hand, then that's an okay compromise between "being clean" and being pragmatic :shrug: 

Anyway, there is one case that currently doesn't work though (I've documented this in a test). If the directory entries within the JAR file are missing, then the respective classes within a certain package won't be found correctly. I had this problem in the past (e.g. the old `rt.jar` decided to leave out the directory entries). Back then I implemented some workaround to scan through all the JARs on the classpath as well and look for entries that start with the respective prefix. But this won't work in this case, because without custom knowledge about the Spring Boot JAR layout it's impossible to know that entries need to be taken relative to `BOOT-INF/classes` only. And that IMHO crosses the point where it should really be a framework dependent extension point, if we want to support this, too.
But since I don't even know if there ever are Spring Boot JARs out there that miss the directory entries (e.g. if the directory entry `BOOT-INF/classes/` is missing AFAIS deriving the nested archive crashes anyway), I'm also fine to add it like that for now and see if there ever will be a follow up issue. Except if you with your background knowledge about Spring can already tell that this will definitely be a problem, then we can ponder if we improve this right away. But then also maybe in a follow up PR so the problem at hand is fixed at least.

Anyway, I tried to make some improvements to the test infrastructure on the way and then added some more tests, I'd be happy if you can take a look if this makes sense to you. The last commit I would squash onto your's if that's alright, since it adds extra tests for the concrete fix.

In any case, I think it would make sense to add to your commit message also a note, that using 
```
jar:file:/foo/bar.jar!/BOOT-INF/classes!/
```
as a root JAR URL (where originally only a real root URL of a JAR file was supported, e.g. `jar:file:/foo/bar.jar`) works because of the custom Spring Boot `URLStreamHandler` magic. Because that was the most puzzling for me in the beginning, why just using this "composite" JAR URL as `new URL(..)` would suddenly be able to derive the nested archive correctly as `ClassFileSource` :joy: Took me quite some time to find that piece of the puzzle.

codecholeric:
PS: Aaaand I see that my tests of course fail on some platforms :man_facepalming: URL handling is so much fun to do cross-platform and across all JDK versions :grimacing: Gonna look into that tomorrow! I think you can still review it if it generally makes sense from your point of view...

codecholeric:
I think by looking at the test fails I actually found a way to fix the "classes in packages can't be found if there are no directory zip entries" problem :slightly_smiling_face: Because, there was a "legacy way" that actually looked through the context for `URLClassLoaders` and considered all their URLs as part of the classpath. For some reason I removed this behavior >= JDK 9 (I think it was only because I thought the `URLClassLoader` would lose it's wide usage post JDK 8 and it wouldn't make a difference anyway cause the classpath property typically has the same URLs). 
But with this behavior in place we actually obtain URLs like `jar:file:/some/path.jar!/BOOT-INF/classes!/` from the Spring Boot `ClassLoader`, and these again with the custom Spring Boot URL handling in place will stream their entries correctly so that also classes in packages without directory zip entry will then be found :partying_face: (at least that's what the tests so far suggest to me...)
Anyway, let me know what you think! If all this makes sense for you I would rearrange the commit history a little once again to put the legacy behavior in action for JDK > 8 in a separate commit and with a clear explanation why.

odrotbohm:
I know too little about the way Boot sets up the class loading in the fat JARs they create. I've asked @wilkinsona to comment, but am not sure when he's going to find time.

Personally, I don't have any objectives except getting ArchUnit to work with classes in a running Spring Boot application. Given that this particular context should be a rather rare case for now (and only affects projects using Spring Modulith's runtime support), I agree that the focus should be on making sure that the currently already working scenarios still continue to work. In other words: if we do not catch all corner cases of the “run ArchUnit within a Boot app” is just fine.

wilkinsona:
The changes here look good to me. I can't see anything that's Spring Boot specific in the main code and that's the way it should be.

Regarding jars without directory entries, that can happen with Spring Boot. It shouldn't happen with the "outer" jar as it's created by our build plugins and they will create directory entries. It may happen with jars nested in `BOOT-INF/lib`, however, as we do not have any control over their contents.

The code in Spring Framework's `PathMatchingResourcePatternResolver` may be of interest here. For example, its `addAllClassLoaderJarRoots` method is similar to the changes that are proposed here for working with `URLClassLoader` and finding jars on the classpath.

codecholeric:
Thanks for the feedback, then I'm gonna get this ready and create a bugfix release!

codecholeric:
Okay, as a last test I've now added a JAR without directory entries to the `BOOT-INF/lib` folder and that class also could be found correctly. So I would expect it's at least a reasonable shot now :grin: (since it's URL handling I'm already waiting for the first "but on platform x with version y" bug report, but hey... One issue at a time :grin:)

codecholeric:
Just FYI: I decided to just release 1.1.0 with this fix. Because some stuff has accumulated meanwhile and I don't think it makes sense to separate this for a bug fix release (the other changes are either also bug fixes or "low danger" improvements)

odrotbohm:
Confirming 1.1.0 solves the original problem! Thanks, Pete! 🥳🙇

PR #1138, opened 2023-07-15
small correction for Predefined Predicates and Conditions documentation

SimonVerhoeven:
whilst the sample code indeed compiles, the added condition will not be used.

### DIFF
--- a/docs/userguide/007_The_Lang_API.adoc
+++ b/docs/userguide/007_The_Lang_API.adoc
@@ -211,7 +211,7 @@ JavaClass.Predicates.type(Serializable.class).and(HasName.Predicates.name(""))
 
 // Does compile, because the compiler now sees name(..) as a predicate for JavaClass
 DescribedPredicate<JavaClass> name = HasName.Predicates.name("").forSubtype();
-name.and(JavaClass.Predicates.type(Serializable.class));
+name = name.and(JavaClass.Predicates.type(Serializable.class));
 ----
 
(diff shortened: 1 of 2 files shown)


codecholeric:
Hey, thanks for the contribution and sorry for the late reply. I'm not sure I'm getting the intent of the PR though. The whole point of the snippet was that it compiles. Why do you think it's important for the point the docs are making here that this condition is used? The other statements further up also don't use the `Predicate` for anything 🤷
Honestly, to me it looks worse to write `name = name.and(type(Foo))` (because in fact this would now be `nameAndType`) than just keeping the statement as it was. And as I said, it's definitely consistent the way it was before with the way the snippets further up are written...

SimonVerhoeven:
No problem, and indeed compilation-wise, it's completely valid and also for the point being made. 
The intent of the PR was more to have the code be fully correct, since as it stood, the result of the operation was not being used.

codecholeric:
But the code is still "fully correct", no? 🤔 Depending on what you call "correct". But again, the snippets above the changed one follow the same pattern, so you'd have to change them all, and on the other hand you would need to rename the variable name of the assignment, since otherwise the name would be misleading 🤷 
But from my point of view that doesn't really help the point of this doc snippet, cause as mentioned it focusses on pure compilability. What you'd do with these predicates in a real life scenario is something completely different, but also not relevant for the type compatibility and compilability discussion 🤷

SimonVerhoeven:
True, I get what you mean. From a showcase of the compatibility, it's fine, and my merge might indeed introduce more complexity.

PR #1140, opened 2023-07-16
updated documentation since accept no longer exists on JavaPackage

SimonVerhoeven:
The `PackageVisitor` could also be replaced with a method reference, but I left that for clarity's sake.


codecholeric:
Thanks for fixing this! 🙂 You're right, I overlooked that part when I renamed this when switching to `1.0.0` 🙈

PR #1156, opened 2023-08-22
Merge release 1.1.0 back into `main`

codecholeric:


hankem:
Ah, sorry that I had merged PR #1154 before this one... 🙈

codecholeric:
Well, my fault that I forgot to create this PR right after the release. I think we should add that to the automation :man_shrugging: Maybe even with another action that merges it automatically once it's green :shrug:

Issue #1157, opened 2023-08-23
Maven ignores tests defined as ArchTest fields


Scaronthesky:
I'm under the impression that Maven seems to ignore ArchUnit tests which are defined as fields with the @ArchTest annotation. 

I copied some of the CodingRulesTest from the ArchUnit-Examples project, e. g. "No Access to Standard streams". These tests work fine when I execute them with IntelliJ. Maven completely ignores them, though. Maven executes only tests which are defined as "proper" test methods. 

Is this a configuration issue or a bug? I had a look at the documentation and issues but could not find anything.

```
import static com.tngtech.archunit.library.GeneralCodingRules.NO_CLASSES_SHOULD_ACCESS_STANDARD_STREAMS;
import com.tngtech.archunit.core.domain.JavaClasses;
import com.tngtech.archunit.junit.AnalyzeClasses;
import com.tngtech.archunit.junit.ArchTest;
import com.tngtech.archunit.lang.ArchRule;

@AnalyzeClasses(packages = "de.package")
public class CodingRulesTest {

  // Will silently be ignored by Maven
  @ArchTest
  public final ArchRule no_access_to_standard_streams = NO_CLASSES_SHOULD_ACCESS_STANDARD_STREAMS;

  // Will be executed
  @ArchTest
  public void assertNoAccessToStandardStreams(JavaClasses classes) {
    NO_CLASSES_SHOULD_ACCESS_STANDARD_STREAMS.check(classes);
  }
}
```
In the examples project the ArchRule fields are private. I made the public but it makes no difference. They are still ignored.

maven-surefire-plugin:3.1.2
Using configured provider org.apache.maven.surefire.junitplatform.JUnitPlatformProvider
com.tngtech.archunit:archunit-junit5-api:1.1.0
com.tngtech.archunit:archunit-junit5-engine:1.1.0
org.junit.jupiter:junit-jupiter-api:5.10.0

hankem:
Thanks for raising the issue! I couldn't exactly reproduce your problem, but saw that ArchUnit 1.1.0 is not fully compatible with [JUnit 5.10](https://junit.org/junit5/docs/5.10.0/release-notes/), so I assume that your issue is related to that.

<details><summary>This example</summary>

```sh
git clone https://github.com/TNG/ArchUnit-Examples.git

cd ArchUnit-Examples/example-junit5

cat > pom.xml << EOF
<?xml version="1.0"?>
<project>
[...]

PR #1159, opened 2023-08-25
Support JUnit Platform 1.10.0

hankem:
The new `getAncestors()` method in `TestDescriptor` (https://junit.org/junit5/docs/5.10.0/release-notes/) 
was obviously not available in
```java
class com.tngtech.archunit.junit.internal.AbstractArchUnitTestDescriptor
extends org.junit.platform.engine.support.descriptor.AbstractTestDescriptor
```
from `org.junit.platform:junit-platform-engine:1.9.x`.

Solves #1157


codecholeric:
Ha, for the first time I remembered that we also have to adjust `release_check/archunit-junit5-engine-api.pom` :joy: 
But thanks a lot! I saw the other issue, I guess this is one of those classpath fun times, probably even changing the order of `archunit-junit5` and `jupiter` in the Maven POM would have been sufficient to stop the problem if one is backwards compatible :zany_face:

codecholeric (approved):

PR #1184, opened 2023-11-01
Issue 1132: Ignore unknow components in dependencies

tfij:


archunit/src/main/java/com/tngtech/archunit/library/plantuml/rules/PlantUmlPatterns.java
```diff
@@ -46,7 +48,7 @@ class PlantUmlPatterns {
     private static final String COLOR_FORMAT = "\\s*(?:#" + anyOf("\\w|/\\\\-") + "+)?";
 
     private static final Pattern PLANTUML_COMPONENT_PATTERN = Pattern.compile(
-            "^\\s*" + COMPONENT_NAME_FORMAT + "\\s*" + STEREOTYPE_FORMAT + "*" + ALIAS_FORMAT + COLOR_FORMAT + "\\s*");
+            "^\\s*" + COMPONENT_TYPE_FORMAT + "\\s*"  + COMPONENT_NAME_FORMAT + "\\s*" + STEREOTYPE_FORMAT + "*" + ALIAS_FORMAT + COLOR_FORMAT + "\\s*");
```
codecholeric, line 51: I think this should (at least) be a separate commit. Because it has nothing to do with the PR title "ignore unknown components in dependencies", right?

I guess you added this, because it seems inconsistent once you add `database ...` declarations to declare components just as `[foo] ...`?

I'm principally fine with supporting the `component` keyword, but then I think we should support it fully? Because if you use the keyword, you can also omit the square brackets, right? I.e. `component foo <..bar..>`. So, we would have to support both cases, with and without square brackets. (That's exactly one reason why the original supported syntax was so limited though 😉 To have a clearly defined and easy to comprehend subset without too many extra cases...)

codecholeric: I guess once we open that box we also need to support dependencies that are defined from those components declared without square brackets...

```
component foo <<..any..>>
[bar] <<..other..>>

foo --> [bar]
```

Not sure, if that's the case already? 🤔  I guess it would already be valid PlantUML to define

```
[foo] <<..any..>>
[bar] <<..other..>>

foo --> bar
```

🙄

archunit/src/test/java/com/tngtech/archunit/library/plantuml/rules/PlantUmlParserTest.java
```diff
@@ -418,6 +406,54 @@ public void parses_a_component_diagram_that_uses_alias_with_and_without_brackets
...
+
+        PlantUmlComponent a = getComponentWithAlias(new Alias("aliasA"), diagram);
+        PlantUmlComponent b = getComponentWithAlias(new Alias("aliasB"), diagram);
+
+        assertThat(a.getDependencies()).containsOnly(b);
+        assertThat(b.getDependencies()).isEmpty();
+    }
+
+    @Test
+    public void components_with_component_keyword() {
```
codecholeric, line 441:
```suggestion
    public void supports_components_declared_with_component_keyword() {
```

codecholeric (changes requested):
Thanks a lot for adding this!! 😃 🙏 Looks quite good, I just have some comments you can check out. One bigger is around adding the `component` keyword support I guess, because for me that's a slightly different issue and should IMHO be split from ignoring dependencies from unknown components.

archunit/src/main/java/com/tngtech/archunit/library/plantuml/rules/PlantUmlComponents.java
```diff
@@ -43,15 +47,16 @@ Collection<PlantUmlComponent> getComponentsWithAlias() {
         return componentsByAlias.values();
     }
 
-    PlantUmlComponent findComponentWith(String nameOrAlias) {
+    Optional<PlantUmlComponent> findComponentWith(String nameOrAlias) {
```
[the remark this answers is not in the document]
tfij: fixed

archunit/src/main/java/com/tngtech/archunit/library/plantuml/rules/PlantUmlComponents.java
```diff
@@ -43,15 +47,16 @@ Collection<PlantUmlComponent> getComponentsWithAlias() {
...
         ComponentName componentName = new ComponentName(nameOrAlias);
         Alias alias = new Alias(nameOrAlias);
         PlantUmlComponent result = componentsByAlias.containsKey(alias) ? componentsByAlias.get(alias) : componentsByName.get(componentName);
         if (result == null) {
-            throw new IllegalDiagramException("There is no Component with name or alias = '%s'. %s", nameOrAlias,
-                    "Components must be specified separately from dependencies.");
+            log.info("There is no Component with name or alias = '{}'. " +
```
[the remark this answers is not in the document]
tfij: fixed

archunit/src/main/java/com/tngtech/archunit/library/plantuml/rules/PlantUmlParser.java
```diff
@@ -78,9 +78,11 @@ private Set<PlantUmlComponent> parseComponents(List<String> plantUmlDiagramLines
     private ImmutableList<ParsedDependency> parseDependencies(PlantUmlComponents plantUmlComponents, List<String> plantUmlDiagramLines) {
         ImmutableList.Builder<ParsedDependency> result = ImmutableList.builder();
         for (PlantUmlDependencyMatcher matcher : plantUmlPatterns.matchDependencies(plantUmlDiagramLines)) {
-            PlantUmlComponent origin = findComponentMatching(plantUmlComponents, matcher.matchOrigin());
-            PlantUmlComponent target = findComponentMatching(plantUmlComponents, matcher.matchTarget());
-            result.add(new ParsedDependency(origin.getIdentifier(), target.getIdentifier()));
+            Optional<PlantUmlComponent> origin = findComponentMatching(plantUmlComponents, matcher.matchOrigin());
+            Optional<PlantUmlComponent> target = findComponentMatching(plantUmlComponents, matcher.matchTarget());
+            if (origin.isPresent() && target.isPresent()) {
```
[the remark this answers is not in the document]
tfij: log added

archunit/src/main/java/com/tngtech/archunit/library/plantuml/rules/PlantUmlPatterns.java
```diff
@@ -34,6 +34,8 @@
 import static java.util.Collections.singletonList;
 
 class PlantUmlPatterns {
+    private static final String COMPONENT_TYPE_GROUP_NAME = "componentType";
+    private static final String COMPONENT_TYPE_FORMAT = capture("component", COMPONENT_TYPE_GROUP_NAME) + "?";
```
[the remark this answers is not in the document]
tfij: separated commit https://github.com/TNG/ArchUnit/pull/1223/commits/39190ba798e524f8bd57472e717a0f0653a3c4f3
outside of this PR (https://github.com/TNG/ArchUnit/pull/1223)

archunit/src/main/java/com/tngtech/archunit/library/plantuml/rules/PlantUmlPatterns.java
tfij: separated commit https://github.com/TNG/ArchUnit/pull/1223/commits/39190ba798e524f8bd57472e717a0f0653a3c4f3
outside of this PR (https://github.com/TNG/ArchUnit/pull/1223)

archunit/src/test/java/com/tngtech/archunit/library/plantuml/rules/PlantUmlParserTest.java
tfij: fixed and moved to separated PR
commit: https://github.com/TNG/ArchUnit/pull/1223/commits/39190ba798e524f8bd57472e717a0f0653a3c4f3
PR: (https://github.com/TNG/ArchUnit/pull/1223)

tfij:
@codecholeric all comments applied

codecholeric (approved):

PR #1200, opened 2023-11-20
Add possibility to check for overriding methods to domain and lang

### LINKED ISSUE #1198: Add predicate to find overridden methods
Because I implement/copy it now probably for the third time: It would be great if ArchUnit would provide methods/predicates with which you can check whether a method overrides / implements a method from a base class or from an interface.

As suggested in https://github.com/TNG/ArchUnit/issues/359#issuecomment-977576053:
* `methods().that().areOverridden()`
* `methods().should().beOverridden()`
* `JavaMethod.isOverridden()`
* `JavaMethod.Predicates.overridden()`

rweisleder:
Resolves: #1198

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaMethod.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaMethod.java
@@ -21,18 +21,29 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.function.Supplier;
+import java.util.stream.Stream;
 
 import com.tngtech.archunit.PublicAPI;
 import com.tngtech.archunit.base.ArchUnitException.InconsistentClassPathException;
+import com.tngtech.archunit.base.DescribedPredicate;
 import com.tngtech.archunit.base.MayResolveTypesViaReflection;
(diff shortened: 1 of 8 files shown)


rweisleder:
I just saw, there is another issue/PR #1040 for the same topic 😅
At the moment, I haven't looked more deeply into whether my implementation is too simple or the other one too verbose.

codecholeric:
Thanks a lot for your contribution and sorry for the slow process 🙈 I see that there is another PR going in the same direction, but from my point of view it a) looks further away from being finished and b) it seems quite abandoned 🤔 So, I think it makes sense to bring this PR here over the finish line.
I think there is still one problem in there with Generics and bridge methods, i.e. when overriding a generic interface, then the bridge method works alright being detected as overriding, but the concrete method containing the fixed type is not detected as overriding. I added a test for this. How do you see this? To me it seems like the concrete overriding method should also be detected as such, right? 🤔

rweisleder:
Generics, how could I have missed them 🤢 You're absolutely right, these are valid cases as well. Maybe I'll find time to complete this soon

> sorry for the slow process

I'm part of the slow process as well, no worries

Issue #1194, opened 2023-11-16
Exception in thread "main" java.lang.IllegalStateException: Never found a JavaCodeUnit that matches supposed origin CodeUnit


heyuxiang1996:
```Exception in thread "main" java.lang.IllegalStateException: Never found a JavaCodeUnit that matches supposed origin CodeUnit{name='access$084', descriptor=(Lcom/transsion/auto/bluetooth/BluetoothUtils;Ljava/lang/Object;)Ljava/lang/String;, declaringClassName='com.transsion.auto.bluetooth.BluetoothUtils'}
	at com.tngtech.archunit.core.importer.RawAccessRecord$CodeUnit.resolveFrom(RawAccessRecord.java:132)
	at com.tngtech.archunit.core.importer.AccessRecord$Factory.lambda$createOriginSupplier$0(AccessRecord.java:294)
	at com.tngtech.archunit.thirdparty.com.google.common.base.Suppliers$NonSerializableMemoizingSupplier.get(Suppliers.java:181)
	at com.tngtech.archunit.core.importer.AccessRecord$Factory$RawAccessRecordProcessed.getOrigin(AccessRecord.java:255)
	at com.tngtech.archunit.core.importer.ClassGraphCreator.tryProcess(ClassGraphCreator.java:147)
	at com.tngtech.archunit.core.importer.ClassGraphCreator.lambda$completeCodeUnitDependencies$2(ClassGraphCreator.java:131)
	at java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
	at java.util.stream.Streams$StreamBuilderImpl.forEachRemaining(Streams.java:419)
	at java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:647)
	at java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:272)
	at java.util.HashMap$KeySpliterator.forEachRemaining(HashMap.java:1580)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
	at java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
	at java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
[...]

Issue #1146, opened 2023-07-26
Use of += on a private String in a nested class causes IllegalStateException when imported


OneGeek:
JDK version: Amazon Corretto 11.0.16
Java Source/Target version: 8
ArchUnit version: 1.0.1

At least this version of the JDK, when targeting Java 8, replaces uses of += with StringBuilder in a way that results in a synthetic access method being created that ArchUnit doesn't seem to know what to do with.

-----
Here's a class, that when imported, that triggers the error in ArchUnit
```java
public class Outer {
    private String somePrivateString = "Hello";
    
    class Inner {
        private void methodThatAccessesOuterClassPrivateField() {
            somePrivateString += ", world!";
        }
    }
}
```

The following test will fail with the below error
```java
@Test
public void broken() {
    new ClassFileImporter().importClass(Outer.class);
}
```

`java.lang.IllegalStateException: Never found a JavaCodeUnit that matches supposed origin CodeUnit{name='access$084', descriptor=(LOuter;Ljava/lang/Object;)Ljava/lang/String;, declaringClassName='Outer'}`
<details><summary>Full Stack Trace:</summary>

```
java.lang.IllegalStateException: Never found a JavaCodeUnit that matches supposed origin CodeUnit{name='access$084', descriptor=(Lcom/ekotrope/broken/Outer;Ljava/lang/Object;)Ljava/lang/String;, declaringClassName='com.ekotrope.broken.Outer'}

	at com.tngtech.archunit.core.importer.RawAccessRecord$CodeUnit.resolveFrom(RawAccessRecord.java:132)
	at com.tngtech.archunit.core.importer.AccessRecord$Factory.lambda$createOriginSupplier$0(AccessRecord.java:294)
	at com.tngtech.archunit.thirdparty.com.google.common.base.Suppliers$NonSerializableMemoizingSupplier.get(Suppliers.java:183)
	at com.tngtech.archunit.core.importer.AccessRecord$Factory$RawAccessRecordProcessed.getOrigin(AccessRecord.java:255)
	at com.tngtech.archunit.core.importer.ClassGraphCreator.tryProcess(ClassGraphCreator.java:144)
	at com.tngtech.archunit.core.importer.ClassGraphCreator.lambda$completeCodeUnitDependencies$2(ClassGraphCreator.java:129)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
[...]

PR #19, opened 2017-07-16
NPE demonstration

koppor:
I wanted to check the behavior of `noClasses: All classes having an annotation should be the same as no classes with not that annotation. But that test case simply throws an NPE. I don't know the reason for it. Maybe @codecholeric sees it directly.

<details>

```
java.lang.NullPointerException
	at java.util.regex.Matcher.getTextLength(Matcher.java:1283)
	at java.util.regex.Matcher.reset(Matcher.java:309)
	at java.util.regex.Matcher.<init>(Matcher.java:229)
	at java.util.regex.Pattern.matcher(Pattern.java:1093)
	at java.util.Formatter.parse(Formatter.java:2547)
	at java.util.Formatter.format(Formatter.java:2501)
	at java.util.Formatter.format(Formatter.java:2455)
	at java.lang.String.format(String.java:2940)
	at com.tngtech.archunit.lang.ArchCondition.as(ArchCondition.java:65)
	at com.tngtech.archunit.lang.syntax.ArchRuleDefinition$1.apply(ArchRuleDefinition.java:119)
	at com.tngtech.archunit.lang.syntax.ArchRuleDefinition$1.apply(ArchRuleDefinition.java:116)
	at com.tngtech.archunit.lang.syntax.ObjectsShouldInternal$FinishedRule.createCondition(ObjectsShouldInternal.java:100)
	at com.tngtech.archunit.lang.syntax.ObjectsShouldInternal$FinishedRule.get(ObjectsShouldInternal.java:96)
	at com.tngtech.archunit.lang.syntax.ObjectsShouldInternal$FinishedRule.get(ObjectsShouldInternal.java:93)
	at com.google.common.base.Suppliers$MemoizingSupplier.get(Suppliers.java:131)
	at com.tngtech.archunit.lang.syntax.ObjectsShouldInternal.check(ObjectsShouldInternal.java:75)
	at com.tngtech.archunit.lang.syntax.elements.GivenClassesThatTest$Evaluator.on(GivenClassesThatTest.java:481)
	at com.tngtech.archunit.lang.syntax.elements.GivenClassesThatTest.areAnnotatedWith_type_inverse(GivenClassesThatTest.java:225)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #1203, opened 2023-11-26
fix importing inner class using string concat of outer field

codecholeric:
We weren't aware that the compiler occasionally generates synthetic `access$123` methods that call constructors. More precisely for the following constellation
```
class Outer {
  String field = "";
  class Inner {
    void concat() {
      field += "any";
    }
  }
}
```
the compiler generates bytecode instantiating and using a `StringBuilder`. But for constructor calls the synthetic access resolution was not hooked in, because we assumed that those methods would never call a constructor. This in turn lead to the bug
```
java.lang.IllegalStateException: Never found a JavaCodeUnit that matches supposed origin CodeUnit{name='access$123'...
```
I.e. the method `access$123` was filtered out as synthetic, but the origin of the constructor call had not been resolved to the actual `Inner.concat` method.

Resolves: #1146
Resolves: #1194

### DIFF
--- a/archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterSyntheticPrivateAccessesTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterSyntheticPrivateAccessesTest.java
@@ -5,6 +5,7 @@ import java.util.function.Supplier;
 
 import com.tngtech.archunit.core.domain.JavaAccess;
 import com.tngtech.archunit.core.domain.JavaClass;
+import com.tngtech.archunit.core.domain.JavaFieldAccess;
 import org.junit.Test;
 
 import static com.tngtech.archunit.core.domain.JavaConstructor.CONSTRUCTOR_NAME;
@@ -12,6 +13,7 @@ import static com.tngtech.archunit.core.domain.JavaFieldAccess.AccessType.GET;
 import static com.tngtech.archunit.core.domain.JavaFieldAccess.AccessType.SET;
 import static com.tngtech.archunit.testutil.Assertions.assertThatAccesses;
 import static com.tngtech.archunit.testutil.Assertions.expectedAccess;
+import static com.tngtech.archunit.testutil.assertion.AccessesAssertion.access;
 import static java.util.stream.Collectors.toSet;
 
 public class ClassFileImporterSyntheticPrivateAccessesTest {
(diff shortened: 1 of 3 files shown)


archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterSyntheticPrivateAccessesTest.java
```diff
@@ -422,6 +424,40 @@ Supplier<Supplier<Data_of_imports_private_constructor_reference_from_lambda.Targ
         );
     }
 
+    /**
+     * This is a special case, because for =+ concatenation of an outer string from an inner class
+     * the compiler creates a synthetic `access$123` method that creates a new `StringBuilder()`.
+     * Before this we wrongly assumed that all that happens from such `access$123` methods
+     * are field accesses and method calls, but no constructor calls.
```
hankem, line 431: 1. `=+` → `+=` 😉
2. Not all compilers create such a synthetic `access$123` method in this case; for this test's example, I've also seen the following byte code (e.g. with openjdk-17.0.2):
  ```
  private void stringConcat();
    Code:
       0: aload_0
       1: getfield      #1                  // Field this$0:LOuter;
       4: dup
       5: getfield      #13                 // Field Outer.outerPrivateString:Ljava/lang/String;
       8: invokedynamic #19,  0             // InvokeDynamic #0:makeConcatWithConstants:(Ljava/lang/String;)Ljava/lang/String;
      13: putfield      #13                 // Field Outer.outerPrivateString:Ljava/lang/String;
      16: return
  ```
  So I'd replace `the compiler creates` with `the compiler may create` (also in the commit message).
3. The `Before this` reference might not be clear in this JavaDoc; I would drop it here.

```suggestion
     * This is a special case, because for `+=` concatenation of an outer string from an inner class
     * the compiler may create a synthetic `access$123` method that creates a new `StringBuilder()`.
```

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterSyntheticPrivateAccessesTest.java
```diff
@@ -422,6 +424,40 @@ Supplier<Supplier<Data_of_imports_private_constructor_reference_from_lambda.Targ
...
+
+        Set<JavaFieldAccess> fieldAccesses = new ClassFileImporter().importClasses(Outer.class, Outer.Inner.class)
+                .get(Outer.Inner.class)
+                .getFieldAccessesFromSelf();
+
+        assertThatAccesses(fieldAccesses).contain(
+                access()
+                        .fromOrigin(Outer.Inner.class, "stringConcat")
+                        .toTarget(Outer.class, "outerPrivateString")
+                        .withAccessType(GET)).contain(
```
hankem, line 454: Just a cosmetic comment, but I'd find this easier to read if the two `.contain` calls were at the same level, e.g.:
```java
        assertThatAccesses(fieldAccesses)
                .contain(access()
                        .fromOrigin(Outer.Inner.class, "stringConcat")
                        .toTarget(Outer.class, "outerPrivateString")
                        .withAccessType(GET)
                )
                .contain(access()
                        .fromOrigin(Outer.Inner.class, "stringConcat")
                        .toTarget(Outer.class, "outerPrivateString")
                        .withAccessType(SET)
                );
```

hankem (approved):

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterSyntheticPrivateAccessesTest.java
codecholeric: Yikes, dunno how that happened 🙈 (the formatting wasn't intentional, maybe fixing a merge conflict or something...)

PR #1223, opened 2024-01-02
support optional `component` keyword in PlantUML diagram rules

tfij:
It is valid PlantUML syntax to prepend the `component` keyword to a component
declaration, e.g.

```
component [MyComponent] <<..mycmp..>>
```

Since it doesn't create a big overhead to support this as an optional part
of the component declaration, we can add it to make it easier for users
that distinguish their components from other types of objects like databases.


tfij:
PR should be rebased after the https://github.com/TNG/ArchUnit/pull/1184 merge so it will contain only a single commit

codecholeric (approved):
Looks good, thanks a lot for your contribution 😃

Issue #1185, opened 2023-11-03
OnionArchitecture: "withOptionalLayers" and "because" in the same statement casues that withOptionalLayers has no effect


pedrodelgado81:
Hi!
I wrote this statement

```
Architectures.onionArchitecture()
                .domainModels("com.myapp.domain.." )
		.applicationServices("com.myapp.application.port..","com.myapp.application.service..","com.myapp.application.usecase.." )
		.adapter( "adapters","com.myapp.adapters.." ) 
                .withOptionalLayers( true ).
		.because("SOME REASON").check( devClasses );
```

When I launched test, the output was: `Layer 'domain service' is empty`

If I remove **because( ... )** statement all works fine.

Dependency info

```	
 <dependency>
	<groupId>com.tngtech.archunit</groupId>
	<artifactId>archunit</artifactId>
	<version>1.1.0</version>
	<scope>test</scope>
</dependency>
```
**Update:** After debuggin I have seen that after call method  `public ArchRule because(String reason)` **optionalLayers** variable changes from true to false.

**Update 2:** The problem is `rule.as` (ArchRule line 133) call that launch the creation of a new Onion Architecture (@PublicAPI(usage = ACCESS) )  but doesn't initialize the flag optionalLayers. It takes the default value (false).

codecholeric:
Thanks for raising the issue and providing the analysis!! You're right, this must have been an oversight when the flag was added 🙈 I tried to fix it in https://github.com/TNG/ArchUnit/pull/1226 if you want to take a look...

pedrodelgado81:
Thanks for the work! ;)

PR #1226, opened 2024-01-07
fix `OnionArchitecture` losing `optionalLayers` after `as`

codecholeric:
The `OnionArchitecture.as(..)` method (and transitively `because(..)`) didn't pass the `optionalLayers` property along when creating a `new OnionArchitecture(..)` on return. Thus, once a test called `as(..)` or `because(..)` on the rule text `optionalLayers` would be reset to `false`.

Resolves: #1185

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/library/Architectures.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/Architectures.java
@@ -790,12 +790,14 @@ public final class Architectures {
                 Optional<DescribedPredicate<? super JavaClass>> domainServicePredicate,
                 Optional<DescribedPredicate<? super JavaClass>> applicationPredicate,
                 Map<String, DescribedPredicate<? super JavaClass>> adapterPredicates,
+                boolean optionalLayers,
                 List<IgnoredDependency> ignoredDependencies,
                 Optional<String> overriddenDescription) {
             this.domainModelPredicate = domainModelPredicate;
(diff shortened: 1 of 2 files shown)


archunit/src/main/java/com/tngtech/archunit/library/Architectures.java
```diff
@@ -1052,7 +1054,7 @@ public ArchRule allowEmptyShould(boolean allowEmptyShould) {
         @Override
         public OnionArchitecture as(String newDescription) {
             return new OnionArchitecture(domainModelPredicate, domainServicePredicate,
-                    applicationPredicate, adapterPredicates, ignoredDependencies,
+                    applicationPredicate, adapterPredicates, optionalLayers, ignoredDependencies,
```
hankem, line 1057: Unrelated to the issue reported in #1185, we also forgot to copy `allClassesAreContainedInArchitectureCheck`.

hankem (approved):
Thanks for fixing what I overlooked in #273!

codecholeric:
> Thanks for fixing what I overlooked in #273!

Well... 🤷‍♂️ "codecholeric approved these changes on Jan 9, 2020"

archunit/src/main/java/com/tngtech/archunit/library/Architectures.java
codecholeric: I just added a commit to fix this when we're already on it...

PR #1250, opened 2024-02-15
Add method toClassDependencies() to JavaAccess and to SliceDependency

dpolivaev:


archunit/src/main/java/com/tngtech/archunit/lang/syntax/elements/ClassesShouldConjunction.java
```diff
@@ -34,7 +34,7 @@
  * </code></pre>
  *
  * The semantics of the new condition will be {@code (public OR private) AND haveNameMatching}
- * – and not {@code public || (private && haveNameMatching)}.
+ * - and not {@code public || (private && haveNameMatching)}.
```
hankem, line 37: Would this work for you as well?
```diff
--- a/buildSrc/src/main/groovy/archunit.java-conventions.gradle
+++ b/buildSrc/src/main/groovy/archunit.java-conventions.gradle
@@ -15,4 +15,5 @@ tasks.withType(JavaCompile) { Task task ->
 
 javadoc {
     options.addBooleanOption('html5', true)
+    options.encoding = UTF_8.toString()
 }
```

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaAccess.java
```diff
@@ -107,6 +107,14 @@ String additionalToStringFields() {
         return "";
     }
 
+    /**
+     * @return all matching class dependencies.
+     */
+    @PublicAPI(usage = ACCESS)
+    public Set<Dependency> toClassDependencies() {
```
hankem, line 114: The name might be inspired by the previous usage of `Dependency.tryCreateFromAccess` in `JavaClassDependencies` (which is no public API). I wonder whether the name `getDependencies` might be more intuitive?

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaAccessTest.java
```diff
@@ -23,6 +26,9 @@ public void when_the_origin_is_an_inner_class_the_toplevel_class_is_displayed_as
                 .inLineNumber(7);
 
         assertThat(access.getDescription()).contains("(SomeClass.java:7)");
+        Dependency dependency = getOnlyElement(access.toClassDependencies());
+        assertThatType(dependency.getTargetClass()).as("target class").isEqualTo(access.getTargetOwner());
+        Assertions.assertThat(dependency.getDescription()).as("description").isEqualTo(access.getDescription());
```
hankem, line 31: I'd move the test for `access.toClassDependencies()` to a separate test, as it's not really related to `when_the_origin_is_an_inner_class_the_toplevel_class_is_displayed_as_location`.

archunit/src/main/java/com/tngtech/archunit/lang/syntax/elements/ClassesShouldConjunction.java
dpolivaev: Yes, it works too, I reverted my change and applied your suggestion.

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaAccessTest.java
dpolivaev: Done

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaAccess.java
dpolivaev: The method name was chosen as in the existing method com.tngtech.archunit.library.modules.ModuleDependency#toClassDependencies

To be honest, I would use getDependencies otherwise.

I have also considered adding an interface declaring this method. It might be called `SomeDependency` and implemented by JavaAccess, SliceDependency, ModuleDependency, and Dependency. I decided against it only because I was not sure about its name.

codecholeric:
As discussed I think this is superseded by #1251. If not, feel free to reopen / re-discuss 🙂

PR #1270, opened 2024-03-21
Handle $ at the end of class names for better Scala support

dpolivaev:
See https://stackoverflow.com/questions/41570148/why-does-scala-place-a-dollar-sign-at-the-end-of-class-names


schulzjo-tng:
We apologize for taking so long to take a look at this. Could you elaborate on what exact problem you encounter in Scala?

- Is this a functionality issue, i.e. is something broken in a Scala project and you need this change to fix it? If so, please provide some sort of example to reproduce the issue.
- Or is this merely a convenience issue (e.g. "the class is written without a `$` in Scala but the ArchUnit report shows a `$` symbol")? In that case, we would stick to what it says in the class file.

Issue #1237, opened 2024-01-23
JavaType#getAllInvolvedRawTypes() throws StackOverflowError with self-referential generic types


rashtao:
`JavaType#getAllInvolvedRawTypes()` throws `StackOverflowError` when invoked on a self-referential generic type.

The problem can be reproduced with the following code snippet:

```java
import com.tngtech.archunit.base.DescribedPredicate;
import com.tngtech.archunit.core.domain.JavaMethod;
import com.tngtech.archunit.core.importer.ClassFileImporter;

import static com.tngtech.archunit.lang.conditions.ArchConditions.be;
import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.methods;

public abstract class Foo<T extends Foo<T>> {
    public abstract T getThis();

    public static void main(String[] args) {
        methods()
                .should(be(new DescribedPredicate<JavaMethod>("allInvolvedRawTypes") {
                    @Override
                    public boolean test(JavaMethod m) {
                        return m.getReturnType().getAllInvolvedRawTypes().stream().allMatch(alwaysTrue());
                    }
                }))
                .check(new ClassFileImporter().importClasses(Foo.class));
    }
}
```

codecholeric:
Thanks for reporting! You're right that this is a bug 🙈 I'm gonna try to fix it as soon as I can find some free time to focus (spontaneously I'm not sure without introducing unwanted public API...)

codecholeric:
(if anybody else has more time available feel free to grab this)

PR #1276, opened 2024-03-24
fix `getAllInvolvedRawTypes()` recursing infinitely

codecholeric:
This fixes the problem of recursing infinitely when calling `JavaType.getAllInvolvedRawTypes()`
for a recursive type parameter declaration (e.g. `class Example<T extends Example<T>>`).
It also adds a new public API to conveniently traverse `JavaType` signatures
without the need to add chains of `instanceof` checks and manually handle how to traverse further.

Resolves: #1237

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaType.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaType.java
@@ -16,13 +16,22 @@
 package com.tngtech.archunit.core.domain;
 
 import java.lang.reflect.Type;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.Set;
+import java.util.function.Function;
+import java.util.function.Supplier;
 
+import com.google.common.collect.Iterables;
 import com.tngtech.archunit.PublicAPI;
 import com.tngtech.archunit.base.ChainableFunction;
 import com.tngtech.archunit.core.domain.properties.HasName;
 
 import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
+import static com.tngtech.archunit.PublicAPI.Usage.INHERITANCE;
+import static com.tngtech.archunit.core.domain.JavaType.SignatureVisitor.Result.CONTINUE;
+import static com.tngtech.archunit.core.domain.JavaType.SignatureVisitor.Result.STOP;
+import static java.util.Collections.singleton;
 
 /**
  * Represents a general Java type. This can e.g. be a class like {@code java.lang.String}, a parameterized type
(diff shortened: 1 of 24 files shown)


archunit/src/main/java/com/tngtech/archunit/core/domain/JavaType.java
```diff
@@ -82,7 +92,113 @@ public interface JavaType extends HasName {
      * @return All raw types involved in this {@link JavaType}
      */
     @PublicAPI(usage = ACCESS)
-    Set<JavaClass> getAllInvolvedRawTypes();
+    default Set<JavaClass> getAllInvolvedRawTypes() {
```
hankem, line 95: The body of the message of the commit ``fix `getAllInvolvedRawTypes()` recursing infinitely`` contains the following line, which probably shouldn't be there:
> This reverts commit 6f23f66fd7f355fe40bea5b5a0fcd18a17e2e783.

(At least I don't know what 6f23f66fd7f355fe40bea5b5a0fcd18a17e2e783 should be. 😉)

hankem (approved):
LGTM! 👍

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaType.java
```diff
@@ -82,7 +92,113 @@ public interface JavaType extends HasName {
...
+     * @see #traverseSignature(SignatureVisitor)
+     */
+    @PublicAPI(usage = INHERITANCE)
+    interface SignatureVisitor {
+        default Result visitClass(JavaClass type) {
+            return CONTINUE;
+        }
+
+        default Result visitArrayType(JavaClass type) {
+            return CONTINUE;
```
hankem, line 164: I wonder whether the default implementation should be
```suggestion
            return visitClass(type);
```

docs/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM jekyll/jekyll:stable
```
hankem, line 1: Does the local setup still work for you?

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaType.java
codecholeric: Mhmm, good catch, probably from rebasing and juggling commits 🙈

docs/Dockerfile
codecholeric: No, but I also recently switched to Mac 🤔 But changing the platform I still receive some strange errors that I think are just some version upgrade problems 🙈 I tried fixing it for 20 mins without success, but I think I'm gonna postpone this for now...

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaTypeTraversalTest.java
```diff
@@ -0,0 +1,386 @@
...
+        return new JavaType.SignatureVisitor() {
+            @Override
+            public Result visitClass(JavaClass type) {
+                traversedTypes.add(type);
+                return JavaType.SignatureVisitor.super.visitClass(type);
+            }
+
+            @Override
+            public Result visitArrayType(JavaClass type) {
+                traversedTypes.add(type);
```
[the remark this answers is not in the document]
codecholeric: Yeah, you're right. I had it differently in the beginning (which would have caught this), but then refactored it last minute to save some LOC 😁 Reverted the original version, can you re-check?

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaType.java
```diff
@@ -101,3 +217,86 @@ public JavaClass apply(JavaType input) {
...
+            // if we've encountered this type already we continue traversing the siblings,
+            // but we won't descend further into this type signature
+            return setLast(CONTINUE);
+        }
+        visited.add(type);
+        if (visitCurrent.apply(type) == CONTINUE) {
+            Result result = visit(nextTypes.get());
+            return setLast(result);
+        } else {
+            return setLast(STOP);
```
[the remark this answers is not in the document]
codecholeric: It does now by the change above, but there is still 
```
if (lastResult == STOP) {
  return STOP;
}
```
missing => now also added in 75a50b5ea768072e521f907bc8481fcc7ce57db2

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaType.java
codecholeric: Hmm, but that would seem strange to me. Because the contract should be that `visitClass` is never called with an array type, right? (Otherwise, why even have `visitClass` and `visitArrayType` separately in the visitor? If you can't rely on `isArray()` to return false whenever `visitClass` is called?)
The traversal behavior will just traverse down to the component type by default, right? So implementing `CONTINUE` will finally call `visitClass` with the base component type, and that seems a reasonable behavior if I e.g. only implement `visitClass` but not `visitArrayType` 🤷

hankem: > Because the contract should be that visitClass is never called with an array type, right?

Let's document the contract. 😉

I was checking whether we could make more use of the new `SignatureVisitor` internally, and found `JavaClassDependencies.dependenciesOfType(JavaType)`. `SignatureVisitor` won't be a nice solution there, but I realized that there might be more use cases where one previously had
```
        if (javaType instanceof JavaClass) {
            // do something with `(JavaClass) javaType`
        } else if (javaType instanceof JavaParameterizedType) {
            // do something with `(JavaParameterizedType) javaType`
        } else if (javaType instanceof JavaWildcardType) {
            // do something with `(JavaWildcardType) javaType`
        }
```
and it might not immediately obvious that refactoring this to the `SignatureVisitor` pattern requires overwriting both `visitClass` and `visitArrayType` for the ``do something with `(JavaClass) javaType`﻿`` part.

So yes, the question is actually whether `SignatureVisitor` should have individual methods for non-arrays and array `JavaClass`es. Having only one might be more consistent (why just distinguish arrays, and not, e.g. interfaces?) and therefore less confusing.

codecholeric: Hmm, I see your point, I just figured it would be kind of nice that the default behavior would traverse up the component types and automatically only give you the element type (as I found the JLS calls our "base component type" 🤪). But maybe that in the end creates a more confusing API instead of helping 🤔

codecholeric: One argument for the separate `visitArrayType` method might be that it's somewhat consistent with `visitGenericArrayType` 🤔 But I guess there the class hierarchy really is also distinct 🤷 
Guess I'm gonna remove the array type method then...

PR #1277, opened 2024-03-26
fix `@ArchIgnore` not being fully respected by JUnit 4 support

codecholeric:
In the case of inheritance we were looking at the wrong class for an `@ArchIgnore`. I.e. rule declarations in base classes would only be ignored if the base class was annotated with `@ArchIgnore`, not the actual class. We now check the actual class for the annotation instead. If the base class is annotated with `@ArchIgnore` we ignore it for now, since this is a corner case and one could argue for both, either ignoring all subclasses or not ignoring subclasses.


archunit-junit/junit4/src/test/java/com/tngtech/archunit/junit/internal/ArchUnitRunnerRunsMethodsTest.java
```diff
@@ -119,6 +121,18 @@ public void ignores_all_methods_in_classes_annotated_with_ArchIgnore() throws In
                 .contains(toBeIgnoredTwo);
     }
 
+    @Test
+    public void ignores_all_methods_in_base_classes_of_classes_annotated_with_ArchIgnore() throws InitializationError {
+        ArchUnitRunnerInternal runner = new ArchUnitRunnerInternal(IgnoredArchTestWithBaseClass.class);
```
hankem, line 126: We could consistently use
```suggestion
        ArchUnitRunnerInternal runner = newRunnerFor(IgnoredArchTestWithBaseClass.class);
```
(eliminating the need to declare the `InitializationError`)

hankem (approved):

PR #1278, opened 2024-03-26
fix issues with build dependencies and test execution

codecholeric:
I noticed that some parts of the build are broken:
* the source set `jdk9main` doesn't declare its dependency on the compiled classes of `main` which randomly leads to build fails
* the test execution of tests with the JUnit Jupiter engine is actually not running during regular CI anymore
This fixes both these issues and removes some obsolete configuration from the build.


hankem (approved):
LGTM! 👍

PR #1280, opened 2024-03-28
support executing single `@ArchTest` members

codecholeric:
Currently there is no convenient way to execute a single `@ArchTest` rule of a test running with JUnit 4 or 5 support.
As a simple mitigation we now provide the possibility to filter a single test via `archunit.properties` / system property.
This change allows to e.g. pass
```
-Darchunit.junit.testFilter=rule_field_one,rule_field_two
```
to a test run, which will then filter the tests and only run the `@Archtest` fields or methods with the given name(s).

For now we don't support any wildcards or narrowing the match down by surrounding class name
to see if this feature is even missed by users. Basic filtering for classes or packages should already be supported
by the respective test launcher (i.e. both JUnit 4 and 5 support should respect filter requests for classes or packages already).

Issue: #641


archunit-junit/junit5/engine/src/test/java/com/tngtech/archunit/junit/internal/ArchUnitTestEngineTest.java
```diff
@@ -725,6 +726,42 @@ void filtering_included_packages() {
...
+            assertThat(leafIds).isNotEmpty();
+            leafIds.forEach(leafId -> {
+                assertThat(leafId.getLastSegment().getType()).isEqualTo(FIELD_SEGMENT_TYPE);
+                assertThat(leafId.getLastSegment().getValue()).isEqualTo(SimpleRules.SIMPLE_RULE_FIELD_TWO_NAME);
+            });
+        }
+
+        @Test
+        void filtering_specific_rules_by_system_property() {
+            System.setProperty("archunit.junit.testFilter", "simple_rule_method_two,simple_rule");
```
hankem, line 749: We could add another filter for a rule that does not exist in order to show that this does not break anything:
```suggestion
            System.setProperty("archunit.junit.testFilter", "simple_rule_method_two,simple_rule,non-existing-rule");
```

docs/userguide/009_JUnit_Support.adoc
```diff
@@ -198,6 +198,25 @@ The runner will include all `@ArchTest` annotated members within `ServiceRules`
...
+It is possible to filter specific rules (e.g. `@ArchTest` fields) via `archunit.properties` (compare <<Advanced Configuration>>).
+
+[source,options="nowrap"]
+.archunit.properties
+----
+# Specify the field or method name here. Multiple names can be joined by ','
+junit.testFilter=my_custom_rule_field
+----
+
+As always with `archunit.properties`, this can also be passed dynamically using a system property.
```
hankem, line 212: I would join `e.g. passing` to this sentence:
```suggestion
As always with `archunit.properties`, this can also be passed dynamically using a system property,
```

archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/ArchUnitSystemPropertyTestFilterJUnit5.java
```diff
@@ -0,0 +1,61 @@
...
+import com.google.common.base.Splitter;
+import com.google.common.collect.ImmutableSet;
+import com.tngtech.archunit.ArchConfiguration;
+import org.junit.platform.engine.TestDescriptor;
+import org.junit.platform.engine.UniqueId;
+
+import static com.tngtech.archunit.junit.internal.ArchUnitTestDescriptor.FIELD_SEGMENT_TYPE;
+import static com.tngtech.archunit.junit.internal.ArchUnitTestDescriptor.METHOD_SEGMENT_TYPE;
+
+class ArchUnitSystemPropertyTestFilterJUnit5 {
```
hankem, line 31: Do you want to preserve the name `ArchUnitSystemPropertyTestFilter` in the first commit (``support filtering `@ArchTest` members by system property for JUnit 5``)?

hankem (approved):
LGTM! 👍

I noticed that this `junit.testFilter` does not allow us to distinguish `@ArchTest`s defined as fields and `@ArchTest`s defined as methods, which could have the same member name, but that's probably not needed in practice.

archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/ArchUnitSystemPropertyTestFilterJUnit5.java
codecholeric: Could do that, I guess 👍

codecholeric:
> I noticed that this junit.testFilter does not allow us to distinguish @ArchTests defined as fields and @ArchTests defined as methods, which could have the same member name, but that's probably not needed in practice.

Yes, I was aware of that, but as you wrote, I doubt that will really be a problem in practice. If it is, we can still ponder about making this more powerful. But, let's see first if there are really requirements for this...

Issue #1285, opened 2024-04-08
Jakarta Inject package name incorrect in BE_ANNOTATED_WITH_AN_INJECTION_ANNOTATION rule


bbadjari:
Jakarta's `Inject` annotation package name used for the `BE_ANNOTATED_WITH_AN_INJECTION_ANNOTATION` rule in the `GeneralCodingRules` class is set to `javax.inject.Inject` when it should be `jakarta.inject.Inject`. See the related [API documentation](https://jakarta.ee/specifications/dependency-injection/2.0/apidocs/jakarta/inject/inject).

hankem:
You're right: The rule had been added (#289) before [Jakarta EE9 was introduced](https://en.wikipedia.org/wiki/Jakarta_EE#History), and it was overlooked to update it to consider the `jakarta` package, too.

PR #1286, opened 2024-04-08
Consider Jakarta EE9 injection annotations for general coding rule

hankem:
The condition `BE_ANNOTATED_WITH_AN_INJECTION_ANNOTATION` and the rule `NO_CLASSES_SHOULD_USE_FIELD_INJECTION` previously considered the `javax.inject.Inject` and `javax.annotation.Resource` annotations, but not their jakarta variants (`jakarta.inject.Inject` and `jakarta.annotation.Resource`).

Resolves: #1285


codecholeric (approved):
lgtm, thanks for taking care of it!! 🙂

PR #1300, opened 2024-05-06
enable `@AnalyzeClasses` annotation to be used as meta annotation

mathze:
so far users are forced to repeat `@AnalyzeClasses` annotation an every test class. This cause additional maintenance overhead when common properties (e.g. package structure) changes. To support the DRY approach, `@AnalzyeClasses` annotation can now be used as meta annotation.

Resolves: #182

### DIFF
--- a/archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/ArchUnitTestDescriptor.java
+++ b/archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/ArchUnitTestDescriptor.java
@@ -39,12 +39,13 @@ import org.junit.platform.engine.support.descriptor.MethodSource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import static com.google.common.base.Preconditions.checkArgument;
 import static com.tngtech.archunit.junit.internal.DisplayNameResolver.determineDisplayName;
+import static com.tngtech.archunit.junit.internal.ReflectionUtils.findAnnotation;
 import static com.tngtech.archunit.junit.internal.ReflectionUtils.getAllFields;
 import static com.tngtech.archunit.junit.internal.ReflectionUtils.getAllMethods;
 import static com.tngtech.archunit.junit.internal.ReflectionUtils.getValueOrThrowException;
 import static com.tngtech.archunit.junit.internal.ReflectionUtils.invokeMethod;
+import static com.tngtech.archunit.junit.internal.ReflectionUtils.tryFindAnnotation;
 import static com.tngtech.archunit.junit.internal.ReflectionUtils.withAnnotation;
 
 class ArchUnitTestDescriptor extends AbstractArchUnitTestDescriptor implements CreatesChildren {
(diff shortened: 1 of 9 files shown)


archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/ArchUnitTestDescriptor.java
```diff
@@ -295,11 +300,14 @@ private static class JUnit5ClassAnalysisRequest implements ClassAnalysisRequest
         }
 
         private static AnalyzeClasses checkAnnotation(Class<?> testClass) {
-            AnalyzeClasses analyzeClasses = testClass.getAnnotation(AnalyzeClasses.class);
-            checkArgument(analyzeClasses != null,
+            List<AnalyzeClasses> analyzeClasses = new AnnotationFinder<>(AnalyzeClasses.class).findAnnotationsOn(testClass);
```
mathze, line 303: Not sure how to reach this from test.

codecholeric: This only gets called when you execute the discovery request. So, all the execute unit tests of `ArchUnitTestEngine` cover this, you can just destroy the place and run the test 😉

mathze: Not quite sure if I understand it correctly and it has been a while.
But after having a closer look, it seems that this method can only be called after `resolve` once run through `ifRequestedButUnresolved` with `createTestDescriptor`.
If this is true, the whole error-handling in this method can be removed, can't it?

codecholeric:
Thanks a lot for your support! 😃 The PR looked really nice to me 👍 I still just made some changes, but that has nothing to do with the quality. In particular:

* I adjusted the style in some places to be closer to existing code
* I integrated `AnnotationFinder` into `ReflectionUtils` - I have mixed feelings about this, because on the one hand it's nice if not too much is mingled into `ReflectionUtils`. But on the other hand, it felt a little bit unmotivated why so many actions related to Reflection are in there, just this specific annotation handling is outside. We might need to split `ReflectionUtils` up at some point if it grows too much, but so far it feels like it's still alright to me
* Biggest change in logic I did was removing the exception if multiple `@AnalyzeClasses` are in the hierarchy. While I really like the approach to be defensive and potentially poking users into a mistake they made, there is one reason that brought me to think we should just look for the "first" one and use that no matter what: for JUnit 5 we could just use the [AnnotationSupport](https://junit.org/junit5/docs/5.11.2/api/org.junit.platform.commons/org/junit/platform/commons/support/AnnotationSupport.html#findAnnotation(java.lang.reflect.AnnotatedElement,java.lang.Class)) provided by JUnit Platform and we wouldn't have to maintain any code. As far as we maintain JUnit 4, I guess it's fair to support this feature also there. But, if we one day drop JUnit 4 support, because the user base got really small, then having a stricter behavior than JUnit Platform's own `AnnotationSupport` would prevent us from using that. Furthermore, I think in most cases this shouldn't be a problem, I can't imagine users going crazy on using the meta-annotation support in whatever nested way, so I think this is more a theoretical confusion. If we really get issues about users being confused, because they shot themselves in the foot with confusing meta-annotation combinations in the same hierarchy, then we can still ponder if we need to do something more. But I assume this just won't happen.
* I added a section in the user guide 🙂 I think the feature is very nice, so we should make sure users can easily find out about it

In any case, I just went ahead and changed this, I really hope you don't mind! It's mainly because I have a hard time these days to make time to work on ArchUnit (as you might notice by the delay of this review 🙈), so having a lot of round-trips on the review would likely slow down the merge considerably. If there is something really wrong with the final state, we can still fix it in a follow-up.

archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/ArchUnitTestDescriptor.java
codecholeric: This place is a little simpler now anyway, because I moved some more logic into the annotation finding part 🙂 I.e. just made it more tailored to the users by offering a method that returns just the single annotation we expect and fails otherwise.

codecholeric: But it's possible that this method was used from one of the places only to retrieve the single annotation, but the preconditions couldn't fail anymore, because they were actually always already checked at that point.

codecholeric (approved):

mathze:
Hi @codecholeric,

don't worry, I'm glad that it made it into code :) and I could support. I know maintaining OSS isn't an easy job especially if done in sparetime.
Thank you very much for your effort ❤️.

Issue #768, opened 2022-01-08
Feature Request: Support for analysis of local variables


MahatmaFatalError:
During https://stackoverflow.com/questions/70507647/how-to-assert-hashmap-keys-to-be-comparable-with-archunit/70509141 we needed a way to analyze local variables.

Hence, I propose to extend JavaCodeUnit https://www.archunit.org/userguide/html/000_Index.html#_domain to also support local variable declaration.

codecholeric:
Technically this should be doable :slightly_smiling_face: But notice that there once already was #227 which pretty much suggested the same I think? However, I didn't get around to it back then and nobody else was willing to tackle it either, so I eventually closed it :disappointed: 
I would be still open to it (would be necessary to analyse performance impact though and possibly make it an opt-in feature if the drain is too big, since the use case is quite specific). But, as back then I personally don't know when I can really tackle this. So any support would be highly appreciated!

dbraun1991:
For more persons to find their way to this thread ...
https://stackoverflow.com/questions/74461422/archunit-test-sensitive-variable-naming
... ;-)

ttho:
Hello, is there still nobody working on this issue? I am willing to do so.

ttho:
Ok, at least someone likes this ;) 
I actually started on it, in case someone wants to join.

ttho:
@hankem , you hearted my intent to work on this. I created #1334 two weeks ago, but nobody took notice. Is there something wrong with it? Quick feedback welcome, even if it just be "need more time". Thanks.

hankem:
I'm very sorry, @ttho, that you didn't get any feedback so far. It's a sad reality that there we currently don't find much time. I'll try to take a look soon™.

olenz:
#1137

PR #1334, opened 2024-08-01
analyze local variable instantiations

ttho:
So far local variable instantiations were not analyzed at all, leaving gaps in the evaluation of dependency rules. If a local variable of a certain type was instantiated in a method, but no other use of that type was present (e.g. a method call on the type, or a field access), the dependency from the method to that type was undetected. Here, if enabled by a new configuration property (to preserve backward compatibility and performance), local variable instantiations add class dependencies from the surrounding method to the type of the variable and, if that type has generic type parameters, also from the method to any concrete generic type.

Issue: #768


olenz:
What has happened to this PR? What is needed for it to be integrated? We are employing the patch successfully since a few months, what is still missing?

hankem:
Thanks for the reminder! I'm sorry; the honest answer is that it was overlooked. I'll try to take a look at it _soon_.

PR #1361, opened 2024-10-07
fix fail test run with enabled junit.testFilter option

Sparkymann:
Hi! There is a problem if you enable the `junit.testFilter` setting, and there are other tests in the project besides arch-unit
Exception:
```
Caused by: org.junit.platform.commons.PreconditionViolationException: cannot remove the root of a hierarchy
```

In PR fix a condition so that non-ArchUnit descriptors are not checked
Tested the changes in my test repository https://github.com/Sparkymann/archunit-test
Failed workflow: https://github.com/Sparkymann/archunit-test/pull/1/checks
Corrected: https://github.com/Sparkymann/archunit-test/pull/3/checks

### DIFF
--- a/archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/ArchUnitSystemPropertyTestFilterJUnit5.java
+++ b/archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/ArchUnitSystemPropertyTestFilterJUnit5.java
@@ -48,7 +48,7 @@ class ArchUnitSystemPropertyTestFilterJUnit5 {
         ImmutableSet.copyOf(descriptor.getChildren())
                 .forEach(child -> removeNonMatching(child, shouldRunPredicate));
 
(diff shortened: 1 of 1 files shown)


nicholasgribanov (approved):

codecholeric:
Thanks a lot for taking care of this! I do have one suggestion that I added as a fixup (see the commit message for the reasoning). In a nutshell, I think we can solve it a little safer by checking `isRoot()` instead of relying on the ArchUnit internal class hierarchy.
Does this make sense to you? Then I'd squash and merge it...

codecholeric:
BTW: To get throught the CI commits also need a DCO sign-off in your commit message (see https://github.com/TNG/ArchUnit/blob/main/DCO#L34). If you're fine with that I could also add the line for you if you want 😉

Sparkymann:
Hi! I'm sorry for the long wait. 

> I do have one suggestion

Thank you, I agree with your suggestion, it looks like more general decision

> If you're fine with that I could also add the line for you if you want 

I would be very grateful to you.

andrzejsydor (approved):

Issue #1367, opened 2024-10-23
Missing use case for 'rule to handle more than one test class matching impl class'


krzysztof-owczarek:
Linked to: https://github.com/TNG/ArchUnit/pull/920

Structure like this:
```
com/tngtech/archunit/library/testclasses/packages/correct/onlyoneimplmatchingdir1/OnlyOneImplementationHaveTestAndItIsMatchingImplPackage.java
com/tngtech/archunit/library/testclasses/packages/correct/onlyoneimplmatchingdir1/OnlyOneImplementationHaveTestAndItIsMatchingImplPackageTest.java
com/tngtech/archunit/library/testclasses/packages/incorrect/onlyoneimplmatchingdir2/OnlyOneImplementationHaveTestAndItIsMatchingImplPackage.java
```

will fail test:
```
    @Test
    public void should_pass_when_only_one_of_two_implementations_have_test_class_and_it_is_in_implementation_package() {
        assertThatRule(testClassesShouldResideInTheSamePackageAsImplementation())
                .checking(new ClassFileImporter().importPackagesOf(
                        OnlyOneImplementationHaveTestAndItIsMatchingImplPackage.class,
                    com.tngtech.archunit.library.testclasses.packages.incorrect.onlyoneimplmatchingdir2.OnlyOneImplementationHaveTestAndItIsMatchingImplPackage.class
                ))
                .hasNoViolation();
    }
```

Right now it seems that if one of the implementation has a test in the correct package and the other one (with the same name) has no test at all, the `possibleTestClasses` will find the test in the wrong package and fail the rule. You have to add at least a blank test class in the right package for each implementation to make the rule pass.

Possible solution could be:
- keeping track of the implementation class list,
- checking if the `possibleTestClasses` have no matching implementation in the right package and discard it from checklist of another implementation class.

PR #1368, opened 2024-10-23
improve rule to not fail for classes with no tests in their packages

krzysztof-owczarek:
This will improve the existing rule to test that test classes reside in the same package as their implementation. In particular the rule:

Should not fail if there are multiple test classes with the same simple name and some of the classes have no tests at all

Issue: #1367

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/library/GeneralCodingRules.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/GeneralCodingRules.java
@@ -452,12 +452,16 @@ public final class GeneralCodingRules {
     private static ArchCondition<JavaClass> resideInTheSamePackageAsTheirTestClasses(String testClassSuffix) {
         return new ArchCondition<JavaClass>("reside in the same package as their test classes") {
             Map<String, List<JavaClass>> testClassesBySimpleClassName = new HashMap<>();
+            Map<String, List<JavaClass>> classesByPackageName = new HashMap<>();
 
             @Override
             public void init(Collection<JavaClass> allClasses) {
                 testClassesBySimpleClassName = allClasses.stream()
                         .filter(clazz -> clazz.getName().endsWith(testClassSuffix))
                         .collect(groupingBy(JavaClass::getSimpleName));
+
+                classesByPackageName = allClasses.stream()
+                        .collect(groupingBy(JavaClass::getPackageName));
             }
 
             @Override
@@ -468,7 +472,8 @@ public final class GeneralCodingRules {
                 List<JavaClass> possibleTestClasses = testClassesBySimpleClassName.getOrDefault(possibleTestClassName, emptyList());
 
                 boolean isTestClassInWrongPackage = !possibleTestClasses.isEmpty()
-                        && possibleTestClasses.stream().noneMatch(clazz -> clazz.getPackageName().equals(implementationClassPackageName));
+                        && possibleTestClasses.stream().noneMatch(clazz -> clazz.getPackageName().equals(implementationClassPackageName))
+                        && !allPossibleTestClassesHaveImplementationInRightPackage(possibleTestClasses);
 
                 if (isTestClassInWrongPackage) {
                     possibleTestClasses.forEach(wrongTestClass -> {
@@ -477,6 +482,13 @@ public final class GeneralCodingRules {
                         events.add(violated(wrongTestClass, message));
                     });
                 }
+            }
+
+            private boolean allPossibleTestClassesHaveImplementationInRightPackage(List<JavaClass> possibleTestClasses) {
+                return possibleTestClasses.stream()
+                        .allMatch(testClazz -> classesByPackageName.getOrDefault(testClazz.getPackageName(), emptyList())
+                                .stream()
+                                 .anyMatch(clazz -> testClazz.getSimpleName().equals(clazz.getSimpleName() + testClassSuffix)));
             }
         };
     }
--- a/archunit/src/test/java/com/tngtech/archunit/library/GeneralCodingRulesTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/library/GeneralCodingRulesTest.java
@@ -5,6 +5,7 @@ import com.tngtech.archunit.library.testclasses.packages.correct.customsuffix.Im
 import com.tngtech.archunit.library.testclasses.packages.correct.defaultsuffix.ImplementationClassWithCorrectPackage;
 import com.tngtech.archunit.library.testclasses.packages.correct.notest.ImplementationClassWithoutTestClass;
 import com.tngtech.archunit.library.testclasses.packages.correct.onedirmatching.ImplementationClassWithOneTestPackageMatchingOutOfTwo;
+import com.tngtech.archunit.library.testclasses.packages.correct.twoimplementationsonetestdir1.SimpleNameThatOccursInSeveralPackages;
 import com.tngtech.archunit.library.testclasses.packages.incorrect.nodirmatching.ImplementationClassWithMultipleTestsNotMatchingImplementationClassPackage;
 import com.tngtech.archunit.library.testclasses.packages.incorrect.wrongsubdir.customsuffix.ImplementationClassWithWrongTestClassPackageCustomSuffix;
 import com.tngtech.archunit.library.testclasses.packages.incorrect.wrongsubdir.customsuffix.subdir.ImplementationClassWithWrongTestClassPackageCustomSuffixTestingScenario;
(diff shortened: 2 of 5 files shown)


archunit/src/test/java/com/tngtech/archunit/library/testclasses/packages/correct/twoimplementationsonetestdir1/OnlyOneImplementationHaveTestAndItIsMatchingImplPackageTest.java
```diff
@@ -0,0 +1,4 @@
+package com.tngtech.archunit.library.testclasses.packages.correct.twoimplementationsonetestdir1;
+
+public class OnlyOneImplementationHaveTestAndItIsMatchingImplPackageTest {
```
hankem, line 3: For your test to cover the use case of #1367, this should be
```suggestion
class MultipleImplementationsWithOnlyOneTestAndInRightPackageTest {
```
to make your new test actually fail with the old implementation. 🙂

In addition: I'd drop the `AndInRightPackage` part from this name: It's a relevant aspect to have classes with the same simple name in different packages, so: _Is it right or is it wrong??_ 
How about `SimpleNameThatOccursInSeveralPackages`?

hankem (commented):
Thanks a lot for your contribution! 💚
I'd like to take a closer look whether we can simplify the rule, but I already wanted to leave one comment on the test. 😉

archunit/src/test/java/com/tngtech/archunit/library/testclasses/packages/correct/twoimplementationsonetestdir1/OnlyOneImplementationHaveTestAndItIsMatchingImplPackageTest.java
krzysztof-owczarek: I was reneming classes at the end and forgot about that one 🤦 Thanks for this suggestion - applied!

krzysztof-owczarek:
> Thanks a lot for your contribution! 💚 I'd like to take a closer look whether we can simplify the rule, but I already wanted to leave one comment on the test. 😉

I think it could be done in a simpler way, sorry for not exploring that further before creating that PR, but I am just before my vacation and had little time to create it.

I am not sure you support test classes that are not suffixed by `Test` - if not, there may be a possibility to achieve same thing like this:

<DRAFT>

```
Map<String, List<JavaClass>> testClassesBySimpleClassName = new HashMap<>();
            Map<String, JavaClass> classesByPackageName = new HashMap<>();

            @Override
            public void init(Collection<JavaClass> allClasses) {
                testClassesBySimpleClassName = allClasses.stream()
                        .filter(clazz -> clazz.getName().endsWith(testClassSuffix))
                        .collect(groupingBy(JavaClass::getSimpleName));

                classesByPackageName = allClasses.stream()
                        .collect(toMap(JavaClass::getFullName, Function.identity()));
            }
```

and then some draft of the test:
```
possibleTestClasses.stream()
   .map(testClazz -> classesByPackageName.getOrDefault(createImplClassNameFromTestClassName(testClazz), null))
   .noneMatch(Objects::nonNull);
```
but it requires a sloppy name construction like this:
```
 private String createImplClassNameFromTestClassName(JavaClass testClazz) {
     return testClazz.getPackageName() + "." + testClazz.getSimpleName().replace("Test", "");
 }
```

I don't like it, I am not proud of it, but maybe it will be a step towords a simpler solution ;)

krzysztof-owczarek:
> Thanks a lot for your contribution! 💚 I'd like to take a closer look whether we can simplify the rule, but I already wanted to leave one comment on the test. 😉

I have a feeling it might be done in with a simpler check. Sorry for not persuing that before I have created this PR, but I am just before my vacation and had a little time to create the change 🌴 

I have another, simpler draft of such a condition, but it requires a bit sloppy name creation function:

```
            Map<String, List<JavaClass>> testClassesBySimpleClassName = new HashMap<>();
            Map<String, JavaClass> classesByPackageName = new HashMap<>();

          @Override
          public void init(Collection<JavaClass> allClasses) {
              testClassesBySimpleClassName = allClasses.stream()
                      .filter(clazz -> clazz.getName().endsWith(testClassSuffix))
                      .collect(groupingBy(JavaClass::getSimpleName));

              classesByPackageName = allClasses.stream()
                      .collect(toMap(JavaClass::getFullName, Function.identity()));
          }
```
and then:
```
possibleTestClasses.stream()
                        .map(testClazz -> classesByPackageName.getOrDefault(
                                createImplClassNameFromTestClassName(testClazz), null))
                        .noneMatch(Objects::nonNull);
```
having:
```
private String createImplClassNameFromTestClassName(JavaClass testClazz) {
                return testClazz.getPackageName() + "." + testClazz.getSimpleName().replace(testClassSuffix, "");
            }
```

I am not super happy with this draft, but the condition is much simpler that way. Maybe it will be a step towords some better solution.

krzysztof-owczarek:
Hey @hankem, I will paste the part of annonynous class here, that has a simpler condition, but is based on name resolving as described before. Give me heads up what do you think about it.

```
private static ArchCondition<JavaClass> resideInTheSamePackageAsTheirTestClasses(String testClassSuffix) {
        return new ArchCondition<JavaClass>("reside in the same package as their test classes") {
            Map<String, List<JavaClass>> testClassesBySimpleClassName = new HashMap<>();
            Map<String, JavaClass> classesByPackageName = new HashMap<>();

            @Override
            public void init(Collection<JavaClass> allClasses) {
                testClassesBySimpleClassName = allClasses.stream()
                        .filter(clazz -> clazz.getName().endsWith(testClassSuffix))
                        .collect(groupingBy(JavaClass::getSimpleName));

                classesByPackageName = allClasses.stream()
                        .collect(toMap(JavaClass::getFullName, Function.identity()));
            }

            @Override
            public void check(JavaClass implementationClass,
                              ConditionEvents events) {
                String implementationClassName = implementationClass.getSimpleName();
                String implementationClassPackageName = implementationClass.getPackageName();
                String possibleTestClassName = implementationClassName + testClassSuffix;
                List<JavaClass> possibleTestClasses = testClassesBySimpleClassName.getOrDefault(possibleTestClassName,
                        emptyList());

                boolean isTestClassInWrongPackage = !possibleTestClasses.isEmpty()
                        && possibleTestClasses.stream().noneMatch(
                        clazz -> clazz.getPackageName().equals(implementationClassPackageName))
                        && implementationNotFoundInTestClassPackages(possibleTestClasses);

                if (isTestClassInWrongPackage) {
                    possibleTestClasses.forEach(wrongTestClass -> {
                        String message = createMessage(wrongTestClass,
                                String.format("does not reside in same package as implementation class <%s>",
                                        implementationClass.getName()));
                        events.add(violated(wrongTestClass, message));
                    });
                }
            }

            private boolean implementationNotFoundInTestClassPackages(List<JavaClass> possibleTestClasses) {
                return possibleTestClasses.stream()
                        .map(testClazz -> classesByPackageName.getOrDefault(
                                resolveImplClassNameFromTestClassName(testClazz), null)
                        )
                        .noneMatch(Objects::nonNull);
            }

            private String resolveImplClassNameFromTestClassName(JavaClass testClazz) {
                return testClazz.getPackageName() + "." + testClazz.getSimpleName().replace(testClassSuffix, "");
            }

        };
    }
```

archunit/src/main/java/com/tngtech/archunit/library/GeneralCodingRules.java
```diff
@@ -478,6 +483,14 @@ public void check(JavaClass implementationClass, ConditionEvents events) {
                     });
                 }
             }
+
+            private boolean allPossibleTestClassesHaveImplementationInRightPackage(List<JavaClass> possibleTestClasses) {
+                return possibleTestClasses.stream()
+                        .allMatch(testClazz -> classesByPackageName.getOrDefault(testClazz.getPackageName(), emptyList())
+                                .stream()
+                                .filter(clazz -> !testClazz.getSimpleName().equals(clazz.getSimpleName()))
+                                .anyMatch(clazz -> testClazz.getSimpleName().startsWith(clazz.getSimpleName())));
```
hankem, line 492: We could check the exact expected test name (to avoid false negatives when the test name coincidentally starts with the name of another unrelated class): 

```suggestion
                                 .anyMatch(clazz -> testClazz.getSimpleName().equals(clazz.getSimpleName() + testClassSuffix)));
```

(which requires passing `testClassSuffix` as another method parameter)

hankem (commented):
I'm really sorry that it took me so long to get back to this! 🙄
But now that a new release is coming up, I'd really like to include your contribution...

I have one suggestion to further improve the rule. If you like it, we could squash all changes into one commit.

krzysztof-owczarek:
Hi @hankem 👋 Looks good, thanks. I applied your suggestion :)

hankem (approved):

PR #1405, opened 2025-01-18
raise error for empty violation store and delete empty rule file

### LINKED ISSUE #1264: Feature suggestion: Freeze files and stored.rules validation / sanity check
Hi ArchUnit Team,

First of you do a great job here. ArchUnit is great.

In our project we currently have >70 freeze files in the stored.rules mentioned.

Digging deeper I noticed several things, which I am concerned about.

1. The files referenced might not exist anymore, but the reference in stored.rules is
2. When freeze files become empty you are not being advised to remove the freezing from your test. 
3. Freeze files might still be in the archunit folder, but not referenced anymore
4. Tests were removed without removing the stored.rules line or freeze file

Of course in the end it is our job to keep our source code in check and clean up correctly etc.

Yet I wonder how ArchUnit itself can help to have a sanity check of its freeze files and stored.rules?

What is your opinion on this?

maxxkia:
These changes introduce two new features for FreezingRule default store
* Raise error when a freezing rule has zero violations. This can be enabled by setting the property `default.warnEmptyRuleViolation=true`. For backward compatibility it is disabled by default.
* Skip rule violation file creation or delete if it already exists, when there are zero violations. This can be enabled by setting the property `default.deleteEmptyRuleViolation=true`, it is disabled by default.

Signed-off-by: Masoud Kiaeeha <6916434+maxxkia@users.noreply.github.com>

Resolves #1264

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/library/freeze/TextFileBasedViolationStore.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/freeze/TextFileBasedViolationStore.java
@@ -73,11 +73,17 @@ public final class TextFileBasedViolationStore implements ViolationStore {
     private static final String ALLOW_STORE_CREATION_DEFAULT = "false";
     private static final String ALLOW_STORE_UPDATE_PROPERTY_NAME = "default.allowStoreUpdate";
     private static final String ALLOW_STORE_UPDATE_DEFAULT = "true";
+    private static final String DELETE_EMPTY_RULE_VIOLATION_PROPERTY_NAME = "default.deleteEmptyRuleViolation";
+    private static final String DELETE_EMPTY_RULE_VIOLATION_DEFAULT = "false";
+    private static final String WARN_EMPTY_RULE_VIOLATION_PROPERTY_NAME = "default.warnEmptyRuleViolation";
+    private static final String WARN_EMPTY_RULE_VIOLATION_DEFAULT = "false";
 
     private final RuleViolationFileNameStrategy ruleViolationFileNameStrategy;
 
     private boolean storeCreationAllowed;
     private boolean storeUpdateAllowed;
+    private boolean deleteEmptyRule;
+    private boolean warnEmptyRuleViolation;
     private File storeFolder;
     private FileSyncedProperties storedRules;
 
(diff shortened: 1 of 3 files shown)


archunit/src/main/java/com/tngtech/archunit/library/freeze/TextFileBasedViolationStore.java
```diff
@@ -73,11 +73,17 @@ public final class TextFileBasedViolationStore implements ViolationStore {
     private static final String ALLOW_STORE_CREATION_DEFAULT = "false";
     private static final String ALLOW_STORE_UPDATE_PROPERTY_NAME = "default.allowStoreUpdate";
     private static final String ALLOW_STORE_UPDATE_DEFAULT = "true";
+    private static final String DELETE_EMPTY_RULE_VIOLATION_PROPERTY_NAME = "default.deleteEmptyRuleViolation";
+    private static final String DELETE_EMPTY_RULE_VIOLATION_DEFAULT = "false";
```
maxxkia, line 77: I'd like to discuss the above default value. I would prefer to have default `true` here, but I'd like to first hear the opinions from the maintainers.

archunit/src/main/java/com/tngtech/archunit/library/freeze/TextFileBasedViolationStore.java
```diff
@@ -140,15 +148,38 @@ public boolean contains(ArchRule rule) {
     @Override
     public void save(ArchRule rule, List<String> violations) {
         log.trace("Storing evaluated rule '{}' with {} violations: {}", rule.getDescription(), violations.size(), violations);
+        if (violations.isEmpty() && warnEmptyRuleViolation) {
```
maxxkia, line 151: this check has to be placed here at the top, because for empty frozen rule violations you don't necessarily need to update the store. By seeing such warning the developer could change the test to non-freezing rather than decide to update the store.

archunit/src/main/java/com/tngtech/archunit/library/freeze/TextFileBasedViolationStore.java
```diff
@@ -140,15 +148,38 @@ public boolean contains(ArchRule rule) {
...
+            throw new StoreEmptyException(String.format("Saving empty violations for freezing rule is disabled (enable by configuration %s.%s=true)",
+                    ViolationStoreFactory.FREEZE_STORE_PROPERTY_NAME, WARN_EMPTY_RULE_VIOLATION_PROPERTY_NAME));
+        }
+        if (violations.isEmpty() && deleteEmptyRule && !contains(rule)) {
```
maxxkia, line 155: another way to write this check would be 
```suggestion
        if (violations.isEmpty() && (deleteEmptyRule || skipEmptyRule) && !contains(rule)) {
```
where `skipEmptyRule` condition is only used for skipping creation of empty rule file. To keep the configuration simple, I decided to use `deleteEmptyRule` for both skipping file creation and deleting the empty file.

PR #1407, opened 2025-01-19
cleanup obsolete frozen rules

### LINKED ISSUE #1264: Feature suggestion: Freeze files and stored.rules validation / sanity check

maxxkia:
These changes bring 2 features
* clean up of frozen rules in `stored.rules` where the corresponding file does not exist in the store directory
* clean up of files in store directory which are not referenced in `stored.rules` file
Both of the above operations are enabled using `default.allowStoreUpdate` property. If `default.allowStoreUpdate=false` and obsolete entries or files are found by either of the above operations, the operation fails with an `StoreUpdateFailedException`.

Signed-off-by: Masoud Kiaeeha <6916434+maxxkia@users.noreply.github.com>

Resolves #1264

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/library/freeze/TextFileBasedViolationStore.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/freeze/TextFileBasedViolationStore.java
@@ -20,11 +20,16 @@ import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Properties;
+import java.util.Set;
 import java.util.UUID;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
+import com.google.common.base.Predicates;
 import com.google.common.base.Splitter;
 import com.tngtech.archunit.PublicAPI;
 import com.tngtech.archunit.lang.ArchRule;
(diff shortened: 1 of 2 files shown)


archunit/src/main/java/com/tngtech/archunit/library/freeze/TextFileBasedViolationStore.java
```diff
@@ -132,6 +139,45 @@ private void checkInitialization(boolean initializationSuccessful, String messag
...
+        Set<String> ruleFiles = storedRules.keySet().stream()
+                .map(storedRules::getProperty)
+                .collect(Collectors.toSet());
+
+        List<String> danglingFiles = Arrays.stream(storeFolder.list())
+                .filter(name -> !name.equals(STORED_RULES_FILE_NAME))
+                .filter(Predicates.not(ruleFiles::contains))
+                .collect(toList());
+
+        if (!danglingFiles.isEmpty() && !storeUpdateAllowed) {
```
maxxkia, line 164: for controlling the new features we can either reuse `storeUpdateAllowed` property (current implementation) or a new property. reusing  the existing property would make this change not backward compatible and is probably not favored. But first I'd like to hear feedback from the maintainers.

PR #1408, opened 2025-01-19
extend CI to run tests with JDK 21

codecholeric:
This was previously blocked by our Gradle version being too old. Now that we can run the tests with JDK 21 it shows though, that the `Annotation` `toString()` behavior has changed again. Unfortunately, this always either causes complexity on the production or test side. If we want to test the `AnnotationProxy.toString()` behavior, then either we have to add a workaround to the test code, or we have to adjust the production code to produce the equivalent `toString()` value. I opted for the latter in hopes that they will finally stop changing the `toString()` style every couple of versions.


hankem (approved):

build.gradle
```diff
@@ -1,6 +1,6 @@
 plugins {
     id 'archunit.base-conventions'
-    id 'com.github.johnrengelman.shadow' version '7.1.2' apply false
+    id 'com.gradleup.shadow' version '8.3.5' apply false
```
hankem, line 3: After applying the `fixup!`, the first commit's message shouldn't be
> upgrade Gradle Shadow plugin to version `8.1.1`

anymore. 😉

hankem (approved):

build.gradle
codecholeric: Huh, I thought I adjusted that 🤯 Weird...

codecholeric: Are you sure you reviewed the latest state? When I check the 3 commits it says "upgrade Gradle Shadow plugin to version `8.3.5`"

hankem: Ooh, sorry: I was indeed not... 🙈 That's a wonderful commit message!

PR #1418, opened 2025-02-04
Support class versions

### LINKED ISSUE #423: Rule to restrict to specific bytecode java version
Hi, I can't seem to determine how to accomplish my use case.  Any help would be greatly appreciated.  

Is it possible to verify that a jar I create is compiled to java 8 or below including all dependencies used by that jar? 

I don't see anything in the API to check for bytecode version but I might just be missing it.

Thank you,
Peter

jvandort:
Add the class version to JavaClass so rules can verify the class version

Fixes https://github.com/TNG/ArchUnit/issues/423

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClassVersion.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClassVersion.java
@@ -1,0 +1,61 @@
+package com.tngtech.archunit.core.domain;
+
+/**
+ * Represents the version of a JVM class file.
+ */
+public class JavaClassVersion {
+
+    private final int major;
+    private final int minor;
+
+    /**
+     * Create a new instance of {@link JavaClassVersion}.
+     *
+     * @param major The major version of the class file.
+     * @param minor The minor version of the class file.
+     */
+    public JavaClassVersion(int major, int minor) {
+        this.major = major;
+        this.minor = minor;
+    }
+
+    /**
+     * Get the Java version corresponding to the major version of the class file.
+     *
+     * @return The Java version
+     */
+    public int getJavaVersion() {
+        return major - 44;
+    }
+
+    /**
+     * Get the major version of the class file.
+     *
+     * @return The major version
+     */
+    public int getBytecodeMajorVersion() {
+        return major;
+    }
+
+    /**
+     * Get the minor version of the class file.
+     *
+     * @return The minor version
+     */
+    public int getBytecodeMinorVersion() {
+        return minor;
+    }
+
+    /**
+     * Create a new instance of {@link JavaClassVersion} from the version as provided by ASM.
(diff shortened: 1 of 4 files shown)


jvandort:
I followed the guidelines in the original issue, but was unsure about testing.

How would you advise testing this?

hankem:
One option could be to commit compiled classes in `src/test/resources` to the repository.

jvandort:
Do you have an example test method/class I should follow the style/pattern of?

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClassVersion.java
```diff
@@ -0,0 +1,61 @@
...
+        this.minor = minor;
+    }
+
+    /**
+     * Get the Java version corresponding to the major version of the class file.
+     *
+     * @return The Java version
+     */
+    public int getJavaVersion() {
+        return major - 44;
```
StefanGraeber, line 28: to be a bit nitpicky:
Do you consider the version old java versions 1.0, 1.1, ... as 1, 1, 2?
https://javaalmanac.io/bytecode/versions/ 
Java 1.0 -> class file version 45.0 -> java version 1
Java 1.1 -> class file version 45.3 -> java version 1
Java 1.2 -> class file version 46.0 -> java version 2
...
starting from Java5 you calculation works perfectly fine, before that the numbering is already weird on java side.
Do we ignore this since these version are so ancient that we can't expect anyone to stumble over java classes compiled before java 5?
Since even Java 8 is currently dyng, this should be fine

StefanGraeber:
A test would be really great for this, including the LTS versions starting from Java 8: 8, 11, 17, 21, 25
checking in compiled classes sounds as the most stable solution, but I'm not sure of how to add these classes to the scanned packages/classes

schulzjo-tng (changes requested):
This looks good! I think adding tests shouldn't be too hard, since we already have some Java-version-specific tests (see e.g. here):
https://github.com/TNG/ArchUnit/blob/6466ac31032b7f1770d4d31f4b67a7f20fe62bd4/archunit/build.gradle#L61-L63
Adding one test case for each of those should be good enough (e.g. Java 8, 9, 16, 25)

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClassVersion.java
schulzjo-tng: We don't officially support anything older than Java 8, so a minor inaccuracy in on something this old is fully acceptable to me. However, we may want to document where this magic value of `44` comes from, e.g. like this:
```suggestion
        // see e.g. https://javaalmanac.io/bytecode/versions/; Java 2 has class file version 46 etc.
        return major - 44;
```

Issue #1434, opened 2025-03-17
Is it possible to use fluent API to set priority for SlicesRuleDefinition? (Related to closed issue#129)


guesshe:
In https://github.com/TNG/ArchUnit/issues/129, it was mentioned that ```ArchRuleDefinition.priority(HIGH).classes().that()...``` can be used to set priority for ArchRule, however, the same fluent API is not available for SlicesRuleDefinition. 

Any reasons why this is not available for SlicesRuleDefinition?

hankem:
I think that slice rules are not available via `ArchRuleDefinition.priority(Priority)` because the latter is defined in the `lang` package, but the former live in the `library` package.

I guess there is no specific reason why `SlicesRuleDefinition.slices()` doesn't offer an API to set the priority (other than that it was overlooked; I'm not aware of much interest in these priorities so far... 🙈). Would you like to prioritize slice rules?

guesshe:
In my user case, I tried to map DDD to package level and use ArchRuleDefinition to restrict the interactions among bounded contexts in different domains. I would like to prioritize these rules since these are the cornerstones of the software architecture. I am also using SlicesRuleDefinition to prevent cyclic dependency, which IMO, should also be at the same priority level as domain boundaries restrictions. Would you like me to work on a PR for this? We can have further discussions on the details and if it doesn't worth the effort or a better solution already exists in current fluent API, the PR can be dropped.

guesshe:
@hankem Looking at SlicesRuleDefinition class. The priority is hardcoded. I am thinking of overloading matching() and assignedFromby() methods by providing another parameter to setup priority. Something like this 
```
matching(String packageIdentifier, Priority priority)
assignedFrom(SliceAssignment assignment, Priority priority)
```

```
@PublicAPI(
    usage = Usage.ACCESS
)
public final class SlicesRuleDefinition {
    private SlicesRuleDefinition() {
    }

    @PublicAPI(
        usage = Usage.ACCESS
    )
    public static Creator slices() {
        return new Creator();
    }

    @PublicAPI(
        usage = Usage.ACCESS
    )
    public static class Creator {
        private Creator() {
        }

        @PublicAPI(
            usage = Usage.ACCESS
        )
        public GivenSlices matching(String packageIdentifier) {
            return new GivenSlicesInternal(Priority.MEDIUM, Slices.matching(packageIdentifier));
        }

        @PublicAPI(
            usage = Usage.ACCESS
        )
[...]

PR #1436, opened 2025-03-20
allow explicit priority config in SlicesRuleDefinition

guesshe:
Resolves: #1434

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/library/dependencies/SlicesRuleDefinition.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/dependencies/SlicesRuleDefinition.java
@@ -64,7 +64,15 @@ public final class SlicesRuleDefinition {
          */
         @PublicAPI(usage = ACCESS)
         public GivenSlices matching(String packageIdentifier) {
-            return new GivenSlicesInternal(Priority.MEDIUM, Slices.matching(packageIdentifier));
+            return matching(packageIdentifier, Priority.MEDIUM);
+        }
+
+        /**
+         * @see Slices#matching(String)
+         */
+        @PublicAPI(usage = ACCESS)
+        public GivenSlices matching(String packageIdentifier, Priority priority) {
+            return new GivenSlicesInternal(priority, Slices.matching(packageIdentifier));
         }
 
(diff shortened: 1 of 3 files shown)


archunit/src/main/java/com/tngtech/archunit/library/dependencies/SlicesRuleDefinition.java
```diff
@@ -67,12 +67,28 @@ public GivenSlices matching(String packageIdentifier) {
             return new GivenSlicesInternal(Priority.MEDIUM, Slices.matching(packageIdentifier));
```
hankem, line 67: We could reduce a tiny bit of code duplication if this now delegated to the new method:
```suggestion
            return matching(packageIdentifier, Priority.MEDIUM);
```

archunit/src/main/java/com/tngtech/archunit/library/dependencies/SlicesRuleDefinition.java
```diff
@@ -67,12 +67,28 @@ public GivenSlices matching(String packageIdentifier) {
...
+        public GivenSlices matching(String packageIdentifier, Priority priority) {
+            return new GivenSlicesInternal(priority, Slices.matching(packageIdentifier));
+        }
+
         /**
          * @see Slices#assignedFrom(SliceAssignment)
          */
         @PublicAPI(usage = ACCESS)
         public GivenSlices assignedFrom(SliceAssignment assignment) {
             return new GivenSlicesInternal(Priority.MEDIUM, Slices.assignedFrom(assignment));
```
hankem, line 83: This could likewise delegate to the other new method:
```suggestion
            return assignedFrom(assignment, Priority.MEDIUM);
```

hankem (commented):
Thanks for the contribution!
I'll push a tiny fixup commit to simplify the previously existing code. If you agree, we can squash it into yours and merge the PR.

guesshe:
> Thanks for the contribution! I'll push a tiny fixup commit to simplify the previously existing code. If you agree, we can squash it into yours and merge the PR.

Sweet! Thanks for tidying it up!

PR #1443, opened 2025-04-06
Ignore synthetic bridge methods in ProxyRules

### LINKED ISSUE #1193: Recommended test for @Transactional does not work in combination with generic interface/super class
I use your recommended test to avoid internal calls to methods with `@Transactional`
```
no_classes_should_directly_call_other_methods_declared_in_the_same_class_that_are_annotated_with(
            Transactional.class);
```

Unfortunately this does not work if the method is an implementation of a generic interface method

```
public interface ItemProcessor<I, O> {
    Integer process(I var1) throws Exception;
}

public class LoadProcessor implements ItemProcessor<String, Integer> {

    @Override
    @Transactional
    public Integer process(String var1) throws Exception {
        return null;
    }
}

```

The result is
`java.lang.AssertionError: Architecture Violation [Priority: MEDIUM] - Rule 'no classes should directly call other methods declared in the same class that are annotated with @Transactional, because it bypasses the proxy mechanism' was violated (1 times):
Method <org.example.LoadProcessor.process(java.lang.Object)> calls method <org.example.LoadProcessor.process(java.lang.String)> in (LoadProcessor.java:5)`

Same result if ItemProcessor is an abstract class. Tested with `archunit-junit5` in version `1.2.0`.

hankem:
Synthetic bridge methods generated for methods overriding parameterized methods give false-positive violations of the [`com.tngtech.archunit.library.ProxyRules`](https://javadoc.io/doc/com.tngtech.archunit/archunit/latest/com/tngtech/archunit/library/ProxyRules.html) (introduced with #539).

With this PR, `ProxyRules` ignore calls from synthetic bridge methods (resolves #1193).

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/library/ProxyRules.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/ProxyRules.java
@@ -20,15 +20,18 @@ import java.lang.annotation.Annotation;
 import com.tngtech.archunit.PublicAPI;
 import com.tngtech.archunit.base.DescribedPredicate;
 import com.tngtech.archunit.core.domain.AccessTarget.MethodCallTarget;
+import com.tngtech.archunit.core.domain.JavaAccess.Functions.Get;
 import com.tngtech.archunit.core.domain.JavaClass;
-import com.tngtech.archunit.core.domain.JavaMethodCall;
 import com.tngtech.archunit.lang.ArchCondition;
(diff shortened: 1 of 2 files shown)


codecholeric:
I added a suggestion as 8a8179c3dd5c089771297718e3dd183e52253132, what do you think?

hankem:
Cool, thanks!

PR #1450, opened 2025-04-24
Update Maven examples (and Gradle wrapper)

hankem:
This PR extends the tasks triggered by `updateArchUnitExample` (in particular for every release), which update the [ArchUnit-Examples](https://github.com/TNG/ArchUnit-Examples), by
* updating the Maven examples (introduced with TNG/ArchUnit-Examples#14)
* updating the Gradle wrapper

It also updates the example dependencies and Maven version used in archunit-maven-test.


build-steps/release/archunit-examples-utils.gradle
```diff
@@ -4,17 +4,19 @@ ext.archunitExamplesGitRepo = 'TNG/ArchUnit-Examples.git'
 ext.updateArchUnitExampleVersion = { File archUnitExampleDir ->
```
hankem, line 4: can be tested with
```
./gradlew clean updateArchUnitExample && cd build/example-update/ && git diff
```
(noting that release.sh additionally does && git add -A && git commit -s -m "release ArchUnit $VERSION" && git push).

archunit-maven-test/pom.xml.template
```diff
@@ -70,15 +70,15 @@
...
+                    <version>3.0.0-M3</version> <!-- Since 3.0.0-M4, the test report for JUnit5 tests does not contain fully qualified names in testcase/@classname anymore. -->
                     <configuration>
                         <argLine>-Xmx2G</argLine>
                     </configuration>
                 </plugin>
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-compiler-plugin</artifactId>
-                    <version>3.8.1</version>
+                    <version>3.15.0</version>
```
schulzjo-tng: These collide with a change on #1659:

https://github.com/TNG/ArchUnit/pull/1659/changes#diff-37d07b0dd191f89abeecc0d77000a86578e68d32149148fbe71bf426ac4ce8f9L73-R81

I prefer my approach where this is covered by dependabot

hankem: I like your approach, and I don't think it's a semantic conflict, but that they complement each other nicely:

* This PR is to update https://github.com/TNG/ArchUnit-Examples based on the version in the ArchUnit repository
* Your PR allows the version in the ArchUnit repository to be updated automatically.

archunit-maven-test/.mvn/wrapper/maven-wrapper.properties
```diff
@@ -14,5 +14,6 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.8.6/apache-maven-3.8.6-bin.zip
-wrapperUrl=https://repo.maven.apache.org/maven2/org/apache/maven/wrapper/maven-wrapper/3.1.1/maven-wrapper-3.1.1.jar
+wrapperVersion=3.3.4
+distributionType=only-script
```
schulzjo-tng: if we are on `only-script` distribution, should we delete `maven-wrapper.jar`? https://maven.apache.org/tools/wrapper/maven-wrapper-plugin/wrapper-mojo.html#distributionType

Also, `distributionType=only-script` is the default, so we could omit it. Doesn't hurt to have it, though

build-steps/release/archunit-examples-utils.gradle
```diff
@@ -4,17 +4,19 @@ ext.archunitExamplesGitRepo = 'TNG/ArchUnit-Examples.git'
...
     fileTree(archUnitExampleDir) {
         include '**/build.gradle'
+        include '**/pom.xml'
     }.each { File buildFile ->
-        buildFile.text = buildFile.text.replaceAll(/(com\.tngtech\.archunit:archunit[^:]*:)[\w.-]*/, "\$1${version}")
+        buildFile.text = buildFile.text.replaceAll(/(com\.tngtech\.archunit(:|<\/groupId>\n\s*<artifactId>)archunit([^:]*:|[^<]*<\/artifactId>\n\s*<version>))[\w.-]*/, "\$1${version}")
```
schulzjo-tng: I can observe that the behaviour is correct, but this regex is very complex - partly because it needs to cover both Gradle and Maven. Do we want to pull apart the Gradle and Maven replacements (either on the `replaceAll` or on the `fileTree` level) and add a comment to each?

build-steps/release/archunit-examples-utils.gradle
```diff
@@ -27,18 +29,66 @@ subprojects {
...
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>3.5.6</version>
+            </plugin>
+        </plugins>
+    </build>
+    <modules>
+        <module>example-junit4</module>
+        <module>example-junit5</module>
```
schulzjo-tng, line 72: Reminder: We must not forget junit6 in the PR that is merged last

archunit-maven-test/pom.xml.template
schulzjo-tng: Good point. We still need to be aware of this regarding merge conflicts and merge order, though.

schulzjo-tng (approved):

archunit-maven-test/.mvn/wrapper/maven-wrapper.properties
hankem: Thank you! Good to know that it is not needed anymore since Maven wrapper 3.2.0.

build-steps/release/archunit-examples-utils.gradle
hankem: I've split the replacement.

archunit-maven-test/pom.xml.template
hankem: (I've dropped updating the pom.xml template from this PR.)

build-steps/release/archunit-examples-utils.gradle
hankem: → PR #1659.

PR #1440, opened 2025-03-31
Upgrade dependencies: Support Java 25's class file major version 69

hankem:
This PR upgrades

* [ASM](https://asm.ow2.io/versions.html): 9.7.1 → 9.8 – supporting Java 25's class file major version 69 
  * resolves #1439
* [Guava](https://github.com/google/guava): 33.3.1-jre → 33.4.8-jre
* [SLF4J](http://www.slf4j.org): 2.0.16 → 2.0.17
* [JUnit](https://junit.org/junit5/): 5.11.2 → 5.12.2

and for tests:
* [Log4j](https://logging.apache.org/log4j/2.x/): 2.24.1 → 2.24.3
* [AssertJ](https://assertj.github.io/doc/): 3.26.3 → 3.27.3

It reduces the memory footprint of  `ClassFileImporterSlowTest` by exluding classes from `sun` packages.
* resolves #1446

PR #1452, opened 2025-04-27
Manage dependencies via Gradle version catalog

hankem:
Unlike the manual dependency management in the `Map` `ext.dependency`. the Gradle [version catalog](https://docs.gradle.org/current/userguide/version_catalogs.html) `libs.version.toml` is [recognized by dependabot](https://docs.github.com/en/code-security/dependabot/ecosystems-supported-by-dependabot/supported-ecosystems-and-repositories#gradle).
Dependency updates (like #1440) should be triggered automatically (cf. hankem/ArchUnit#7), so we could probably revert #672 (although the daily nagging helps to remember that a new version is available 🙈).


archunit-junit/junit5/engine/build.gradle
```diff
@@ -12,7 +12,6 @@ dependencies {
     api project(path: ':archunit-junit5-engine-api')
     implementation project(path: ':archunit-junit')
     dependency.addGuava { dependencyNotation, config -> implementation(dependencyNotation, config) }
-    implementation dependency.slf4j
```
codecholeric, line 15: I think this one here is not "unused" though, no? I would only remove it, if it's really not used, not just cause it "randomly" compiles by pulling in the dependency transitively? 🤔

hankem: Oops, I don't know how I could miss the `Logger` in `ArchUnitTestDescriptor`... Thanks for checking thoroughly!

codecholeric:
Thanks a lot, that looks like a really good idea to improve maintainability 😃 I finally found some time and added some fixup suggestions, please let me know what you think!

hankem:
Cool, thank you! I'm happy to include these.

PR #1484, opened 2025-07-09
implement `DescribedPredicate#negate()`

### LINKED ISSUE #1480: Predicate.negate()
When trying to find a solution to #1479 I tried checking `JavaClass.Predicates.SimpleNameStartingWith("Abstract")` which I'm certain would have worked fine, except `negate()` is not the correct type and I couldn't find any method that allowed me to pass the predicate in a negated fashion.

hankem:
overwriting [`java.util.function.Predicate#negate()`](https://docs.oracle.com/javase/8/docs/api/java/util/function/Predicate.html#negate--) to return a [`DescribedPredicate`](https://javadoc.io/doc/com.tngtech.archunit/archunit/latest/com/tngtech/archunit/base/DescribedPredicate.html)

Resolves #1480

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/base/DescribedPredicate.java
+++ b/archunit/src/main/java/com/tngtech/archunit/base/DescribedPredicate.java
@@ -52,6 +52,15 @@ public abstract class DescribedPredicate<T> implements Predicate<T> {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Jakzi666 (commented):
I am not allowed to approve your PR, but it looks good to me and has my written approval - for what it's worth ;)

PR #1496, opened 2025-08-02
[Feature] Add Architecture Decision Record implementation to be used in ArchRule reason

### LINKED ISSUE #1495: Use an Architecture Decision Record (ADR) as a reason for a Rule
Hi all,

I'd like to have to possibility to describe the reason of a rule with an Architecture Decision (ADR), possibly in Markdown, using the following template: https://github.com/adr/madr

Thank you.
Regards.

RoRoche:
Hello,

This Pull Request includes:
- Contracts to define an ADR and its content
- Concrete implementation in Markdown using toString method overriding
- Unit test of a Markdown ADR generation

Resolves #1495

Thank you for reviewing.

Regards.

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/library/adr/Adr.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/adr/Adr.java
@@ -15,43 +15,64 @@
  */
 package com.tngtech.archunit.library.adr;
 
+import com.tngtech.archunit.PublicAPI;
+
 import java.util.List;
 import java.util.Optional;
+
+import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
 
 /**
  * Represents an Architecture Decision Record (ADR)
  * such as <a href="https://github.com/adr/madr">these templates</a>.
  */
+@PublicAPI(usage = ACCESS)
 public interface Adr {
+    @PublicAPI(usage = ACCESS)
     Optional<Metadata> metadata();
 
+    @PublicAPI(usage = ACCESS)
     Adr withMetadata(final Metadata metadata);
 
+    @PublicAPI(usage = ACCESS)
     String contextAndProblemStatement();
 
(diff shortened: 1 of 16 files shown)


hankem:
Thanks for your contribution! While I think it's nice to be able to define MADRs programmatically, this feature seems to be quite independent of ArchUnit. (The only dependency of your classes on ArchUnit is the `com.tngtech.archunit.PublicAPI` annotation...) Why do you think it should be in ArchUnit and not a separate library?

RoRoche:
Yes, you're right, it would be more complicated than I think to generate ADR from ArchUnit rule 😢

hankem:
Apologies for the late response!

I still think that 12 new classes that don't interact with ArchUnit at all, but build an MADR `String` that _can_ be used for  [`ArchRule#because(String)`](https://javadoc.io/doc/com.tngtech.archunit/archunit/latest/com/tngtech/archunit/lang/ArchRule.html#because(java.lang.String)) should – following the separation of concerns and to keep each tool focused _("Do one thing, and do it well.")_ – rather be released as an independent library, maybe similar to [`io.github.adr:e-adr`](https://github.com/adr/e-adr).

Does that sound reasonable?

RoRoche:
> Apologies for the late response!
> 
> I still think that 12 new classes that don't interact with ArchUnit at all, but build an MADR `String` that _can_ be used for [`ArchRule#because(String)`](https://javadoc.io/doc/com.tngtech.archunit/archunit/latest/com/tngtech/archunit/lang/ArchRule.html#because(java.lang.String)) should – following the separation of concerns and to keep each tool focused _("Do one thing, and do it well.")_ – rather be released as an independent library, maybe similar to [`io.github.adr:e-adr`](https://github.com/adr/e-adr).
> 
> Does that sound reasonable?

Yes I totally agree with you. Thank you.

PR #1550, opened 2025-11-10
Update Gradle Wrapper from 8.14.3 to 9.5.0 & also run tests with JDK 25

hankem:
* Update Gradle Wrapper from 8.14.3 to 9.5.0
  Release notes:
    * https://docs.gradle.org/9.0.0/release-notes.html
    * https://docs.gradle.org/9.1.0/release-notes.html (which supports Java 25)
    * https://docs.gradle.org/9.2.0/release-notes.html
    * https://docs.gradle.org/9.3.0/release-notes.html
    * https://docs.gradle.org/9.4.0/release-notes.html (which supports Java 26)
    * https://docs.gradle.org/9.5.0/release-notes.html

* Also run tests with JDK 25  
  Release notes:
    * https://openjdk.org/projects/jdk/25/


stefan-heilig-mw:
Thank you for your reply. We have managed to get it working since then. There was an error in our setup, which initially pointed to ArchUnit.

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
```diff
@@ -459,7 +459,10 @@ String covariantlyOverriddenCausingBridgeMethod() {
         JavaMethodCall callFromBridgeMethodToOverriddenOne = getOnlyElement(
                 new ClassFileImporter().importClasses(Parent.class, Child.class)
                         .get(Child.class)
-                        .getMethodCallsFromSelf());
+                        .getMethodCallsFromSelf()
```
StefanGraeber: It took me a really long time reproduce the second method call locally, as the build always falls back to java8 as the minimum supported version.
The easiest was in the end to set `testJavaVersion=25` manually in the `gradle.properties`.
Is it document somewhere how to choose the java version for local execution?

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
StefanGraeber, line 1137: very optional: remove duplicated "the"

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
```diff
@@ -1146,10 +1149,11 @@ void call(Target target) {
             }
         }
 
-        JavaMethodCall call = getOnlyElement(new ClassFileImporter()
-                .importClasses(Origin.class, Target.class)
-                .get(Origin.class)
-                .getMethodCallsFromSelf());
+        JavaMethodCall call = getOnlyElement(
+                new ClassFileImporter().importClasses(Origin.class, Target.class)
```
StefanGraeber: optional: keep line breaks as before
same for other tests

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
```diff
@@ -459,7 +459,10 @@ String covariantlyOverriddenCausingBridgeMethod() {
         JavaMethodCall callFromBridgeMethodToOverriddenOne = getOnlyElement(
                 new ClassFileImporter().importClasses(Parent.class, Child.class)
                         .get(Child.class)
-                        .getMethodCallsFromSelf());
+                        .getMethodCallsFromSelf()
+                        .stream()
+                        .filter(call -> call.getOrigin().isMethod())
```
StefanGraeber: this relies entirely on the fact the the new null check is added in the constructor which doesn't count as method.
I think this would be easier to understand if you write the filter in a way that makes clear that the target `Objects.requireNonNull` is excluded and adding a small comment that this is added in java 25+
both the bridge method and the new null check added in byte code where new to me.

As a user of ArchUnit I would be very surprised to see the new synthetic method call to Objects.requireNonNull.
Is that an expected migration effort for Java25 or do we need a special handling for this call (e.g. ignoring it)?

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAutomaticResolutionTest.java
```diff
@@ -349,9 +350,13 @@ void resolvesMethodCallTargetOwner() {
...
 
-        JavaClass javaClass = ImporterWithAdjustedResolutionRuns.disableAllIterationsExcept(MAX_ITERATIONS_FOR_ACCESSES_TO_TYPES_PROPERTY_NAME)
-                .importClass(Origin.class);
-        JavaMethodCall call = getOnlyElement(javaClass.getMethodCallsFromSelf());
+        JavaMethodCall call = getOnlyElement(
+                ImporterWithAdjustedResolutionRuns.disableAllIterationsExcept(MAX_ITERATIONS_FOR_ACCESSES_TO_TYPES_PROPERTY_NAME)
+                        .importClass(Origin.class)
+                        .getMethodCallsFromSelf()
+                        .stream()
+                        .filter(c -> c.getOrigin().isMethod())
```
StefanGraeber: if you are not extremly aware of the new null check, the logic behind this is quite surprising.
Could we filter for origin `Origin.resolvesMethodCallTargetOwner` instead? it would at least be clear that exactly that one method call in it is meant

or functionally equivalent: again use `.getMethod("resolvesMethodCallTargetOwner").getMethodCallsFromSelf()` as in other tests

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterLambdaDependenciesTest.java
```diff
@@ -284,8 +295,12 @@ Function<String, Target> call2() {
             }
         }
 
-        JavaClasses classes = new ClassFileImporter().importClasses(Target.class, Caller.class);
-        Set<JavaMethodCall> calls = classes.get(Caller.class).getMethodCallsFromSelf();
+        Set<JavaMethodCall> calls = new ClassFileImporter().importClasses(Target.class, Caller.class)
+                .get(Caller.class)
+                .getMethodCallsFromSelf()
+                .stream()
+                .filter(call -> call.getOrigin().isMethod())
```
StefanGraeber: `call.getOrigin().getName().matches("call\\d")`?
you filter for the name afterwards anyway and this would free us from isMethod (meant as isNotConstructor) which seems a pretty random check to me, if I wouldn't have read the commit message

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,7 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionSha256Sum=bd71102213493060956ec229d946beee57158dbd89d0e62b91bca0fa2c5f3531
```
StefanGraeber, line 3: did you drop the checksum intentionally?
according to https://gradle.org/release-checksums/ it is now for binary version 9.4.1 `2ab2958f2a1e51120c326cad6f385153bb11ee93b3c216c5fccebfdfbb7ec6cb`
and for the wrapper `55243ef57851f12b070ad14f7f5bb8302daceeebc5bce5ece5fa6edb23e1145c` (I verified that your new wrapper file has the correct checksum)

If we actually check it somewhere in the pipeline, I'd prefer to keep it, otherwise, deletion is fine

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
hankem: It's probably not documented. 🙈
(You could also use `-PtestJavaVersion=25`.)

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
hankem: I've tried to format the `new ClassFileImporter().importClasses`... invocations in a `getOnlyElement` function call consistently with these tests.

gradle/wrapper/gradle-wrapper.properties
hankem: I just did the update with
```
./gradlew wrapper --gradle-version latest && ./gradlew wrapper
```
and will now add the checksum, thanks for the hint! (Future updates can hopefully be done automatically by https://github.com/TNG/ArchUnit/actions/workflows/update-gradle-wrapper.yml again.)

I remember from https://github.com/gradle/actions/issues/12 that the `setup-gradle` action was said to validate the wrapper.

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAccessesTest.java
hankem: 1. For the tests, I've extracted `getMethodCallsFromClassWithoutAutomaticNullCheck` which documents the idea.
2. Users migrating to Java 25 will see the new `Objects.requireNonNull` call. I guess if that breaks their rules, they'll just have to see how to deal with it; the method call is factually there. 🤷‍♂️

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAutomaticResolutionTest.java
hankem: You're right!
I hope the updated test using [`getMethodCallsFromClassWithoutAutomaticNullCheck`](https://github.com/TNG/ArchUnit/pull/1550#discussion_r3234085612) is clearer again.

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterLambdaDependenciesTest.java
hankem: You're right!
I hope the updated test using [`getMethodCallsFromClassWithoutAutomaticNullCheck`](https://github.com/TNG/ArchUnit/pull/1550#discussion_r3234085612) is clearer again.

StefanGraeber (approved):

gradle/wrapper/gradle-wrapper.properties
hankem: It turned out that [`gradle/actions/setup-gradle@v6` does _not_ give an error on a wrong `distributionSha256Sum`](https://github.com/TNG/ArchUnit/actions/runs/25880072529/job/76057403251) (but Gradle seems to be taken from a cache).
I think this validation is being discussed in https://github.com/gradle/actions/issues/286.

[`gradle/actions/setup-gradle@v6` however errors on a manipulated `gradle-wrapper.jar`.](https://github.com/TNG/ArchUnit/actions/runs/25880587397/job/76059233668)

PR #1555, opened 2025-11-24
add support for dependencies from caught exceptions

### LINKED ISSUE #1232: Layer dependency rule ignores "catch" blocks
Hi!

I have a rule that verifies dependency between layers:
```
...
.whereLayer(LAYER_INFRASTRUCTURE).mayNotBeAccessedByAnyLayer()
...
```

The problem is that **it doesn't work with catch blocks defined in other layers**.

If I use an exception like this:
`MyException ex = new MyException ("you will be caught");`  then ArchUnit catches it.

However, if I use a catch block like so:
`} catch (MyException e) {`
then ArchUnit simply ignores it.

bannmann:
Referencing an exception class in the `catch` clause of a try-catch block can certainly be considered a dependency on that class, similar to a `throws` clause or an `instanceof` check: if the referenced class is renamed or deleted, the dependent code will break.

To prevent this from happening, one could write architecture rules that e.g. limit the use of a certain third-party library to a specific layer, or allow-list only certain "stable" or "vetted" classes.

Unfortunately, such rules will not notice when exception classes are referenced in `catch` clauses of try-catch blocks. The reason is that so far, ArchUnit does not generate a `Dependency` for them (it only does that once a member of the exception class is used, e.g. calling a method). This behavior defeats the purpose of implementing such a rule in the first place.

Thus, we change ArchUnit so that it considers each exception type referenced in `catch` clauses of try-catch blocks to be a dependency on that type. As the information is already available in the model via `TryCatchBlock.caughtThrowables`, all it takes is wiring it up to creation and handling of `Dependency` instances.

Resolves: #1232

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/Dependency.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/Dependency.java
@@ -19,6 +19,7 @@ import java.util.HashSet;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ComparisonChain;
@@ -51,6 +52,7 @@ import static java.util.Collections.singleton;
  * <li>a class (or method/constructor of this class) declares a type parameter referencing another class</li>
  * <li>a class (or method/constructor of this class) is annotated with an annotation of a certain type or referencing another class as annotation parameter</li>
  * <li>a method/constructor of a class references another class in a throws declaration</li>
+ * <li>a class references another class in a {@code catch} clause</li>
  * <li>a class references another class object (e.g. {@code Example.class})</li>
  * <li>a class references another class in an {@code instanceof} check</li>
  * </ul>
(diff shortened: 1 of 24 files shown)


archunit/src/test/java/com/tngtech/archunit/core/domain/testobjects/ClassWithDependencyOnTryCatchBlock.java
```diff
@@ -0,0 +1,39 @@
+package com.tngtech.archunit.core.domain.testobjects;
+
+import java.io.IOException;
+
+@SuppressWarnings("unused")
+public class ClassWithDependencyOnTryCatchBlock {
```
hankem:
```suggestion
public class ClassWithDependencyOnCaughtException {
```

archunit/src/test/java/com/tngtech/archunit/core/domain/testobjects/ClassWithDependencyOnTryCatchBlock.java
```diff
@@ -0,0 +1,39 @@
...
+        try {
+            throw new IOException();
+        } catch (IOException e) {
+        }
+    }
+
+    ClassWithDependencyOnTryCatchBlock()
+    {
+        try {
+            throw new IOException();
```
hankem: I'd use different exceptions for each of the cases to make the test even more clear.

hankem (commented):
Thank you so much for this contribution, which I think is a very nice addtion to ArchUnit!

On a first look, I could only spot very minor code style improvements.

hankem (commented):
I had a closer look, but still nothing but minor comments.

Feel free to already rebase and squash your commits, thank you.

hankem:
Ooh, I'm sincerely sorry that I left your PR haning for so long! 🙈
I'll quickly rebase it beyond a tiny technical conflict with #1655, and make sure we'll have the contribution in the next release.

hankem (approved):

PR #1565, opened 2025-12-10
Allow package info classes to be resolved as a dependency

### LINKED ISSUE #1564: package-info should be resolvable through DependencyResolutionProcess
when importing a class, which depends on a class which is not imported, the JavaPackage of that dependency class is not fully populated (with annotations, etc). This should be possible via DependencyResolutionProcess

wakingrufus:
Allow package info classes to be resolved as a dependency 
exclude package-info classes from package metrics calculations

Resolves #1564

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
@@ -16,6 +16,7 @@
 package com.tngtech.archunit.core.domain;
 
 import java.lang.annotation.Annotation;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
@@ -1429,6 +1430,13 @@ public final class JavaClass
         completionProcess.markClassHierarchyComplete();
     }
 
+    void completePackageInfoFrom(ImportContext context) {
+        this.javaPackage = JavaPackage.from(Arrays.asList(
+                this,
+                context.resolveClass(this.getPackageName() + ".package-info")));
+        completionProcess.markPackageComplete();
+    }
+
     private void completeSuperclassFrom(ImportContext context) {
         Optional<JavaClass> rawSuperclass = context.createSuperclass(this);
         if (rawSuperclass.isPresent()) {
@@ -1663,6 +1671,10 @@ public final class JavaClass
             @Override
             public void markDependenciesComplete() {
             }
+
+            @Override
+            public void markPackageComplete() {
+            }
         };
 
         abstract boolean hasFinished();
@@ -1683,6 +1695,8 @@ public final class JavaClass
 
         public abstract void markDependenciesComplete();
 
+        public abstract void markPackageComplete();
+
         static CompletionProcess start() {
             return new FullCompletionProcess();
         }
@@ -1701,6 +1715,7 @@ public final class JavaClass
         private boolean membersComplete = false;
         private boolean annotationsComplete = false;
         private boolean dependenciesComplete = false;
+        private boolean packageComplete = false;
 
         @Override
         boolean hasFinished() {
@@ -1711,7 +1726,8 @@ public final class JavaClass
                     && genericInterfacesComplete
                     && membersComplete
                     && annotationsComplete
-                    && dependenciesComplete;
+                    && dependenciesComplete
+                    && packageComplete;
         }
 
         @Override
@@ -1752,6 +1768,11 @@ public final class JavaClass
         @Override
         public void markDependenciesComplete() {
             this.dependenciesComplete = true;
+        }
+
+        @Override
+        public void markPackageComplete() {
+            this.packageComplete = true;
         }
     }
 
--- a/archunit/src/main/java/com/tngtech/archunit/library/metrics/MetricsComponents.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/metrics/MetricsComponents.java
@@ -19,6 +19,7 @@ import java.util.Collection;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableSet;
@@ -121,7 +122,12 @@ public final class MetricsComponents<T> extends ForwardingCollection<MetricsComp
     public static MetricsComponents<JavaClass> fromPackages(Collection<JavaPackage> packages) {
         ImmutableSet.Builder<MetricsComponent<JavaClass>> components = ImmutableSet.builder();
         for (JavaPackage javaPackage : packages) {
-            components.add(MetricsComponent.of(javaPackage.getName(), javaPackage.getClassesInPackageTree()));
+            components.add(MetricsComponent.of(
+                    javaPackage.getName(),
+                    javaPackage.getClassesInPackageTree().stream()
+                            .filter(jc -> !jc.getSimpleName().equals("package-info"))
+                            .collect(Collectors.toList())
+            ));
         }
         return MetricsComponents.of(components.build());
     }
--- a/archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTest.java
@@ -25,6 +25,8 @@ import com.google.common.collect.ImmutableSet;
 import com.tngtech.archunit.base.ArchUnitException.InvalidSyntaxUsageException;
 import com.tngtech.archunit.base.DescribedPredicate;
 import com.tngtech.archunit.base.HasDescription;
+import com.tngtech.archunit.core.domain.packageexamples.annotated.PackageLevelAnnotation;
+import com.tngtech.archunit.core.domain.packageexamples.annotated.WithinAnnotatedPackage;
 import com.tngtech.archunit.core.domain.testobjects.AAccessingB;
 import com.tngtech.archunit.core.domain.testobjects.AExtendingSuperAImplementingInterfaceForA;
 import com.tngtech.archunit.core.domain.testobjects.AReferencingB;
@@ -1593,6 +1595,13 @@ public class JavaClassTest {
         assertThat(JavaClass.Functions.GET_PACKAGE_NAME.apply(javaClass))
                 .as("result of GET_PACKAGE_NAME(clazz)")
                 .isEqualTo(javaClass.getPackageName());
+
+        JavaClass javaClassInAnnotatedPackage = importClassWithContext(WithinAnnotatedPackage.class);
+
+        assertThat(JavaClass.Functions.GET_PACKAGE.apply(javaClassInAnnotatedPackage)
+                .isAnnotatedWith(PackageLevelAnnotation.class))
+                .as("package info is available")
+                .isTrue();
     }
 
     @Test
--- a/archunit/src/test/java/com/tngtech/archunit/core/domain/JavaPackageTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/core/domain/JavaPackageTest.java
@@ -10,6 +10,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.BlockingQueue;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
 import com.tngtech.archunit.base.DescribedPredicate;
 import com.tngtech.archunit.core.domain.packageexamples.annotated.PackageLevelAnnotation;
@@ -423,6 +424,15 @@ public class JavaPackageTest {
         JavaPackage nonAnnotatedPackage = importPackage("packageexamples");
 
         assertThat(annotatedPackage.getPackageInfo()).isNotNull();
+        assertThat(annotatedPackage.getClasses())
+                .hasSize(3);
+        assertThat(annotatedPackage.getClasses().stream().map(JavaClass::getFullName).collect(Collectors.toList()))
+                .as("all classes and package-info are loaded as JavaClass")
+                .containsExactlyInAnyOrder(
+                        "com.tngtech.archunit.core.domain.packageexamples.annotated.package-info",
+                        "com.tngtech.archunit.core.domain.packageexamples.annotated.PackageLevelAnnotation",
+                        "com.tngtech.archunit.core.domain.packageexamples.annotated.WithinAnnotatedPackage"
+                );
 
         assertThatThrownBy(nonAnnotatedPackage::getPackageInfo)
                 .isInstanceOf(IllegalArgumentException.class)
--- a/archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAutomaticResolutionTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAutomaticResolutionTest.java
@@ -9,6 +9,7 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Optional;
 import java.util.Map;
 import java.util.function.Supplier;
 
@@ -22,10 +23,12 @@ import com.tngtech.archunit.core.domain.JavaConstructor;
 import com.tngtech.archunit.core.domain.JavaConstructorCall;
 import com.tngtech.archunit.core.domain.JavaConstructorReference;
 import com.tngtech.archunit.core.domain.JavaEnumConstant;
+import com.tngtech.archunit.core.domain.JavaField;
 import com.tngtech.archunit.core.domain.JavaFieldAccess;
 import com.tngtech.archunit.core.domain.JavaMethod;
 import com.tngtech.archunit.core.domain.JavaMethodCall;
 import com.tngtech.archunit.core.domain.JavaMethodReference;
+import com.tngtech.archunit.core.domain.JavaPackage;
 import com.tngtech.archunit.core.domain.JavaParameterizedType;
 import com.tngtech.archunit.core.domain.JavaType;
 import com.tngtech.archunit.core.domain.JavaWildcardType;
@@ -33,6 +36,7 @@ import com.tngtech.archunit.core.domain.ReferencedClassObject;
 import com.tngtech.archunit.core.domain.ThrowsDeclaration;
 import com.tngtech.archunit.core.domain.properties.HasAnnotations;
 import com.tngtech.archunit.core.importer.DependencyResolutionProcessTestUtils.ImporterWithAdjustedResolutionRuns;
+import com.tngtech.archunit.core.importer.testexamples.OtherClass;
 import com.tngtech.archunit.core.importer.testexamples.SomeAnnotation;
 import com.tngtech.archunit.core.importer.testexamples.annotatedclassimport.ClassWithUnimportedAnnotation;
 import com.tngtech.archunit.core.importer.testexamples.annotatedparameters.ClassWithMethodWithAnnotatedParameters;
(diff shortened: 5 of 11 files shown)


wakingrufus:
Happy New Year @codecholeric and @hankem !
Just a gentle reminder that I would appreciate if this PR could be looked at.
Thanks again!

wakingrufus:
Hi again @codecholeric  and @hankem. I know I have a few PRs open at the moment, but this one is quite important. I would really appreciate of someone could take a look at this please.

hankem:
I'm very sorry for the late response! I'll try to find more time. 🤞

wakingrufus:
> I'm very sorry for the late response! I'll try to find more time. 🤞

No problem, Thanks!

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTest.java
```diff
@@ -1595,6 +1597,15 @@ public void function_getPackage() {
                 .isEqualTo(javaClass.getPackageName());
     }
 
+    @Test
+    public void function_getPackageInfo() {
```
StefanGraeber: looking at the previous tests in this class, I would expect that a test with this name would test
`JavaClass.Functions.GET_PACKAGE_INFO`, which doesn't exist.
This test doesn't fit the pattern of the other tests.
I'm not sure whether we would want to have that function, as I'm not sure what I should return apart from the package itself

archunit/src/test/java/com/tngtech/archunit/core/importer/testexamples/packageinforesolution/ClassThatReferencesOtherClass.java
```diff
@@ -0,0 +1,7 @@
+package com.tngtech.archunit.core.importer.testexamples.packageinforesolution;
```
StefanGraeber, line 1: I don't understand this new package name.
is it important that it is subpackage of the package in which `OtherClass` is defined or shall just be any other package?

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
```diff
@@ -1429,6 +1430,13 @@ void completeClassHierarchyFrom(ImportContext context) {
         completionProcess.markClassHierarchyComplete();
     }
 
+    void completePackageInfoFrom(ImportContext context) {
+        this.javaPackage = JavaPackage.from(Arrays.asList(
```
StefanGraeber, line 1434: I'm not familiar with the class resolution yet, but this looks to me as if the package is re-created with just the current class and the `package-info` class.
If you look e.g. at the package `testexamples`, you have these classes (and more):
- `SomeClass`
- `OtherClass`
- `package-info`

assuming we are now in the `JavaClass` representing `SomeClass`, I expect that `getPackage().getClasses()` contains all three of these classes, but I don't see how it could return anything else then `Set.of(SomeClass, package-info)`, as this statement replaces the package and just passes these 2 classes into the new factory.
I don't see any call in `JavaPackage.from` checking the package of the class, which could contain the information.

It is also not clear to me how the dependencies and subpackages are build if all the other classes in teh package are missing.

Wouldn't it be much cleaner if we could resolve the `package-info` class when initially building the `JavaPackage`?
Why should each `JavaClass` in that package be responsible for resolving the `package-info`?

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAutomaticResolutionTest.java
```diff
@@ -620,6 +627,33 @@ class Innermost {
         assertThatType(outermost).matches(Outermost.class);
     }
 
+    @Test
+    public void automatically_resolves_packages() {
+        JavaClass otherClass = new ClassFileImporter()
+                .importClass(OtherClass.class);
+
+        JavaPackage checkedClassPackageInfo = otherClass.getPackage();
```
StefanGraeber, line 636: The `JavaPackage` has the methods `tryGetPackageInfo` and `getPackageInfo`.
please include in the test they work as well.
as one calls the other, I would prefer to tests sth like `assertThat(checkedClassPackageInfo.getPackageInfo()).isPresent()`, even thought this is kind of redundant with the `getAnnotations` method which also relies on that field.

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAutomaticResolutionTest.java
```diff
@@ -620,6 +627,33 @@ class Innermost {
...
+        assertThat(checkedClassPackageInfo.getAnnotations()).hasSize(1);
+        assertThat(checkedClassPackageInfo.isAnnotatedWith(SomeAnnotation.class)).isTrue();
+    }
```
StefanGraeber, line 639: When running `checkedClassPackageInfo.getClasses()`, this set now includes the `package-info`, which it didn't before.
I don't think that that class should be included in that Set as it is not a real class and shall instead be accessed via `JavaPackage.getPackageInfo()` which returns the `JavaClass` representing `package-info` as `HasAnnotations`. This change in type seems deliberate to me as this special class doesn't have all features of a `JavaClass`.

StefanGraeber (changes requested):
I dislike that the `package-info` is now included in `JavaPackage.getClasses()` (even as stub if it doesn't exist, which might be a breaking change for some use cases) and don't understand whether or how all classes in the package are retained when recreating the package in the `JavaClass` class.
IMO it also is the wrong location for the logic of resolving package info.

Can you explain in more detail which problem you want to solve?
To my understanding, the issue seems to be, that the `package-info` is not always resolved when you expect it to be loaded, but I'm not sure whether that's limited to loading single classes via `new ClassFileImporter()` or whether this also affects the "normal" execution with `@AnalyzeClasses` annotation

StefanGraeber:
I can't really localize the issue (and it might be a me-problem), but using windows 11 and IntelliJ, incremental builds fail with this error
```
> Task :archunit:compileJdk9mainJava FAILED
repopath\ArchUnit\archunit\src\jdk9main\java\com\tngtech\archunit\core\domain\Java9DomainPlugin.java:30: error: cannot access AnnotationPropertiesFormatter
                        .formatProperties(config -> config
                        ^
  class file for com.tngtech.archunit.core.domain.AnnotationFormatter$AnnotationPropertiesFormatter not found
1 error
```
as soon as I include your changes in `JavaClass`.
I can't reproduce the issue with adding some random new method in that class

wakingrufus:
> I dislike that the `package-info` is now included in `JavaPackage.getClasses()` (even as stub if it doesn't exist, which might be a breaking change for some use cases) and don't understand whether or how all classes in the package are retained when recreating the package in the `JavaClass` class. IMO it also is the wrong location for the logic of resolving package info.
> 
> Can you explain in more detail which problem you want to solve? To my understanding, the issue seems to be, that the package-info`is not always resolved when you expect it to be loaded, but I'm not sure whether that's limited to loading single classes via`new ClassFileImporter()`or whether this also affects the "normal" execution with`@AnalyzeClasses` annotation

Thanks for taking a look at this. I will add a more detailed description of the use case in the PR description.

wakingrufus:
> > I dislike that the `package-info` is now included in `JavaPackage.getClasses()` (even as stub if it doesn't exist, which might be a breaking change for some use cases) and don't understand whether or how all classes in the package are retained when recreating the package in the `JavaClass` class. IMO it also is the wrong location for the logic of resolving package info.
> > Can you explain in more detail which problem you want to solve? To my understanding, the issue seems to be, that the package-info`is not always resolved when you expect it to be loaded, but I'm not sure whether that's limited to loading single classes via`new ClassFileImporter()`or whether this also affects the "normal" execution with`@AnalyzeClasses` annotation
> 
> Thanks for taking a look at this. I will add a more detailed description of the use case in the PR description.



> > I dislike that the `package-info` is now included in `JavaPackage.getClasses()` (even as stub if it doesn't exist, which might be a breaking change for some use cases) and don't understand whether or how all classes in the package are retained when recreating the package in the `JavaClass` class. IMO it also is the wrong location for the logic of resolving package info.
> > Can you explain in more detail which problem you want to solve? To my understanding, the issue seems to be, that the package-info`is not always resolved when you expect it to be loaded, but I'm not sure whether that's limited to loading single classes via`new ClassFileImporter()`or whether this also affects the "normal" execution with`@AnalyzeClasses` annotation
> 
> Thanks for taking a look at this. I will add a more detailed description of the use case in the PR description.

@StefanGraeber Actually, I updated the associated issue, instead: https://github.com/TNG/ArchUnit/issues/1564

archunit/src/test/java/com/tngtech/archunit/core/importer/testexamples/packageinforesolution/ClassThatReferencesOtherClass.java
wakingrufus: It is important that this class is in a different package than `OtherClass`. other than that, this package could be anything

archunit/src/main/java/com/tngtech/archunit/core/domain/JavaClass.java
wakingrufus: because `initially building JavaPackage` is only something that happens if the package-info class is in the initial list of files passed to ClassfileImporter.  when class dependencies are resolved, if the other class lives in another (non-imported) package, only the package name is known. this method that is added allows us to resolve the rest of the package info for this package, even if the entire package was not included in the import.

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAutomaticResolutionTest.java
wakingrufus: Looks like these exist in JavaPackageTest:
```
    @Test
    public void test_getPackageInfo() {
        JavaPackage annotatedPackage = importPackage("packageexamples.annotated");
        JavaPackage nonAnnotatedPackage = importPackage("packageexamples");

        assertThat(annotatedPackage.getPackageInfo()).isNotNull();

        assertThatThrownBy(nonAnnotatedPackage::getPackageInfo)
                .isInstanceOf(IllegalArgumentException.class)
                .hasMessageContaining(nonAnnotatedPackage.getDescription() + " does not contain a package-info.java");
    }

    @Test
    public void test_tryGetPackageInfo() {
        JavaPackage annotatedPackage = importPackage("packageexamples.annotated");
        JavaPackage nonAnnotatedPackage = importPackage("packageexamples");

        assertThat(annotatedPackage.tryGetPackageInfo()).isPresent();
        assertThat(nonAnnotatedPackage.tryGetPackageInfo()).isEmpty();
    }
```

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAutomaticResolutionTest.java
wakingrufus: This is already the normal behavior in scanning a package. Ill add this assertion to JavaPackageTest.test_getPackageInfo() to show this:
```
        assertThat(annotatedPackage.getClasses().stream().map(JavaClass::getFullName).collect(Collectors.toList()))
                .as("all classes and package-info are loaded as JavaClass")
                .containsExactlyInAnyOrder(
                        "com.tngtech.archunit.core.domain.packageexamples.annotated.package-info",
                        "com.tngtech.archunit.core.domain.packageexamples.annotated.PackageLevelAnnotation",
                        "com.tngtech.archunit.core.domain.packageexamples.annotated.WithinAnnotatedPackage"
                );
```

wakingrufus:
> I can't really localize the issue (and it might be a me-problem), but using windows 11 and IntelliJ, incremental builds fail with this error
> 
> ```
> > Task :archunit:compileJdk9mainJava FAILED
> repopath\ArchUnit\archunit\src\jdk9main\java\com\tngtech\archunit\core\domain\Java9DomainPlugin.java:30: error: cannot access AnnotationPropertiesFormatter
>                         .formatProperties(config -> config
>                         ^
>   class file for com.tngtech.archunit.core.domain.AnnotationFormatter$AnnotationPropertiesFormatter not found
> 1 error
> ```
> 
> as soon as I include your changes in `JavaClass`. I can't reproduce the issue with adding some random new method in that class



> I can't really localize the issue (and it might be a me-problem), but using windows 11 and IntelliJ, incremental builds fail with this error
> 
> ```
> > Task :archunit:compileJdk9mainJava FAILED
> repopath\ArchUnit\archunit\src\jdk9main\java\com\tngtech\archunit\core\domain\Java9DomainPlugin.java:30: error: cannot access AnnotationPropertiesFormatter
>                         .formatProperties(config -> config
>                         ^
>   class file for com.tngtech.archunit.core.domain.AnnotationFormatter$AnnotationPropertiesFormatter not found
> 1 error
> ```
> 
> as soon as I include your changes in `JavaClass`. I can't reproduce the issue with adding some random new method in that class

I was able to run `./gradlew :archunit:compileJdk9mainJava` on ubuntu with openjdk 17. Maybe try running the gradle command directly?

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaClassTest.java
wakingrufus: thanks. this is fixed now.

archunit/src/test/java/com/tngtech/archunit/core/importer/testexamples/packageinforesolution/ClassThatReferencesOtherClass.java
StefanGraeber: I already suspected that, but "package in for resolution" doesn't really tell me anything.
maybe sth like 'otherclassrefrencefromanotherpackage`?
damn is that long :D
and not really better than yours... maybe think a bit about a better name, but if you don't find one, it's also fine for me

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAutomaticResolutionTest.java
StefanGraeber: yes, but since you re-create the JavaPackage, it might be worth checking that the re-created package behaves as expected in these methods

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAutomaticResolutionTest.java
StefanGraeber: how about packages without a package-info file?
IIRC you added somewhere a stub of it as JavaClass. If I don't have the package-info as explicit file in the package, I find it even weirder to have it returned as PackageInfo.
That is was like that before might not have been desired.
As you know, I'm new to the ArchUnit dev side (so far I only used it), but I find it surprising that this pseudo class is present at all in the result of getClasses() and even more surprising that it is now present even though I haven't even written that class.

I would prefer to never have package-info in this result

StefanGraeber:
> I was able to run ./gradlew :archunit:compileJdk9mainJava on ubuntu with openjdk 17. Maybe try running the gradle command directly?

whether I used the IntelliJ buttons or the gradle command via terminal made no difference.
The issue was related to something with the gradle logic of which packages need a rebuild. when cleaning the package before the build, everything is fine. It was really weird, I might try to reproduce it in more detail on another day

wakingrufus:
rebased on latest main

archunit/src/main/java/com/tngtech/archunit/library/metrics/MetricsComponents.java
```diff
@@ -121,7 +122,12 @@ public static <T> MetricsComponents<T> from(Collection<T> elements, Function<? s
     public static MetricsComponents<JavaClass> fromPackages(Collection<JavaPackage> packages) {
         ImmutableSet.Builder<MetricsComponent<JavaClass>> components = ImmutableSet.builder();
         for (JavaPackage javaPackage : packages) {
-            components.add(MetricsComponent.of(javaPackage.getName(), javaPackage.getClassesInPackageTree()));
+            components.add(MetricsComponent.of(
+                    javaPackage.getName(),
+                    javaPackage.getClassesInPackageTree().stream()
+                            .filter(jc -> !jc.getSimpleName().equals("package-info"))
```
StefanGraeber, line 128: Sorry for opening a new frontier, but I needed a fresh review from the start after a month...

this is a behavior change in case of having a package-info file.

please look at MetricsComponentsTest.creates_components_from_Java_packages and make explicit that the package-info is not imported (you need to create one in the test scenario, probably best in case 1 for top-level and subpackage.

adding the package-info might also require changes to the the other test methods. it would be interesting to see whether creates_components_from_Java_classes stick with 3 components.

I hope this doesn't break too much for our users becasue the old behavior is unexpected for me and I haven't seen MetricsCompment before

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaPackageTest.java
```diff
@@ -423,6 +424,15 @@ public void test_getPackageInfo() {
         JavaPackage nonAnnotatedPackage = importPackage("packageexamples");
 
         assertThat(annotatedPackage.getPackageInfo()).isNotNull();
+        assertThat(annotatedPackage.getClasses())
+                .hasSize(3);
+        assertThat(annotatedPackage.getClasses().stream().map(JavaClass::getFullName).collect(Collectors.toList()))
+                .as("all classes and package-info are loaded as JavaClass")
+                .containsExactlyInAnyOrder(
+                        "com.tngtech.archunit.core.domain.packageexamples.annotated.package-info",
```
StefanGraeber, line 432: should we really get the package-info in the classes?
I wouldn't expect that as user and don't really like having it as the package info is not really a class.

I just wanted to write that you can't get it via reflection as Class.forName("com.tngtech.archunit.core.domain.packageexamples.annotated.package-info") but you actually get it as an interface... there is always sth to learn about java...

What are your thoughts on including the package-info class in the classes of the package?
I have the feeling that it was like this before, but I don't like it.

What about a missing package-info file? I just checked that it is NOT included in that case. please extend the 
test, I write a separate comment https://github.com/TNG/ArchUnit/pull/1565/changes/f0fca6fef53d7c2b881d0a525aba79e2b4d049b7#r3332408963 for it

related/caused by to f0fca6f#r3136501904

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaPackageTest.java
```diff
@@ -423,6 +424,15 @@ public void test_getPackageInfo() {
...
+                        "com.tngtech.archunit.core.domain.packageexamples.annotated.PackageLevelAnnotation",
+                        "com.tngtech.archunit.core.domain.packageexamples.annotated.WithinAnnotatedPackage"
+                );
 
         assertThatThrownBy(nonAnnotatedPackage::getPackageInfo)
```
StefanGraeber, line 437: created by https://github.com/TNG/ArchUnit/pull/1565/changes/f0fca6fef53d7c2b881d0a525aba79e2b4d049b7#r3332389020 

I'd prefer to exchange the tested package by one with some classes in it (e.g. `packageexamples.first`) and asserting that those 2 classes but not the `package-info` are returned from `getClasses()`

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAutomaticResolutionTest.java
StefanGraeber: still open, please add an assertion like `assertThat(checkedClassPackageInfo.getPackageInfo()).isPresent()` to make explicit that the package-info is loaded.
Asserting it before the annotations will likely give better error messages if it ever breaks

archunit/src/test/java/com/tngtech/archunit/core/importer/ClassFileImporterAutomaticResolutionTest.java
StefanGraeber: related to https://github.com/TNG/ArchUnit/pull/1565/changes/f0fca6fef53d7c2b881d0a525aba79e2b4d049b7#r3332389020

StefanGraeber (changes requested):

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaPackageTest.java
StefanGraeber: I also just checked how the libraries Reflections and ClassGraph handle this:
Reflections includes it if you look for all subclasses of Object and disable the filterr preventing obejct sublcasses like this:
```
        Reflections reflections = new Reflections(
                new ConfigurationBuilder()
                        .forPackages("com.tngtech.archunit.core.domain.packageexamples.annotated")
                        .setScanners(Scanners.SubTypes.filterResultsBy(c -> true)) // disable the Object filter
        );
        Set<Class<?>> subTypesOfObject = reflections.getSubTypesOf(Object.class);
```
which is quite hacky.
ClassGraph doesn't include it at all but provides special methods for this

StefanGraeber: You also can't write the class literal `package-info.class` to reference it

PR #1569, opened 2025-12-28
Adjust AnalyzeClasses annotation to support individual classes as parameter

### LINKED ISSUE #1195: Add support to analyse list of classes
There are times when there are architecture rules that only apply to some classes, but that are not on the same package.
It would be nice if we could do something similar to this
```java
@AnalyzeClasses(classes = {First.class, Second.class})
public class MyArchitectureTest {

    @ArchTest
    public static final ArchRule myRule = ...
```

----
Something similar can be achieved using `locations`, but it implies either adding a lot of boilerplate code

```java
class ListOfClassesLocations implements LocationProvider {
    @Override
    public Set<Location> get(Class<?> testClass) {
        return Stream.of(First.class, Second.class).map(clazz -> Locations.ofClass(clazz).iterator().next()).collect(Collectors.toSet());
    }
}

@AnalyzeClasses(locations = ListOfClassesLocations.class)
public class MyArchitectureTest { ...
```

 or having to develop a custom solution like this one, that still isn't completely clean as it implies adding an extra annotation

```java
@Target(TYPE)
@Retention(RUNTIME)
public @interface AnalyzeListOfClasses {
    Class<?>[] value();
}
```

```java
public class ListOfClassesLocations implements LocationProvider {
    @Override
    public Set<Location> get(Class<?> testClass) {
        return Arrays.stream(testClass.getAnnotation(AnalyzeListOfClasses.class).value()).map(clazz -> Locations.ofClass(clazz).iterator().next()).collect(Collectors.toSet());
    }
}
```

```java
@AnalyzeClasses(locations = ListOfClassesLocations.class)
@AnalyzeListOfClasses({First.class, Second.class})
public class MyArchitectureTest {

    @ArchTest
    public static final ArchRule myRule = ...
```

TheManWhoStaresAtCode:
As suggested in the ticket adjust the annotation to have a new classes property to select individual classes as an additional option.

Resolves #1195

### DIFF
--- a/archunit-junit/junit4/src/main/java/com/tngtech/archunit/junit/AnalyzeClasses.java
+++ b/archunit-junit/junit4/src/main/java/com/tngtech/archunit/junit/AnalyzeClasses.java
@@ -38,6 +38,18 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  * a rule checking for no accesses to classes assignable to C will not fail, since ArchUnit does not know about the details
  * of class B, but only simple information like the fully qualified name. For information how to configure the import and
  * resolution behavior of missing classes, compare {@link ClassFileImporter}.
+ * <br><br>
+ * Classes to be analyzed can be specified in different ways:
+ * <ul>
+ *     <li>{@link #packages()} - specify package names as strings</li>
+ *     <li>{@link #packagesOf()} - specify packages relative to classes</li>
+ *     <li>{@link #classes()} - specify individual classes to analyze</li>
+ *     <li>{@link #locations()} - specify custom locations via {@link LocationProvider}</li>
+ *     <li>{@link #wholeClasspath()} - import all classes on the classpath</li>
+ * </ul>
+ * These options can be combined, whereby the final result is the union of all classes derived from each option,
+ * i.e. each option just adds more classes to be analyzed. If no option is specified, the package of the annotated
+ * test class will be imported.
  *
  * @see ArchUnitRunner
  * @see ClassFileImporter
@@ -83,4 +95,9 @@ public @interface AnalyzeClasses {
      * @return The {@link CacheMode} to use for this test class.
      */
     CacheMode cacheMode() default CacheMode.FOREVER;
+
+    /**
+     * @return Classes to be used for testing instead of packages
+     */
+    Class<?>[] classes() default {};
 }
--- a/archunit-junit/junit5/api/src/main/java/com/tngtech/archunit/junit/AnalyzeClasses.java
+++ b/archunit-junit/junit5/api/src/main/java/com/tngtech/archunit/junit/AnalyzeClasses.java
@@ -41,6 +41,18 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  * a rule checking for no accesses to classes assignable to C will not fail, since ArchUnit does not know about the details
  * of class B, but only simple information like the fully qualified name. For information how to configure the import and
  * resolution behavior of missing classes, compare {@link ClassFileImporter}.
+ * <br><br>
+ * Classes to be analyzed can be specified in different ways:
+ * <ul>
+ *     <li>{@link #packages()} - specify package names as strings</li>
+ *     <li>{@link #packagesOf()} - specify packages relative to classes</li>
+ *     <li>{@link #classes()} - specify individual classes to analyze</li>
+ *     <li>{@link #locations()} - specify custom locations via {@link LocationProvider}</li>
+ *     <li>{@link #wholeClasspath()} - import all classes on the classpath</li>
+ * </ul>
+ * These options can be combined, whereby the final result is the union of all classes derived from each option,
+ * i.e. each option just adds more classes to be analyzed. If no option is specified, the package of the annotated
+ * test class will be imported.
  *
  * @see ClassFileImporter
  */
@@ -87,4 +99,9 @@ public @interface AnalyzeClasses {
      * @return The {@link CacheMode} to use for this test class.
      */
     CacheMode cacheMode() default CacheMode.FOREVER;
+
+    /**
+     * @return Classes to be used for testing instead of packages
+     */
+    Class<?>[] classes() default {};
 }
--- a/archunit-junit/junit5/engine/src/test/java/com/tngtech/archunit/junit/internal/ArchUnitTestEngineTest.java
+++ b/archunit-junit/junit5/engine/src/test/java/com/tngtech/archunit/junit/internal/ArchUnitTestEngineTest.java
@@ -20,6 +20,7 @@ import com.tngtech.archunit.junit.ArchTest;
 import com.tngtech.archunit.junit.engine_api.FieldSelector;
 import com.tngtech.archunit.junit.engine_api.FieldSource;
 import com.tngtech.archunit.junit.internal.ArchUnitTestEngine.SharedCache;
+import com.tngtech.archunit.junit.internal.testexamples.AnalyzeClassesWithClassesProperty;
 import com.tngtech.archunit.junit.internal.testexamples.ClassWithPrivateTests;
 import com.tngtech.archunit.junit.internal.testexamples.ComplexMetaTags;
 import com.tngtech.archunit.junit.internal.testexamples.ComplexRuleLibrary;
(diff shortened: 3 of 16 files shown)


archunit-junit/junit4/src/main/java/com/tngtech/archunit/junit/AnalyzeClasses.java
```diff
@@ -38,6 +38,16 @@
...
+ * <br><br>
+ * Classes to be analyzed can be specified in different ways:
+ * <ul>
+ *     <li>{@link #packages()} - specify package names as strings</li>
+ *     <li>{@link #packagesOf()} - specify packages relative to classes</li>
+ *     <li>{@link #classes()} - specify individual classes to analyze</li>
+ *     <li>{@link #locations()} - specify custom locations via {@link LocationProvider}</li>
+ *     <li>{@link #wholeClasspath()} - import all classes on the classpath</li>
+ * </ul>
+ * These options can be combined. If no option is specified, the package of the annotated test class will be imported.
```
StefanGraeber: I don't find it obvious whether the different options are used as union or intersect.
AFAIK it is a union, so each option can only add more classes.
Should we make that clear?


docs/userguide/009_JUnit_Support.adoc and archunit-junit/src/main/java/com/tngtech/archunit/junit/internal/ClassCache.java confirm this assumption

archunit-junit/junit5/api/src/main/java/com/tngtech/archunit/junit/AnalyzeClasses.java
```diff
@@ -41,6 +41,16 @@
...
+ * <br><br>
+ * Classes to be analyzed can be specified in different ways:
+ * <ul>
+ *     <li>{@link #packages()} - specify package names as strings</li>
+ *     <li>{@link #packagesOf()} - specify packages relative to classes</li>
+ *     <li>{@link #classes()} - specify individual classes to analyze</li>
+ *     <li>{@link #locations()} - specify custom locations via {@link LocationProvider}</li>
+ *     <li>{@link #wholeClasspath()} - import all classes on the classpath</li>
+ * </ul>
+ * These options can be combined. If no option is specified, the package of the annotated test class will be imported.
```
StefanGraeber: same here. can we make clear that each option adds more classes and never reduce the selection?

docs/userguide/009_JUnit_Support.adoc and archunit-junit/src/main/java/com/tngtech/archunit/junit/internal/ClassCache.java confirm this assumption

archunit-junit/junit5/engine/src/test/java/com/tngtech/archunit/junit/internal/ArchUnitTestEngineTest.java
```diff
@@ -1104,6 +1106,21 @@ void a_class_with_analyze_classes_as_meta_annotation() {
             assertThat(request.scanWholeClasspath()).as("scan whole classpath").isTrue();
             assertThat(request.getImportOptions()).isEqualTo(expected.importOptions());
         }
+
+        @Test
+        void passes_AnalyzeClasses_with_new_classes_property_to_cache() {
```
StefanGraeber: I don't like names like "new" because it will not be new at some time. Please remove that from the test name

archunit-junit/junit5/engine/src/test/java/com/tngtech/archunit/junit/internal/ArchUnitTestEngineTest.java
```diff
@@ -1104,6 +1106,21 @@ void a_class_with_analyze_classes_as_meta_annotation() {
...
+            execute(createEngineId(), AnalyzeClassesWithClassesProperty.class);
+
+            verify(classCache).getClassesToAnalyzeFor(eq(AnalyzeClassesWithClassesProperty.class), classAnalysisRequestCaptor.capture());
+            ClassAnalysisRequest request = classAnalysisRequestCaptor.getValue();
+            AnalyzeClasses expected = AnalyzeClassesWithClassesProperty.class.getAnnotation(AnalyzeClasses.class);
+            assertThat(request.getClassesToAnalyze()).isEqualTo(expected.classes());
```
StefanGraeber, line 1117: why do you use the classes from the annotation here but an explicit empty for other properties like `getPackageNames`?
Feels inconsistent to me

archunit-junit/src/test/java/com/tngtech/archunit/junit/internal/ClassCacheTest.java
```diff
@@ -108,6 +108,15 @@ public void gets_all_classes_relative_to_class() {
         assertThat(classes.contain(getClass())).as("root class is contained itself").isTrue();
     }
 
+    @Test
+    public void gets_all_classes_specified() {
+        JavaClasses classes = cache.getClassesToAnalyzeFor(TestClass.class, new TestAnalysisRequest()
+                .withClassesToAnalyze(getClass()));
```
StefanGraeber: maybe `getClass()` -> `this.getClass()` in both places to make it more obvious?

StefanGraeber (commented):
looks good, just some comment and naming updates I'd prefer

archunit-junit/junit4/src/main/java/com/tngtech/archunit/junit/AnalyzeClasses.java
TheManWhoStaresAtCode: Thanks for pointing this out. I (hopefully) improved the documentation to address this point as well. Furthermore, I added a test to ensure that the behavior is indeed a union as pointed out by you.
For the later, I had the choice between using mocks and having a fast test run (< 1s) or use the actual class loader resulting in longer test runs (5 seconds or more). I decided to prefer test speed however, I'm interested in getting your opinion on that topic.

archunit-junit/src/test/java/com/tngtech/archunit/junit/internal/ClassCacheTest.java
```diff
@@ -112,6 +112,25 @@ public void gets_all_classes_specified() {
...
+
+        cache.getClassesToAnalyzeFor(TestClass.class, new TestAnalysisRequest()
+                .withPackages("com.tngtech.archunit.junit.internal")
+                .withPackagesRoots(Rule.class)
+                .withLocationProviders(TestLocationProviderOfClass_String.class)
+                .withClassesToAnalyze(getClass()));
+
+        verify(cacheClassFileImporter).importClasses(anySet(), locationCaptor.capture());
+        assertThat(locationCaptor.getValue())
+                .has(locationContaining("archunit/junit/internal"))
```
StefanGraeber: both `.withPackages("com.tngtech.archunit.junit.internal")` and `.withClassesToAnalyze(getClass())` satisfy this condition as ClassCacheTest resides in that package.
i.e. you can remove `withPackages("com.tngtech.archunit.junit.internal")` and the test is still green
please choose another package.
I tried java.lang.util but it doesn't work, likely because the setup restrict the scan to the current package.
adding an empty class in a subpackage and referencing that package works.
an empty package doesn't work, it must contain a class

archunit-junit/junit4/src/main/java/com/tngtech/archunit/junit/AnalyzeClasses.java
StefanGraeber: The new documentation looks good to me.
The new test is IMO sufficient regarding mocking or real class loader, but has another problem with poorly chosen test data. There is a separate comment for that issue

StefanGraeber (changes requested):
please rebase and cleanup the commits in the end. it's currently really hard to follow with merging old branches

the only blocking issue for me is now the overlapping testdata effectively skipping one assertion

archunit-junit/src/test/java/com/tngtech/archunit/junit/internal/ClassCacheTest.java
TheManWhoStaresAtCode: Good catch - I managed to get the test working with `java.utils`.

StefanGraeber: of course java.lang..util didn't work... that is called java.util 🤦 
nice that we both screwed up the name in the comment 😄

StefanGraeber (approved):
LGTM
please cleanup the git history before merging
(commit messages like "implement review comments" should not survive in the end ;)

StefanGraeber (approved):

PR #1579, opened 2026-01-19
fix description of targetOwner predicate

wakingrufus:
targetOwner was producing the same message as target() eg. "access target where target is annotated with..." instead of "access target where owner is annotated with..."

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaAccess.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/domain/JavaAccess.java
@@ -15,9 +15,6 @@
  */
 package com.tngtech.archunit.core.domain;
 
-import java.util.Collections;
-import java.util.Set;
-
 import com.tngtech.archunit.PublicAPI;
 import com.tngtech.archunit.base.ChainableFunction;
 import com.tngtech.archunit.base.DescribedPredicate;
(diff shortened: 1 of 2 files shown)


archunit/src/main/java/com/tngtech/archunit/core/domain/JavaAccess.java
```diff
@@ -15,9 +15,6 @@
  */
 package com.tngtech.archunit.core.domain;
 
-import java.util.Collections;
```
StefanGraeber, line 18: please fix you code formatting as described in https://github.com/TNG/ArchUnit/blob/main/CONTRIBUTING.md#formatting

archunit/src/test/java/com/tngtech/archunit/core/domain/JavaAccessTest.java
```diff
@@ -60,6 +60,32 @@ public void origin_predicate() {
...
+        assertThat(predicate).rejects(anyAccess());
+    }
+
+    @Test
+    public void targetOwner_predicate() {
+        DescribedPredicate<JavaAccess<?>> predicate =
+                JavaAccess.Predicates.targetOwner(DescribedPredicate.<JavaClass>alwaysTrue().as("some text"));
+
+        assertThat(predicate)
+                .hasDescription("owner some text")
```
StefanGraeber, line 82: this text doesn't specify that the owner is related in any way to the target or any hint what is owned.
Is it the target owner or the origin owner?

I think the issue you are fixing here also exists for the origin(Owner) in exactly the same constellation.

The test also doesn't test that the correct object is passed to the nested predicate (which also not done by origin_predicate)

wakingrufus:
@StefanGraeber thanks for taking a look! I'll close this in favor of https://github.com/TNG/ArchUnit/pull/1603/

PR #1580, opened 2026-01-19
add haveDependenciesThat condition to  ArchConditions

wakingrufus:
there is not a good way to craft such a condition with current Public APIs

the use case I want this for is:
```
noClasses()
.should(
  haveAnyDependenciesThat(
    not(resideInSamePackage())
      .and(dependencyTarget(annotatedWith(Deprecated.class)))
    )
  )
``` 
where `resideInSamePackage()` is a DescribedPredicate<Dependency> which requires comparing the target and the owner in the same condition

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/lang/conditions/ArchConditions.java
+++ b/archunit/src/main/java/com/tngtech/archunit/lang/conditions/ArchConditions.java
@@ -315,6 +315,14 @@ public final class ArchConditions {
     }
 
     @PublicAPI(usage = ACCESS)
+    public static ArchCondition<JavaClass> haveAnyDependenciesThat(DescribedPredicate<? super Dependency> predicate) {
+        return new AnyDependencyCondition(
+                "have any dependencies that " + predicate.getDescription(),
+                predicate,
+                GET_DIRECT_DEPENDENCIES_FROM_SELF);
+    }
+
+    @PublicAPI(usage = ACCESS)
     public static ArchCondition<JavaClass> transitivelyDependOnClassesThat(DescribedPredicate<? super JavaClass> predicate) {
         return new TransitiveDependencyCondition(predicate);
     }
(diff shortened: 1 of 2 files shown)


StefanGraeber (approved):
Looks good to me :)

wakingrufus:
Thanks! I rebased. Just need approval for workflows

archunit/src/test/java/com/tngtech/archunit/lang/conditions/ArchConditionsTest.java
```diff
@@ -164,6 +166,38 @@ public void only_have_dependents_where() {
...
+        JavaClasses classes = importClasses(CallingClass.class, SomeClass.class);
+        JavaClass callingClass = classes.get(CallingClass.class);
+
+        assertThat(haveAnyDependenciesThat(alwaysFalse()))
+                .checking(callingClass)
+                .haveAtLeastOneViolationMessageMatching(String.format(".*%s.*%s.*",
+                        quote(CallingClass.class.getName()), quote(SomeClass.class.getName())));
+
+        assertThat(haveAnyDependenciesThat(DescribedPredicate.<Dependency>alwaysTrue().as("custom")))
+                .hasDescription("have dependencies that custom")
```
hankem: Whoopsie. 😉
```suggestion
                .hasDescription("have any dependencies that custom")
```

StefanGraeber (approved):
I guess you still need the signed-off line in the commit message.
otherwise this change is ready to submit

hankem:
Ah yes, and please squash the two commits. Thank you!

wakingrufus:
> Ah yes, and please squash the two commits. Thank you!

rebased and squashed. Thanks again!

archunit/src/main/java/com/tngtech/archunit/lang/conditions/ArchConditions.java
```diff
@@ -314,6 +314,14 @@ public static ArchCondition<JavaClass> dependOnClassesThat(DescribedPredicate<?
                 GET_DIRECT_DEPENDENCIES_FROM_SELF);
     }
 
+    @PublicAPI(usage = ACCESS)
+    public static ArchCondition<JavaClass> haveAnyDependenciesThat(DescribedPredicate<? super Dependency> predicate) {
```
hankem, line 318: I wonder whether there can be any confusion about the name. Will it be clear that `haveAnyDependenciesThat(…)` means _"have any direct dependencies that …"_ as opposed to _"have any transitive dependencies that …"_?

hankem (approved):
Thank you for your contribution!
Just one last thought from my side, as naming things is hard...

archunit/src/main/java/com/tngtech/archunit/lang/conditions/ArchConditions.java
StefanGraeber: other methods like `dependOnClassesThat` (or getting all dependencies) also mean just direct dependencies.
So if it's no problem there, it shouldn't be here.
IMO our users are used to this and the current naming is consistent

StefanGraeber (approved):

PR #1602, opened 2026-04-10
enforce unix linebreaks to allow builds on windows machines

StefanGraeber:
When running a gradle build on a windows machine, spotless complaints about incorrect formatting after a fresh checkout, which is not fixed by `spotlessApply`. By enforcing UNIX line breaks in spotless and git, the build and style check also work on windows machines

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Enforce Unix line endings (LF) for all java files
+*.java text=auto eol=lf


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Enforce Unix line endings (LF) for all text files
+* text=auto eol=lf
```
hankem: We _could_ use
```suggestion
* text=auto eol=lf
*.bat text eol=crlf
```
and keep `gradlew.bat` as it was.
(While modern Windows should be able to handle .bat files with LF endings fine, `cmd.exe` historically required CRLF endings in batch files.)

Alternatively, we could probably define `eol` settings for `*.java` files only, as we might not have to change the `.svg` either.

buildSrc/src/main/groovy/archunit.java-code-quality-conventions.gradle
```diff
@@ -17,6 +17,8 @@ spotbugsMain {
 spotbugsTest.enabled = false
 
 spotless {
+    lineEndings 'UNIX'
```
hankem: Do we need this configuration?

[Spotless' documentation](https://github.com/diffplug/spotless/tree/main/plugin-gradle#line-endings-and-encodings-invisible-stuff) mentions
> The default value is `GIT_ATTRIBUTES_FAST_ALLSAME`, and _we highly recommend that you **do not change** this value._ Git has opinions about line endings, and if Spotless and git disagree, then you're going to have a bad time.

Did you see a different behavior with this extra configuration?

hankem (commented):
Thank you for the contribution!

I initially wasn't able to reproduce a problem with `spotlessCheck` on Windows (11, using git version 2.47.1.windows.2, neither in Powershell nor in git bash – [explicitly executing the `spotlessCheck` task on Windows GitHub action runners](https://github.com/TNG/ArchUnit/actions/runs/25255209111/job/74053392598) did not fail either), but I see failures when I have `core.eol` configuration in my `~/.gitconfig`, which are prevented by your PR.

I have two tiny comments for potential simplifications. Could you check whether they work for your use case?

.gitattributes
StefanGraeber: since I don't used the bat file, I don't care too much about that. I also thought about it before and whether I should exclude it from the rule.
both your proposal of explicit bat file exclusion and explicit java rule work as well together with the explicit UNIX  rule in the spotless config.
without the explicit spotless config, it works as well.

buildSrc/src/main/groovy/archunit.java-code-quality-conventions.gradle
StefanGraeber: after having the correct gitattribute config, it is no longer needed.
I discovered this change before the git attribute and didn't check afterwards, whether it's needed.
I reverted this.

StefanGraeber:
@hankem do you get any useful information out of the failed windows pipeline?
I just see an error without any message

hankem:
[`Execution failed for task ':archunit-junit5-api:spotlessJavaCheck'.` on Windows](https://github.com/TNG/ArchUnit/actions/runs/25306589338/job/74183805550?pr=1602#step:7:200)
> The following files had format violations:
> ```
>      src\main\java\com\tngtech\archunit\junit\AnalyzeClasses.java
>          @@ -1,18 +1,18 @@
>          -/*\r\n
> ```

I think `.gitattributes` might only be applied when `git` writes a file to the working tree. When the file doesn't change when switching branches, the CI build for your branch still sees the line endings from `main`.

What puzzles me is why this failure didn't occur in the first version of your  PR.

hankem:
I was convinced that the failure would disappear when the `.gitattributes` is in place during the first checkout already, and merged the PR to `main`. However it seems that my theory was wrong... 🙈

PR #1603, opened 2026-04-10
fix descriptions of targetOwner and originOwner predicates in JavaAccess

StefanGraeber:
targetOwner and originOwner() were producing the same message as target() and origin() e.g. "access target where target is annotated with..." instead of "access target where target owner is annotated with..." The existing test is also updated to assert that the correct element is passed to the given predicate, for which new testUtils are introduced.

replacement for #1579

### DIFF
--- a/archunit/src/test/java/com/tngtech/archunit/core/domain/TestUtils.java
+++ b/archunit/src/test/java/com/tngtech/archunit/core/domain/TestUtils.java
@@ -2,10 +2,12 @@ package com.tngtech.archunit.core.domain;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Predicate;
 
 import com.google.common.collect.ImmutableSet;
 import com.google.common.io.Files;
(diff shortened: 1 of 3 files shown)


archunit/src/test/java/com/tngtech/archunit/core/domain/TestUtils.java
```diff
@@ -102,8 +106,47 @@ public static AccessesSimulator simulateCall() {
...
+
+        @Override
+        public boolean test(T t) {
+            this.capturedValues.add(t);
+            return check.test(t);
+        }
+
+        public List<T> getCapturedValues() {
+            return Collections.unmodifiableList(new ArrayList<>(capturedValues));
+        }
```
hankem: I wouldn't introduce unused methods, but keep it simple:
```suggestion
    public static <T> CapturingDescribedPredicate<T> alwaysTrueCapturingPredicateWithDescription(String description) {
        return new CapturingDescribedPredicate<T>(true, description);
    }

    public static class CapturingDescribedPredicate<T> extends DescribedPredicate<T> {
        private final List<T> capturedValues = new ArrayList<>();
        private final Predicate<T> check;

        CapturingDescribedPredicate(boolean result, String description) {
            super(description);
            this.check = o -> result;
        }

        @Override
        public boolean test(T t) {
            this.capturedValues.add(t);
            return check.test(t);
        }
```
(Note that this also removes the need to `import java.util.Collections;`.)

hankem (approved):
Thanks for the PR!

archunit/src/test/java/com/tngtech/archunit/core/domain/TestUtils.java
StefanGraeber: I removed all unused methods and imports added by me

StefanGraeber:
@wakingrufus this PR is merged now. thanks again for your original PR #1579 finding the issue 🙂

PR #1609, opened 2026-04-21
enable using classfile API on JDK 24+ via asm-jdk-bridge

### LINKED ISSUE #1608: Use classfile API on JDK 26+
JDK 24+ ships a [classfile API](https://openjdk.org/jeps/484) which is a built-in replacement for ASM. 
https://github.com/raphw/asm-jdk-bridge is available which uses this classfile API to create ASM object, which can make the transition to this API smooth.

wakingrufus:
fixes #1608

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -6,6 +6,7 @@ assertj = "3.27.7"
 
 [libraries]
 asm = { group = "org.ow2.asm", name = "asm", version = "9.10.1" }
+asmjdkbridge = { module = "codes.rafael.asmjdkbridge:asm-jdk-bridge", version = "0.0.13"}
 guava = { group = "com.google.guava", name = "guava", version = "33.6.0-jre" }
 slf4j = { group = "org.slf4j", name = "slf4j-api", version = "2.0.18" } # in sync with buildSrc/src/main/resources/release_check/archunit.pom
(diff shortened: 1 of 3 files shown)


gradle/libs.versions.toml
```diff
@@ -6,6 +6,7 @@ assertj = "3.27.7"
 
 [libraries]
 asm = { group = "org.ow2.asm", name = "asm", version = "9.9.1" }
+asmjdkbridge = { module = "codes.rafael.asmjdkbridge:asm-jdk-bridge", version = "0.0.13"}
```
wakingrufus, line 9: This is not exactly a production-grade library. You may prefer to copy the code from it and own that yourself.

.github/workflows/build.yml
```diff
@@ -48,6 +48,7 @@ jobs:
           - 11
           - 17
           - 21
+          - 25
```
hankem, line 51: I think the tests can run against Java 25 after #1550.

wakingrufus: oh nice, thanks. Ill rebase this once that PR is merged

hankem: I've just merged it.

schulzjo-tng:
As discussed in #1608, we will wait until this becomes a problem, or until we no longer support Java < 24.

PR #1619, opened 2026-05-17
Do not include web fonts in javadoc jar

hankem:
Commit 912abfd95233b477ffe98d362246fe5a6d348c54 advanced `maxSupportedJavaVersion`, which is also used as `languageVersion` for the `javadocTool`, from Java 21 to Java 25.

Unlike javadoc 21, javadoc 25 by default includes ~4 MB of fonts in every `*-javadoc.jar`:
`javadoc --no-fonts` reduces the total size of published `*-javadoc.jar`s by a factor of ~10, e.g.
- `archunit-javadoc.jar` from 5.4 MiB to 1.6 MiB (was 1.4 MiB with javadoc 21)
- `archunit-junit5-api-javadoc.jar` from 4.0 MiB to 0.1 MiB (was 0.1 MiB with javadoc 21)

### DIFF
--- a/buildSrc/src/main/groovy/archunit.java-conventions.gradle
+++ b/buildSrc/src/main/groovy/archunit.java-conventions.gradle
@@ -15,4 +15,5 @@ tasks.withType(JavaCompile) { Task task ->
 
 javadoc {
     options.addBooleanOption('html5', true)
+    options.addBooleanOption("-no-fonts", true)
 }


buildSrc/src/main/groovy/archunit.java-conventions.gradle
```diff
@@ -15,4 +15,5 @@ tasks.withType(JavaCompile) { Task task ->
 
 javadoc {
     options.addBooleanOption('html5', true)
-}
\ No newline at end of file
+    options.addBooleanOption("-no-fonts", true)
```
schulzjo-tng, line 18: Did you check that this works? I think the option should be `--no-fonts` (two dashes), see e.g. https://github.com/apache/maven-javadoc-plugin/pull/1309/changes

zimmermj (approved):

buildSrc/src/main/groovy/archunit.java-conventions.gradle
hankem: Oh, for sure. 🙂

With `addBooleanOption("--no-fonts", true)`, 
```
./gradlew javadoc
```
fails with
> ```
> > Task :archunit:javadoc FAILED
> error: invalid flag: ---no-fonts
> Did you mean: --no-fonts
> For more details on available options, use --help or --help-extra
> 1 error
> ```

schulzjo-tng (approved):

PR #1629, opened 2026-05-25
perf(metrics): optimize LakosMetrics transitive dependency graph

### LINKED ISSUE #1628: perf(metrics): Optimize transitive reachability in LakosMetrics via SCC condensation and BitSet propagation
### **Problem Description**
Currently, `LakosMetrics` computes architectural metrics by querying transitive component dependencies. The underlying implementation in `MetricsComponentDependencyGraph` resolves these queries by executing a fresh Depth-First Search (DFS) from scratch for every single component. 

On complex codebase graphs with shared downstream subgraphs or deep/cyclic dependencies, this approach results in heavily redundant traversals. The runtime complexity scales toward O(V * (V + E)), causing high CPU overhead and severe memory allocation pressure due to the repeated generation of temporary `HashSet` instances.

### **Code Locations**
* `com.tngtech.archunit.library.metrics.LakosMetrics` (where `getNumberOfTransitiveDependencies` is called iteratively)
* `com.tngtech.archunit.library.metrics.MetricsComponentDependencyGraph` (where `getTransitiveDependenciesOf` triggers a fresh traversal via `addTransitiveDependenciesFrom`)

### **Proposed Solution**
Instead of a query-time DFS, we can shift to a robust, one-time precomputation model when the dependency graph is constructed. The proposed architecture applies graph-theoretic optimizations to decouple query performance from graph depth:

1. **Identify Strongly Connected Components (SCCs):** Use [Kosaraju's algorithm ](https://en.wikipedia.org/wiki/Kosaraju%27s_algorithm)to find and collapse component cycles.
2. **DAG Condensation:** Condense the original graph into a Directed Acyclic Graph (DAG) where each node represents an SCC.
3. **Topological BitSet Propagation:** Determine the topological order of the DAG and propagate downstream reachability using fast bitwise operations (`BitSet.or()`). This ensures that shared tails are computed exactly once and combined at the hardware level.
4. **O(1) Lookups:** Map the final bitsets back to their respective components and store them in an immutable structure, making all subsequent `getTransitiveDependenciesOf` calls instant.

### **Why This Is Better**

| Dimension | Current Implementation | Proposed Solution |
| :--- | :--- | :--- |
| **Time Complexity** | O(V * (V + E)) worst-case | **O(V + E)** for precomputation, **O(1)** per query |
| **Redundant Work** | High (re-traverses identical sub-paths) | **Zero** (sub-paths evaluated once on the DAG) |
| **Allocation Footprint**| High (repeated transient `HashSet` creation) | **Minimal** (reuses primitive indices & bit vectors) |
| **Call-site Overhead**| Increases with graph size and density | **Constant-time** index array offset lookup |
| **Stack Security** | Vulnerable to recursion depth limits | **Guaranteed safe** via explicit iterative stack frames |

ThanosTsiamis:
## Description

This Pull Request addresses the performance hotspot in `LakosMetrics` by overhauling how transitive reachability is computed within `MetricsComponentDependencyGraph`. 

### The Problem
Previously, `getTransitiveDependenciesOf()` executed a fresh Depth-First Search (DFS) from scratch for every single component. On complex architecture graphs with shared downstream subgraphs or deep/cyclic dependencies, this caused heavily redundant traversals, driving the runtime complexity up to O(V * (V + E)). It also introduced significant garbage collection pressure by instantiating new `HashSet` collections on every query.

### The Solution
Instead of query-time DFS traversals, this PR introduces a robust, primitive-backed precomputation engine executed once during the graph's initialization:

1. **Cycle Detection (Kosaraju's Algorithm):** `determineStronglyConnectedComponents()` runs a two-pass iterative DFS (forward to find finishing times, reverse on the transposed graph) to safely group cyclic dependencies into Strongly Connected Components (SCCs).
2. **DAG Condensation:** The component graph is condensed into a Directed Acyclic Graph (DAG) where each node represents an isolated SCC.
3. **Topological BitSet Propagation:** The DAG is sorted topologically, and downstream reachability is propagated using fast bitwise operations (`BitSet.or()`). This ensures that shared downstream paths are evaluated exactly once.
4. **O(1) Queries:** The final reachability matrices are mapped back to their respective components and cached, changing `getTransitiveDependenciesOf()` from an expensive graph traversal into an O(1) map lookup.

### Additional Robustness & Safeguards
* **Stack Safety:** The algorithm completely avoids recursion in favor of an explicit iterative stack approach (`ArrayDeque<TraversalFrame>`), protecting the JVM against `StackOverflowError` on deep, linear dependency chains.
* **Cache Locality:** Internal graph representations are flattened into primitive `int[][]`, `int[]`, and `boolean[]` arrays to optimize CPU cache locality and reduce memory footprint.
* **Duplicate Resilience:** Rather than utilizing a strict `ImmutableMap.Builder` (which crashes with an `IllegalArgumentException` if duplicate elements are fed from upstream), the precomputation safely compiles results into an intermediate `LinkedHashMap` before generating the final immutable copy.
* **Semantic Parity:** Existing cycle and self-reachability rules are perfectly preserved. A component only transitively reaches itself if it belongs to an SCC of size > 1, or contains a direct self-loop.

---

## Related Issue
Closes #1628

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/library/metrics/MetricsComponentDependencyGraph.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/metrics/MetricsComponentDependencyGraph.java
@@ -15,13 +15,21 @@
  */
 package com.tngtech.archunit.library.metrics;
 
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.BitSet;
 import java.util.Collection;
+import java.util.Deque;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Function;
 
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSetMultimap;
 import com.google.common.collect.SetMultimap;
@@ -29,11 +37,14 @@ import com.google.common.collect.SetMultimap;
 class MetricsComponentDependencyGraph<T> {
     private final SetMultimap<MetricsComponent<T>, MetricsComponent<T>> outgoingComponentDependencies;
     private final SetMultimap<MetricsComponent<T>, MetricsComponent<T>> incomingComponentDependencies;
+    private final Map<MetricsComponent<T>, Set<MetricsComponent<T>>> transitiveComponentDependencies;
 
     private MetricsComponentDependencyGraph(Iterable<MetricsComponent<T>> components, Function<T, Collection<T>> getDependencies) {
-        ImmutableSetMultimap<MetricsComponent<T>, MetricsComponent<T>> componentDependencies = createComponentDependencies(components, getDependencies);
+        ImmutableList<MetricsComponent<T>> componentList = ImmutableList.copyOf(components);
+        ImmutableSetMultimap<MetricsComponent<T>, MetricsComponent<T>> componentDependencies = createComponentDependencies(componentList, getDependencies);
         this.outgoingComponentDependencies = componentDependencies;
         this.incomingComponentDependencies = componentDependencies.inverse();
+        this.transitiveComponentDependencies = new TransitiveDependencyPrecomputation<>(componentList, componentDependencies).getTransitiveDependenciesByComponent();
     }
 
     private ImmutableSetMultimap<MetricsComponent<T>, MetricsComponent<T>> createComponentDependencies(Iterable<MetricsComponent<T>> components, Function<T, Collection<T>> getDependencies) {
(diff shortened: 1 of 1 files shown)


archunit/src/main/java/com/tngtech/archunit/library/metrics/MetricsComponentDependencyGraph.java
```diff
@@ -77,27 +87,316 @@ Set<MetricsComponent<T>> getDirectDependenciesTo(MetricsComponent<T> target) {
...
+            StronglyConnectedComponents stronglyConnectedComponents = determineStronglyConnectedComponents();
+            CondensedGraph condensedGraph = condense(stronglyConnectedComponents);
+            BitSet[] reachableStronglyConnectedComponents = determineReachableStronglyConnectedComponents(condensedGraph);
+            return mapToTransitiveDependenciesByComponent(stronglyConnectedComponents, reachableStronglyConnectedComponents);
+        }
+
+        private StronglyConnectedComponents determineStronglyConnectedComponents() {
+            List<Integer> nodesByFinishingOrder = determineFinishingOrder();
+            int[] stronglyConnectedComponentIndexByComponentIndex = new int[components.size()];
+            java.util.Arrays.fill(stronglyConnectedComponentIndexByComponentIndex, -1);
```
hankem: Let's `import java.util.Arrays;`
```suggestion
            Arrays.fill(stronglyConnectedComponentIndexByComponentIndex, -1);
```
and also simplify the three usages of `java.util.Arrays.copyOf` below
(or even use static imports for _`fill`_ and _`copyOf`_).

hankem (approved):
Thank you for this improvement!

On my system, it reduces the time to compute
```
ArchitectureMetrics.lakosMetrics(MetricsComponents.fromClasses(classes))
```
for [`hibernate-core:7.3.6`](https://github.com/hibernate/hibernate-orm/tree/7.3.6/hibernate-core) (with 37,281 classes)
- from **(101±2) s** with ArchUnit 1.4.2
- by a factor of more than 16 to **(6.2±0.3) s**. 🎉

archunit/src/main/java/com/tngtech/archunit/library/metrics/MetricsComponentDependencyGraph.java
```diff
@@ -77,27 +87,316 @@ Set<MetricsComponent<T>> getDirectDependenciesTo(MetricsComponent<T> target) {
...
+
+        private Map<MetricsComponent<T>, Set<MetricsComponent<T>>> getTransitiveDependenciesByComponent() {
+            // Collapse cycles first, then compute reachability once on the condensed DAG.
+            StronglyConnectedComponents stronglyConnectedComponents = determineStronglyConnectedComponents();
+            CondensedGraph condensedGraph = condense(stronglyConnectedComponents);
+            BitSet[] reachableStronglyConnectedComponents = determineReachableStronglyConnectedComponents(condensedGraph);
+            return mapToTransitiveDependenciesByComponent(stronglyConnectedComponents, reachableStronglyConnectedComponents);
+        }
+
+        private StronglyConnectedComponents determineStronglyConnectedComponents() {
```
hankem, line 142: I'm absolutely not an expert on graph algorithms, but some AI code review suggested:

>  Kosaraju's algorithm requires two passes. Tarjan's algorithm for SCCs is usually faster than the two-pass approach (Finishing order + reversed graph) used here.

> The main disadvantage of Tarjan's algorithm is the use of recursion.
>
> The previous implementation used an iterative DFS for the finishing order (`depthFirstSearchOrderingByFinishTime`), which is safer against `StackOverflowError` on extremely deep or linear graphs. Tarjan's algorithm, as implemented here, is recursive; if the dependency graph is exceptionally deep (thousands of components in a single chain), it could potentially trigger a stack overflow.
>
> However, for most software dependency graphs, the depth is rarely sufficient to cause this, and the performance gains from a single pass and reduced memory footprint (no reversed graph) typically outweigh this risk.

I've run [the analysis](https://github.com/TNG/ArchUnit/pull/1629#pullrequestreview-4356980071) of [`hibernate-core:7.3.6`](https://github.com/hibernate/hibernate-orm/tree/7.3.6/hibernate-core) against an AI-implementation of Tarjan's algorithm, and arrived at **(5.9±0.4) s**, compared to **(6.2±0.3) s** for your solution.

_**What do you think?**_
(I've pushed that suggestion to [`optimize-lakos-metrics-using-Tarjans-algorithm-to-find-SCCs`](https://github.com/TNG/ArchUnit/commit/047c83d244365ccb50277c3a792cd477d57ba69b).)

archunit/src/main/java/com/tngtech/archunit/library/metrics/MetricsComponentDependencyGraph.java
ThanosTsiamis: Fixed as suggested

archunit/src/main/java/com/tngtech/archunit/library/metrics/MetricsComponentDependencyGraph.java
ThanosTsiamis: Thank you so much for taking the time to review this and for running those benchmarks, @hankem ! Seeing the execution time for hibernate-core drop from ~101 seconds down to 6.2 seconds is an amazing result. 🎉

To be completely transparent, the primary reason I initially went with Kosaraju's algorithm is that it was the simplest one to implement safely. Given how severe the performance bottleneck was in the original solution, Kosaraju felt like the perfect "low-hanging fruit" to grab first, delivering an immediate optimization without overcomplicating the codebase right out of the gate.


### Looking Ahead
Honestly, looking ahead, it feels like it's finally time to dedust some useless knowledge I never thought I would actually implement outside of random uni courses and Google interview prep questions! Choosing an alternative SCC algorithm down the road ultimately boils down to which specific constraints ArchUnit wants to prioritize as the library evolves:

| Algorithm | Strengths | Trade-off / Fit for ArchUnit |
| :--- | :--- | :--- |
| **Kosaraju's** (Current) | Highly maintainable, robust, easily made iterative. | Requires a transpose graph array. |
| **Tarjan's** | Industry-standard single-pass optimization. | High bookkeeping complexity if written iteratively. |
| **Gabow's** | Elegant, stack-based single-pass. No integer low-link math. | Can completely eliminate the transpose graph, saving RAM. |
| **Pearce's** | Extreme space-optimization. | Strips away tracking arrays to minimize memory footprint. |
| **Fleischer's** | Massively parallelizable (uses BFS instead of DFS). | Only beneficial if ArchUnit moves toward multi-threaded graph evaluation. |

### Personal Opinion

Since the current Kosaraju implementation already delivers a massive 95% of the performance gains while guaranteeing absolute stack safety and code readability, I believe it makes sense to merge this as a baseline win.

If we want to push the boundaries further, we could absolutely look into parameterizing MetricsComponentDependencyGraph to support interchangeable graph engines, or explore a single-pass iterative alternative like Gabow's (to drop the transpose graph entirely) in a future PR.

Cheers! 🎉

hankem: Let's just keep your implementation. The additional performance gain of that implementation of Tarjan's algorithm is hardly detectable – and anyway irrelevant compared to the boost you already provided.

I was just curious about your expert opinion. Thank you! 🙂

archunit/src/main/java/com/tngtech/archunit/library/metrics/MetricsComponentDependencyGraph.java
hankem: Thank you! If you like, you could maybe squash the commit with the imports into the other?
Otherwise I'd say it's good to be merged.

ThanosTsiamis: Single commit policy enforced! 
To follow the usual programming standards, I’ll leave it to you to resolve the comments as you see fit since you're the original author. 
Please note that Github says that" _merging is currently blocked and direct pushes to this branch are restricted._" so I guess you also need to enable the pipelines
Cheers! =]

PR #1636, opened 2026-06-15
Separate integration tests by JUnit version

schulzjo-tng:
This is in preparation for adding JUnit 6 support. Since JUnit 5 and JUnit 6 have a lot of overlap in their classes, it is not possible (or, at the very least, not easy or clean) to integration test both JUnit 5 and JUnit 6 from the same Gradle subproject.

I split the shared configuration because the explicit dependencies on JUnit 5 would leak into the integration test classpaths.

I decided against removing all JUnit 5 code from the JUnit 4 integration test since I considered that to be too much effort for a JUnit version that has been deprecated for quite a while now.


archunit-integration-test/shared/src/main/java/com/tngtech/archunit/testutils/ExpectedTestFailures.java
hankem, line 1: I'd like to avoid duplicating this code that's largely unrelated to the test framework.
I've just pushed a suggestion to branch `feature/restructure-integration-tests.review` (d0644dfa69f0ceaa72d17bad5263742b0ecdf665).

You probably don't want the `junit5Jupiter` dependency in `archunit-integration-test:shared`?
I also have an idea how to push that down into `ExpectedJUnit4TestFailures` & `ExpectedJUnit5TestFailures`, if needed.

hankem: @schulzjo-tng [wrote](https://github.com/TNG/ArchUnit/commit/d0644dfa69f0ceaa72d17bad5263742b0ecdf665#r189946071):
> I am not happy with [the `junit5Jupiter` dependency in `archunit-integration-test/shared`], since both the JUnit5 and JUnit 6 integration tests will have to import the shared module, so there is some risk of mixing JUnit5 and JUnit 6.

I've updated [branch `feature/restructure-integration-tests.review`](https://github.com/TNG/ArchUnit/commits/feature/restructure-integration-tests.review/) and implemented the idea I mentioned before in 4fa676fa31bee038dc80c59b61a88865df3ac228.

What do you think?

schulzjo-tng: Looks good to me, feel free to push (edit: done by myself)

archunit-integration-test/shared/src/main/java/com/tngtech/archunit/integration/ExamplesIntegrationTestBase.java
hankem, line 1: This is a huge code duplication with the JUnit4 version.
I'll push another commit to move the logic of the tests to `archunit-integration-test.shared`.

archunit-integration-test/junit5/src/test/java/com/tngtech/archunit/ArchUnitArchitectureTest.java
hankem, line 1: Since these tests are meant to test ArchUnit itself, there's no need to run them redundantly with JUnit4 and JUnit5. I'll push a commit to move them to some other module.

archunit-integration-test/archunit-test/build.gradle
```diff
@@ -0,0 +1,22 @@
...
+ext.moduleName = 'com.tngtech.archunit.integrationtest.archunit-test'
+description = "ArchUnit tests for ArchUnit and ArchUnit-Examples themselves"
+
+ext.minimumJavaVersion = JavaVersion.VERSION_1_8
+
+dependencies {
+    implementation project(path: ':archunit')
+    implementation project(path: ':archunit-junit4')
+    implementation project(path: ':archunit-junit5')
+    implementation project(path: ':archunit-example:example-plain')
```
schulzjo-tng: Is it intended that these tests no longer run on the junit4 and junit5 examples, but only on the plain examples? For example, introducing
```
import com.google.common.base.Joiner;
// ...
    static Joiner joiner = Joiner.on("a");
```
in `archunit-example/example-junit5/src/test/java/com/tngtech/archunit/exampletest/junit5/ControllerRulesTest.java` causes a test failure of `archunit-integration-test/src/test/java/com/tngtech/archunit/ArchUnitExampleArchitectureTest.java` on `main` but not on this branch.

This can be fixed by adding
```
    implementation project(path: ':archunit-example:example-plain', configuration: 'tests')
    implementation project(path: ':archunit-example:example-junit4', configuration: 'tests')
    implementation project(path: ':archunit-example:example-junit5', configuration: 'tests')
```
here. I am not sure if that comes with any side effects.

archunit-integration-test/archunit-test/build.gradle
```diff
@@ -0,0 +1,22 @@
...
+
+ext.moduleName = 'com.tngtech.archunit.integrationtest.archunit-test'
+description = "ArchUnit tests for ArchUnit and ArchUnit-Examples themselves"
+
+ext.minimumJavaVersion = JavaVersion.VERSION_1_8
+
+dependencies {
+    implementation project(path: ':archunit')
+    implementation project(path: ':archunit-junit4')
+    implementation project(path: ':archunit-junit5')
```
schulzjo-tng: Is there a potential issue in adding `:archunit-junit6` here in the future? We only every compile this mix of functions and run ArchUnit on it, so runtime incompatibilities aren't an issue here. However, I am not sure how the collisions between ArchUnit 5 and 6 would be resolved.

A different approach could be that we create a shared subproject which provides the self-tests, then we run them as part of the test suite in every other subproject that is supposed to be tested. Not sure about the pros and cons of each approach, looking forward to your opinion.

archunit-integration-test/archunit-test/build.gradle
```diff
@@ -0,0 +1,22 @@
+plugins {
+    id 'archunit.java-conventions'
+}
+
+ext.moduleName = 'com.tngtech.archunit.integrationtest.archunit-test'
```
schulzjo-tng: Maybe `com.tngtech.archunit.integrationtest.archunit-self-test`? Would affect the name of the directory as well. Also, does this even belong under `archunit-integration-test`?

archunit-integration-test/archunit-test/build.gradle
```diff
@@ -0,0 +1,22 @@
...
+    implementation project(path: ':archunit-junit4')
+    implementation project(path: ':archunit-junit5')
+    implementation project(path: ':archunit-example:example-plain')
+
+    testImplementation libs.junit5Jupiter
+    testImplementation libs.junit5Platform
+    testImplementation libs.assertj
+    testImplementation project(path: ':archunit-junit5-api')
+    testImplementation project(path: ':archunit-junit5-engine-api')
+    testRuntimeOnly project(path: ':archunit-junit5-engine')
```
schulzjo-tng: I like the approach of running the self-tests on the latest JUnit version only. Maybe add a comment like the following?
```
// The ArchUnit self-tests are not part of the integration tests and therefore only need to run on one version of JUnit
```

archunit-integration-test/archunit-test/build.gradle
hankem: Good question! I don't know; I'd just see whether it works – we can still split the module if it doesn't.

hankem (approved):

archunit-integration-test/archunit-test/build.gradle
hankem: I've moved it outside of `archunit-integration-tests`.

archunit-self-test/build.gradle
```diff
@@ -12,6 +12,7 @@ dependencies {
     implementation project(path: ':archunit')
     implementation project(path: ':archunit-junit4')
     implementation project(path: ':archunit-junit5')
+    implementation project(path: ':archunit-junit5-engine')
```
hankem, line 15: Good catch; thank you @schulzjo-tng!

schulzjo-tng: The actual catch was https://github.com/TNG/ArchUnit/pull/1636/changes/8d862cde3bf57338a5f4cba05a2bbd425dc9eb52#diff-183b694ab58372ebbaa315985338f3b14398dc85114c97a62f6870af41e16144R119, this was just required to get the change above to compile

PR #1648, opened 2026-06-28
Stabilize ModuleDependency descriptions

### LINKED ISSUE #1630: Nondeterministic violation description for ModuleDependency: leads to failing test when used as frozen rule
We've noticed that when declaring a rule about allowed module dependencies, like so:

`modules().definedByPackages("..root.(*)..").should().respectTheirAllowedDependencies(allow().fromModule("main").toModules("util"))`

If there are multiple violations of this rule, the generated violation message has non-deterministic ordering.

Example violation message:

> Module Dependency [main -> util]:
> Method <util.UtilClass.foo()> calls method <main.SomeMainClass.bar()> in (UtilClass.java:12)
> Method <othermodule.OtherClass.baz()> calls method <util.UtilClass.foo()> in (OtherClass.java:34)

Although it is functionally correct (the list includes all offending places), the fact that it may give a different ordering every time leads to problems when using the frozen rule mechanism, which checks against already known violations. Because the violation text isn't completely equal, it is seen as a new violation and therefore fails the test.

**Problem:**
The `getDescription()` method of the `ModuleDependency` class iterates over the `classDependencies` field, which is a Set. Although that set should preserve its insertion order, it seems that the order in which it gets populated can differ with each ArchUnit analysis run.

**Solution alternatives:**
1. Make sure that the set of class dependencies that get inserted here always have deterministic ordering. Seems error prone / difficult to achieve: how will we ensure we have catched all places from which this could be populated?
2. Make sure that, regardless of insertion order, the `getDescription()` method always produces deterministic output, for instance by putting a `.sorted()` operation into its stream. Seems like a small enough fix and even leads to better readable violation messages.

DragonFSKY:
Fixes #1630.

This canonicalizes the emitted class-dependency descriptions lexicographically before joining them, so the same dependency set produces stable `ModuleDependency` report text independently of the backing set encounter order. The ordering is defined at the textual representation boundary and does not depend on `Dependency`'s line-number-first natural order.

The regression test imports real classes, forces opposite dependency encounter orders, and verifies that `Dependency` natural order conflicts with description order. It therefore fails both without sorting and with dependency-object sorting, while passing with description-string sorting.

Tested with:

```text
./gradlew :archunit:clean :archunit:test --tests com.tngtech.archunit.library.modules.ModuleDependencyTest
./gradlew :archunit:spotlessCheck
./gradlew :archunit:check
```

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/library/modules/ModuleDependency.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/modules/ModuleDependency.java
@@ -85,6 +85,7 @@ public final class ModuleDependency<DESCRIPTOR extends ArchModule.Descriptor> im
     public String getDescription() {
         String classDependencyDescriptions = classDependencies.stream()
                 .map(HasDescription::getDescription)
+                .sorted()
                 .collect(joining(lineSeparator()));
         return String.format("Module Dependency [%s -> %s]:%n%s", origin.getName(), target.getName(), classDependencyDescriptions);
     }
--- a/archunit/src/test/java/com/tngtech/archunit/library/modules/ModuleDependencyTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/library/modules/ModuleDependencyTest.java
@@ -1,12 +1,20 @@
 package com.tngtech.archunit.library.modules;
 
+import java.util.LinkedHashSet;
+import java.util.Set;
+
 import com.tngtech.archunit.core.domain.Dependency;
+import com.tngtech.archunit.core.domain.JavaClass;
+import com.tngtech.archunit.core.domain.JavaClasses;
 import com.tngtech.archunit.core.importer.ClassFileImporter;
 import com.tngtech.archunit.library.modules.syntax.testexamples.test_modules.one.ClassOne;
 import com.tngtech.archunit.library.modules.syntax.testexamples.test_modules.two.ClassTwo;
 import org.junit.Test;
 
 import static com.tngtech.archunit.testutil.Assertions.assertThatConversionOf;
+import static java.util.Arrays.asList;
+import static java.util.Collections.singleton;
+import static org.assertj.core.api.Assertions.assertThat;
 
 public class ModuleDependencyTest {
     @Test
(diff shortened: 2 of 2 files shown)


hankem:
Thank you so much for your contribution, and first of all sorry for the late feedback!

The only thing I'd like to test is how much sorting affects the performance, but other than that, I have no concerns that we can integrate your fix _soon._

DragonFSKY:
Thanks for taking a look, and no worries about the timing! I fixed the DCO issue and normalized both commits to the same GitHub identity/sign-off.

I also ran a targeted local benchmark on JDK 25 using real ModuleDependency instances built from ArchUnit classes. It compared the previous unsorted description rendering with the sorted rendering in this PR; the figures below are medians from seven alternating rounds:

| Class dependencies | Unsorted | Sorted | Added time |
| ---: | ---: | ---: | ---: |
| 81 | 17.3 us | 22.0 us | 4.7 us |
| 926 | 126.4 us | 455.7 us | 329.3 us |
| 5,449 | 1.35 ms | 3.41 ms | 2.05 ms |

So the relative cost is visible for large dependency sets, but the absolute overhead stayed around 2 ms even for the largest real module dependency in this sample, and it is limited to the getDescription/reporting path. I hope this helps with the performance assessment.

archunit/src/main/java/com/tngtech/archunit/library/modules/ModuleDependency.java
```diff
@@ -84,6 +84,7 @@ public Set<Dependency> toClassDependencies() {
     @PublicAPI(usage = ACCESS, state = EXPERIMENTAL)
     public String getDescription() {
         String classDependencyDescriptions = classDependencies.stream()
+                .sorted()
                 .map(HasDescription::getDescription)
```
hankem: Thanks for your performance measurements already!

As expected, sorting adds some time. 🤷‍♂️

In an even more extreme test case of
```java
JavaClasses classes = new ClassFileImporter().importPackages("org.hibernate");
ArchModules<?> modules = ArchModules.defineByPackages("org.hibernate.(*)..").modularize(classes);
long t0 = System.nanoTime();
modules.stream()
	.flatMap(m -> m.getModuleDependenciesFromSelf().stream())
	.forEach(md -> md.getDescription());  // this is where .sorted() now runs
long duration = System.nanoTime() - t0;
```
on [hibernate-core 7.4.5.Final](https://central.sonatype.com/artifact/org.hibernate/hibernate-core/7.4.5.Final) (where one `ModuleDependency` consists of 11 418 `classDependencies`),

I measured `duration` to increase from **30±4 ms** to **114±6 ms**.

We could, however, easily save some time by not sorting `Dependency` objects, but just their description `String`s:

```suggestion
                .map(HasDescription::getDescription)
                .sorted()
```
In my case that at least brings us down to **88±6 ms**. 

What do you think?

DragonFSKY: Thanks for the suggestion and the measurements! I’ve updated the PR to sort the rendered descriptions and added a regression test for stable ordering. The checks pass locally.

hankem (approved):
Thanks for the contribution and the test that clearly demonstrates which order is considered!

PR #1653, opened 2026-07-06
Replace junit-dataprovider with junit-jupiter-params

hankem:
Resolves #1471


hankem:
This PR was accidentally merged, but has been removed from `main` again and recreated as PR #1655.

PR #1655, opened 2026-07-06
Replace junit-dataprovider with junit-jupiter-params

hankem:
Resolves #1471

### DIFF
--- a/archunit/src/jdk16test/java/com/tngtech/archunit/lang/syntax/elements/ClassesShouldNewerJavaVersionTest.java
+++ b/archunit/src/jdk16test/java/com/tngtech/archunit/lang/syntax/elements/ClassesShouldNewerJavaVersionTest.java
@@ -1,37 +1,34 @@
 package com.tngtech.archunit.lang.syntax.elements;
 
+import java.util.stream.Stream;
+
 import com.tngtech.archunit.lang.ArchRule;
 import com.tngtech.archunit.lang.EvaluationResult;
 import com.tngtech.archunit.lang.conditions.ArchConditions;
 import com.tngtech.archunit.lang.syntax.elements.testobjects.SomeRecord;
-import com.tngtech.java.junit.dataprovider.DataProvider;
-import com.tngtech.java.junit.dataprovider.DataProviderRunner;
-import com.tngtech.java.junit.dataprovider.UseDataProvider;
-import org.junit.Test;
-import org.junit.runner.RunWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
 import static com.tngtech.archunit.core.domain.TestUtils.importClasses;
 import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.classes;
 import static com.tngtech.archunit.lang.syntax.elements.ClassesShouldTest.locationPattern;
 import static com.tngtech.archunit.lang.syntax.elements.ClassesShouldTest.singleLineFailureReportOf;
 import static com.tngtech.archunit.testutil.Assertions.assertThat;
-import static com.tngtech.java.junit.dataprovider.DataProviders.$;
-import static com.tngtech.java.junit.dataprovider.DataProviders.$$;
+import static com.tngtech.archunit.testutil.DataProviders.$;
 import static java.util.regex.Pattern.quote;
 
-@RunWith(DataProviderRunner.class)
 public class ClassesShouldNewerJavaVersionTest {
 
-    @DataProvider
-    public static Object[][] data_beRecords() {
-        return $$(
+    static Stream<Arguments> be_records() {
+        return Stream.of(
                 $(classes().should().beRecords(), SomeRecord.class, String.class),
                 $(classes().should(ArchConditions.beRecords()), SomeRecord.class, String.class));
     }
 
-    @Test
-    @UseDataProvider
-    public void test_beRecords(ArchRule rule, Class<?> satisfied, Class<?> violated) {
+    @ParameterizedTest
+    @MethodSource
+    void be_records(ArchRule rule, Class<?> satisfied, Class<?> violated) {
         EvaluationResult result = rule.evaluate(importClasses(satisfied, violated));
 
         assertThat(singleLineFailureReportOf(result))
@@ -42,16 +39,15 @@ public class ClassesShouldNewerJavaVersionTest {
                 .doesNotMatch(String.format(".*%s.* record.*", quote(satisfied.getName())));
     }
 
-    @DataProvider
-    public static Object[][] data_notBeRecords() {
-        return $$(
+    static Stream<Arguments> not_be_records() {
+        return Stream.of(
                 $(classes().should().notBeRecords(), String.class, SomeRecord.class),
                 $(classes().should(ArchConditions.notBeRecords()), String.class, SomeRecord.class));
     }
 
-    @Test
-    @UseDataProvider
-    public void test_notBeRecords(ArchRule rule, Class<?> satisfied, Class<?> violated) {
+    @ParameterizedTest
+    @MethodSource
+    void not_be_records(ArchRule rule, Class<?> satisfied, Class<?> violated) {
         EvaluationResult result = rule.evaluate(importClasses(satisfied, violated));
 
         assertThat(singleLineFailureReportOf(result))
--- a/archunit/src/jdk16test/java/com/tngtech/archunit/lang/syntax/elements/ClassesShouldNewerJavaVersionTest.java
+++ b/archunit/src/jdk16test/java/com/tngtech/archunit/lang/syntax/elements/ClassesShouldNewerJavaVersionTest.java
@@ -15,15 +15,15 @@ import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.classes;
 import static com.tngtech.archunit.lang.syntax.elements.ClassesShouldTest.locationPattern;
 import static com.tngtech.archunit.lang.syntax.elements.ClassesShouldTest.singleLineFailureReportOf;
 import static com.tngtech.archunit.testutil.Assertions.assertThat;
-import static com.tngtech.archunit.testutil.DataProviders.$;
 import static java.util.regex.Pattern.quote;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
 
 public class ClassesShouldNewerJavaVersionTest {
 
     static Stream<Arguments> be_records() {
         return Stream.of(
-                $(classes().should().beRecords(), SomeRecord.class, String.class),
-                $(classes().should(ArchConditions.beRecords()), SomeRecord.class, String.class));
+                arguments(classes().should().beRecords(), SomeRecord.class, String.class),
+                arguments(classes().should(ArchConditions.beRecords()), SomeRecord.class, String.class));
     }
 
     @ParameterizedTest
@@ -41,8 +41,8 @@ public class ClassesShouldNewerJavaVersionTest {
 
     static Stream<Arguments> not_be_records() {
         return Stream.of(
-                $(classes().should().notBeRecords(), String.class, SomeRecord.class),
-                $(classes().should(ArchConditions.notBeRecords()), String.class, SomeRecord.class));
+                arguments(classes().should().notBeRecords(), String.class, SomeRecord.class),
+                arguments(classes().should(ArchConditions.notBeRecords()), String.class, SomeRecord.class));
     }
 
     @ParameterizedTest
--- a/archunit/src/test/java/com/tngtech/archunit/core/importer/ImportOptionsTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/core/importer/ImportOptionsTest.java
@@ -29,10 +29,10 @@ import static com.tngtech.archunit.core.importer.ImportOption.Predefined.DO_NOT_
 import static com.tngtech.archunit.core.importer.ImportOption.Predefined.DO_NOT_INCLUDE_TESTS;
 import static com.tngtech.archunit.core.importer.ImportOption.Predefined.DO_NOT_INCLUDE_TEST_FIXTURES;
 import static com.tngtech.archunit.core.importer.ImportOption.Predefined.ONLY_INCLUDE_TESTS;
-import static com.tngtech.archunit.testutil.DataProviders.$;
 import static com.tngtech.archunit.testutil.TestUtils.relativeResourceUri;
 import static java.util.Arrays.stream;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
 
 public class ImportOptionsTest {
     @TempDir
@@ -95,8 +95,8 @@ public class ImportOptionsTest {
         List<Arguments> includeMainFolderInput = new ArrayList<>();
         List<Arguments> includeTestFolderInput = new ArrayList<>();
         for (FolderPattern folderPattern : getFolderPatterns()) {
-            includeMainFolderInput.add($(folderPattern.folders, folderPattern.isMainFolder));
-            includeTestFolderInput.add($(folderPattern.folders, !folderPattern.isMainFolder));
+            includeMainFolderInput.add(arguments(folderPattern.folders, folderPattern.isMainFolder));
+            includeTestFolderInput.add(arguments(folderPattern.folders, !folderPattern.isMainFolder));
         }
 
         Stream<Arguments> doNotIncludeTestsDataPoints = crossProduct(do_not_include_tests(), includeMainFolderInput);
--- a/archunit/src/test/java/com/tngtech/archunit/core/importer/ImportOptionsTest.java
+++ b/archunit/src/test/java/com/tngtech/archunit/core/importer/ImportOptionsTest.java
@@ -1,29 +1,27 @@
 package com.tngtech.archunit.core.importer;
 
-import java.io.File;
 import java.io.IOException;
 import java.net.URI;
-import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
 import com.tngtech.archunit.core.importer.ImportOption.DoNotIncludeArchives;
+import com.tngtech.archunit.core.importer.ImportOption.DoNotIncludeGradleTestFixtures;
 import com.tngtech.archunit.core.importer.ImportOption.DoNotIncludeJars;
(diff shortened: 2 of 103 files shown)


archunit/src/test/java/com/tngtech/archunit/library/freeze/ViolationLineMatcherFactoryTest.java
```diff
@@ -1,19 +1,16 @@
...
+import org.junit.jupiter.params.provider.CsvSource;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-@RunWith(DataProviderRunner.class)
 public class ViolationLineMatcherFactoryTest {
-    @Test
-    @DataProvider(splitBy = "\\|", value = {
+    @ParameterizedTest
+    @CsvSource(delimiter = '|', value = {
```
StefanGraeber: I find this really hard to read.
I pushed you a commit on top of the chain which migrates this to a MethodSource

archunit/src/jdk16test/java/com/tngtech/archunit/lang/syntax/elements/ClassesShouldNewerJavaVersionTest.java
```diff
@@ -1,37 +1,34 @@
...
 
-@RunWith(DataProviderRunner.class)
 public class ClassesShouldNewerJavaVersionTest {
 
-    @DataProvider
-    public static Object[][] data_beRecords() {
-        return $$(
+    static Stream<Arguments> be_records() {
+        return Stream.of(
                 $(classes().should().beRecords(), SomeRecord.class, String.class),
```
StefanGraeber: let's get rid of the `$` syntax and switch to standard junit 5 methods.
You already made this an alias for `Arguments.of` which is an alias for `Arguments.arguments` which is realy nice to read with a static import.
The is also `Arguments.argumentSet` which has an additional parameter in the front which is used as name of the text.
Since I'm a big fan of the latter one, I would use it in all places which benefit from a custom test name (I haven't spotted one currently using a custom test name)
I pushed you a commit which replaces the `$` method with `arguments`.
If you like it, keep or squash it into yours, otherwise you can also drop it (and please tell me what and why you dislike)

archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/ClassesShouldConjunctionTest.java
```diff
@@ -19,27 +17,23 @@
...
 import static java.util.regex.Pattern.quote;
 import static org.assertj.core.api.Assertions.assertThat;
 
-@RunWith(DataProviderRunner.class)
 public class ClassesShouldConjunctionTest {
-    @DataProvider
-    public static Object[][] ORed_conditions() {
-        return $$(
-                $(classes()
+    static Stream<ArchRule> ORed_conditions() {
```
StefanGraeber, line 24: was it intentional to no wrap into `Arguments`?
I didn't know that that wrapping is optional and always used it so far

archunit/src/test/java/com/tngtech/archunit/core/importer/ImportOptionsTest.java
```diff
@@ -98,42 +92,44 @@ private static List<FolderPattern> getFolderPatterns() {
         );
     }
 
-    @DataProvider
-    public static Object[][] test_location_predicates_and_expected_folder_patterns() {
-        List<Object[]> includeMainFolderInput = new ArrayList<>();
-        List<Object[]> includeTestFolderInput = new ArrayList<>();
+    static Stream<Arguments> test_location_predicates_and_expected_folder_patterns() {
+        Set<Arguments> includeMainFolderInput = new HashSet<>();
```
StefanGraeber: using any object in a `Set` requires proper implementations of hashCode and equals.
since `Arguments` is implemented as functional interface, I think it has just the default implementation of it.
I think it is safer to stay with `List<Arguments>`. this also gives a deterministic test case order, which I consider desirable (we also had it before)

archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/ClassesShouldTest.java
```diff
@@ -72,31 +72,27 @@
...
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
-@RunWith(DataProviderRunner.class)
 public class ClassesShouldTest {
     static final String FAILURE_REPORT_NEWLINE_MARKER = "#";
 
-    @Rule
-    public final ArchConfigurationRule configurationRule = new ArchConfigurationRule();
+    @RegisterExtension
+    ArchConfigurationExtension archConfiguration = new ArchConfigurationExtension();
```
StefanGraeber, line 84: optional: include in the commit message the migration ArchConfigurationRule -> ArchConfigurationExtension
I know that it must be done because of junit4 ->junit5 but I think it wouldn't hurt to make it explicit

you even created that Extension in this commit! 🙂 
I think that's worth a mention

archunit/src/test/java/com/tngtech/archunit/testutil/ArchConfigurationRule.java
```diff
@@ -1,48 +1,38 @@
...
     }
 
     @Override
     protected void after() {
-        ArchConfiguration.get().reset();
+        ExtensionContext unusedContext = null; // good enough for now, and ArchConfigurationRule is deprecated anyways
+        configuration.afterEach(unusedContext);
     }
 
     public static <T> T resetConfigurationAround(Callable<T> callable) {
```
StefanGraeber, line 38: what about this method?
Do we not need it in the new Extension?
to be a full replacement, I'd expect to have that method in the extension as well better: move it

archunit/src/test/java/com/tngtech/archunit/testutil/ArchConfigurationRule.java
```diff
@@ -1,48 +1,38 @@
...
-import java.util.ArrayList;
-import java.util.List;
 import java.util.concurrent.Callable;
 
 import com.tngtech.archunit.ArchConfiguration;
+import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.rules.ExternalResource;
 
+/**
+ * @deprecated use JUnit 5 and {@link ArchConfigurationExtension} instead
```
StefanGraeber, line 10: Is it feasible to entirely remove this Rule?
I might give it a try later, there seem to be only 25 tests left

archunit/src/test/java/com/tngtech/archunit/core/importer/resolvers/ClassResolverFactoryTest.java
```diff
@@ -12,23 +12,23 @@
...
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 public class ClassResolverFactoryTest {
-    @Rule
-    public final ArchConfigurationRule archConfigurationRule = new ArchConfigurationRule();
-    @Rule
-    public final OutsideOfClassPathRule outsideOfClassPathRule = new OutsideOfClassPathRule();
-    @Rule
-    public final ContextClassLoaderRule contextClassLoaderRule = new ContextClassLoaderRule();
+    @RegisterExtension
```
StefanGraeber: 2 of these extensions are never accessed as fields in this class (and all are created with default constructor).
any reason why they are not used via `@RegisterExtension({ArchConfigurationExtension.class, ContextClassLoaderExtension.class})`?

archunit/src/test/java/com/tngtech/archunit/core/importer/LocationsTest.java
```diff
@@ -56,7 +56,7 @@ public void locations_of_packages_within_JAR_URIs() throws Exception {
         Set<Location> locations = Locations.ofPackage("org.junit");
```
StefanGraeber, line 55: if you replace this string by `org.junit.jupiter.api` you can replace `org.junit.Test.class` by `Test.class`
I see no reason to use the junit4 variant of it. 
let's break that dependency (there are 3 more usages in this class of the class literal)

archunit/src/test/java/com/tngtech/archunit/testutil/ContextClassLoaderExtension.java
```diff
@@ -0,0 +1,19 @@
+package com.tngtech.archunit.testutil;
+
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+public class ContextClassLoaderExtension implements BeforeEachCallback, AfterEachCallback {
```
StefanGraeber, line 7: please deprecate `ContextClassLoaderRule`

archunit/src/test/java/com/tngtech/archunit/testutil/SystemPropertiesExtension.java
```diff
@@ -0,0 +1,21 @@
+package com.tngtech.archunit.testutil;
+
+import java.util.Properties;
+
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+public class SystemPropertiesExtension implements BeforeEachCallback, AfterEachCallback {
```
StefanGraeber, line 9: please deprecate `SystemPropertiesRule`

StefanGraeber (changes requested):

archunit/src/test/java/com/tngtech/archunit/library/freeze/ViolationLineMatcherFactoryTest.java
hankem: Hm, I did find it easier to spot the differences 
<details><summary>when the corresponding strings are right below each other</summary>

```
            "" +
                    "''|" +
                    "''|" + true,
            "" +
                    "abc|" +
                    "abc|" + true,
            "" +
                    "abc|" +
                    "abcd|" + false,
            "" +
                    "(A.java:1)|" +
                    "(A.java:2)|" + true,
            "" +
                    "A.java:1|" +
                    "A.java:2|" + false,
            "" +
                    "(A.java:1)|" +
                    "(A.java:2|" + false,
            "" +
                    "A$1 B$2 C$4 (X.java:111)|" +
                    "A$2 B$3 C$5 (X.java:222)|" + true,
            "" +
                    "A$a|" +
                    "A$b|" + false,
            "" +
                    "A:1|" +
                    "A$2|" + false,
            "" +
                    "Method <MyClass.someSyntheticStuff$2()> has a violation in (MyClass.java:123)|" +
                    "Method <MyClass.someSyntheticStuff$123()> has a violation in (MyClass.java:0)|" + true,
            "" +
                    "Method <C.someSyntheticStuff$200()> is bad in (C.java:123)|" +
                    "Method <C.someSyntheticStuff$200()> is bad in (C.java:123), too|" + false,
```

</details>

as opposed to 
<details><summary>when they are on the same line.</summary>

```
                arguments("", "", expectMatch),
                arguments("abc", "abc", expectMatch),
                arguments("abc", "abcd", expectNoMatch),
                arguments("(A.java:1)", "(A.java:2)", expectMatch),
                arguments("A.java:1", "A.java:2", expectNoMatch),
                arguments("(A.java:1)", "(A.java:2", expectNoMatch),
                arguments("A$1 B$2 C$4 (X.java:111)", "A$2 B$3 C$5 (X.java:222)", expectMatch),
                arguments("A$a", "A$b", expectNoMatch),
                arguments("A:1", "A$2", expectNoMatch),
                arguments("Method <MyClass.someSyntheticStuff$2()> has a violation in (MyClass.java:123)", "Method <MyClass.someSyntheticStuff$123()> has a violation in (MyClass.java:0)", expectMatch),
                arguments("Method <C.someSyntheticStuff$200()> is bad in (C.java:123)", "Method <C.someSyntheticStuff$200()> is bad in (C.java:123), too", expectNoMatch),
```

</details>

Of course the syntax for the `@CsvSource` is a bit crazy, so I like the migration towards `@MethodSource`.
But maybe we could still

<details><summary>keep the string to be compared on two lines.</summary>

```
                expectMatch(
                        "",
                        ""
                ),
                expectMatch(
                        "abc",
                        "abc"
                ),
                expectNoMatch(
                        "abc",
                        "abcd"
                ),
                expectMatch(
                        "(A.java:1)",
                        "(A.java:2)"
                ),
                expectNoMatch(
                        "A.java:1",
                        "A.java:2"
                ),
                expectNoMatch(
                        "(A.java:1)",
                        "(A.java:2"
                ),
                expectMatch(
                        "A$1 B$2 C$4 (X.java:111)",
                        "A$2 B$3 C$5 (X.java:222)"
                ),
                expectNoMatch(
                        "A$a",
                        "A$b"
                ),
                expectNoMatch(
                        "A:1",
                        "A$2"
                ),
                expectMatch(
                        "Method <MyClass.someSyntheticStuff$2()> has a violation in (MyClass.java:123)",
                        "Method <MyClass.someSyntheticStuff$123()> has a violation in (MyClass.java:0)"
                ),
                expectNoMatch(
                        "Method <C.someSyntheticStuff$200()> is bad in (C.java:123)",
                        "Method <C.someSyntheticStuff$200()> is bad in (C.java:123), too"
                ),
```
using
```
    private static Arguments expectMatch(String str1, String str2) {
        return arguments(str1, str2, true);
    }

    private static Arguments expectNoMatch(String str1, String str2) {
        return arguments(str1, str2, false);
    }
```

</details>

What do you think?

archunit/src/jdk16test/java/com/tngtech/archunit/lang/syntax/elements/ClassesShouldNewerJavaVersionTest.java
hankem: Fine with me! I had only introduced the `$` method as described in the commit message of 
`migrate some junit-dataprovider tests to junit-jupiter-params' @MethodSource`:
> For convenience (and to keep the diff smaller)
> `com.tngtech.archunit.testutil.DataProviders.$` provides the same API for `junit-jupiter-params`
> as `com.tngtech.java.junit.dataprovider.DataProviders.$` did for `junit-dataprovider`.

archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/ClassesShouldConjunctionTest.java
hankem: I just prefer to keep it simple & stupid. 💋
We only need `Arguments` to wrap multiple arguments or add a description,
and returning `Stream<TheActualTypeOfInterest>` (when possible) is even more type-safe.

archunit/src/test/java/com/tngtech/archunit/core/importer/ImportOptionsTest.java
hankem: Since we have to reimplement `com.tngtech.java.junit.dataprovider.DataProviders.crossProduct`,
I was quite happy to be able to build on `com.google.common.collect.Sets.cartesianProduct`.
I'd hope that the default implementation of `Object.hashCode` and `Object.equals` is good enough for us.

hankem: Haha, nevermind: There is `com.google.common.collect.Lists.cartesianProduct` as well! 🙈

archunit/src/test/java/com/tngtech/archunit/testutil/ArchConfigurationRule.java
hankem: I'd just move it there when migrating the tests that use it. It's anyway a static method that could be anywhere.

archunit/src/test/java/com/tngtech/archunit/testutil/ArchConfigurationRule.java
hankem: It is certainly possible.
It would just be unrelated to the goal _Replace `junit-dataprovider` with `junit-jupiter-params`_.

archunit/src/test/java/com/tngtech/archunit/core/importer/resolvers/ClassResolverFactoryTest.java
hankem: Good catch, thank you:
`@ExtendWith({ArchConfigurationExtension.class, ContextClassLoaderExtension.class})` ≡ 💋!

archunit/src/test/java/com/tngtech/archunit/core/importer/LocationsTest.java
hankem: I didn't want the commit `migrate ClassFileImporterTest to junit-jupiter-params` to change any test logic.
Changing `org.junit.Test` to `org.junit.jupiter.api.Test` would perfectly fit in a PR whose goal is to get rid of JUnit 4, but is irrelevant for the goal to eliminate the `junit-dataprovider`.

archunit/src/test/java/com/tngtech/archunit/library/freeze/ViolationLineMatcherFactoryTest.java
StefanGraeber: that sounds like a good addition 🙂

archunit/src/test/java/com/tngtech/archunit/lang/syntax/elements/ClassesShouldConjunctionTest.java
StefanGraeber: "Type-Safe" 😄 
only between different values in the same stream, not with the consumer method, but fine
Since I need more than a single value most of the time (or ValueSource/EnumSource is sufficient), I've never seen that notation in my project

archunit/src/jdk16test/java/com/tngtech/archunit/lang/syntax/elements/ClassesShouldNewerJavaVersionTest.java
StefanGraeber: oops, I missed that part of the commit message.
In the same commit would indeed be a damn large diff.

archunit/src/test/java/com/tngtech/archunit/core/importer/ImportOptionsTest.java
StefanGraeber: > I'd hope that the default implementation of Object.hashCode and Object.equals is good enough for us.

only as long as there is no collision and the uniqueness property of the Set is not relevant.
But since you found a List, this doesn't matter anymore

archunit/src/test/java/com/tngtech/archunit/testutil/ArchConfigurationRule.java
StefanGraeber: then we'll do it in another PR.
maybe one to get rid of junit4 for the tests for archunit entirely. let's see when we have how much time.
There haven't been any Techdays for me for some time

StefanGraeber (approved):
LGTM but you lost a commit gin-off ;)
also don't forget to squash the fixup commits

PR #1656, opened 2026-07-06
Make TextFileBasedViolationStore thread-safe under parallel test execution

### LINKED ISSUE #1654: `TextFileBasedViolationStore` is not thread-safe under parallel test execution
## Summary

When multiple `FreezingArchRule` instances are evaluated in parallel (e.g. via JUnit's parallel test execution), the shared `stored.rules` index file can become corrupted or lose entries, causing subsequent rule evaluations to fail with spurious errors or silently drop previously frozen violations.

## Background

`FreezingArchRule.freeze(ArchRule)` calls `ViolationStoreFactory.create()` on every evaluation, which creates a fresh `TextFileBasedViolationStore` instance. Each instance loads the current content of `stored.rules` into its own private in-memory `Properties` snapshot during `initialize()`. When `save()` is called, the instance writes back only what it knows — its own snapshot — truncating the file and discarding any entries that were written by a concurrently running instance after the snapshot was taken.

## Steps to reproduce

1. Define multiple `FreezingArchRule`s in the same or different test classes.
2. Configure the test runner to execute tests in parallel (e.g. JUnit Platform parallel execution).
3. Run the tests for the first time so violations are frozen.
4. Observe that `stored.rules` contains fewer entries than the number of distinct rules, or that some rules are no longer found in the store on subsequent runs.

## Expected behaviour

All frozen rules are persisted in `stored.rules` regardless of whether their owning `FreezingArchRule` instances run concurrently.

## Actual behaviour

Because each `TextFileBasedViolationStore` instance holds its own snapshot of `stored.rules` and overwrites the file with only that snapshot, the last writer wins and all intermediate writes are lost. With N rules running in parallel, the final `stored.rules` may contain as few as one entry.

Additionally, even within a single shared instance, `FileSyncedProperties.syncFileSystem()` opens a `FileOutputStream` without any synchronization, so two concurrent writes can interleave their truncation and content, producing a corrupted file.

We've observed both behaviors in our real test suite.

## Environment

- ArchUnit version: 1.4.1
- Java version: 25.0.3
- Triggered by: JUnit Jupiter parallel execution

kelunik:
Multiple FreezingArchRule instances each create a fresh TextFileBasedViolationStore and load stored.rules into their own private snapshot. Concurrent saves caused a lost-update: the last writer overwrote the file with only its own entries.

Fix by sharing one FileSyncedProperties per stored.rules canonical path via a static ConcurrentHashMap, and synchronizing writes with a putIfAbsent method that atomically checks, sets, and flushes to disk under one lock.

Fixes #1654.

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/library/freeze/TextFileBasedViolationStore.java
+++ b/archunit/src/main/java/com/tngtech/archunit/library/freeze/TextFileBasedViolationStore.java
@@ -23,6 +23,7 @@ import java.nio.file.Files;
 import java.util.List;
 import java.util.Properties;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Pattern;
 
 import com.google.common.base.Splitter;
@@ -32,7 +33,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.io.Files.toByteArray;
 import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
 import static com.tngtech.archunit.PublicAPI.Usage.INHERITANCE;
@@ -74,6 +74,8 @@ public final class TextFileBasedViolationStore implements ViolationStore {
     private static final String ALLOW_STORE_UPDATE_PROPERTY_NAME = "default.allowStoreUpdate";
     private static final String ALLOW_STORE_UPDATE_DEFAULT = "true";
 
+    private static final ConcurrentHashMap<String, FileSyncedProperties> STORED_RULES_BY_PATH = new ConcurrentHashMap<>();
+
     private final RuleViolationFileNameStrategy ruleViolationFileNameStrategy;
 
     private boolean storeCreationAllowed;
@@ -105,11 +107,18 @@ public final class TextFileBasedViolationStore implements ViolationStore {
         storeUpdateAllowed = Boolean.parseBoolean(properties.getProperty(ALLOW_STORE_UPDATE_PROPERTY_NAME, ALLOW_STORE_UPDATE_DEFAULT));
         String path = properties.getProperty(STORE_PATH_PROPERTY_NAME, STORE_PATH_DEFAULT);
         storeFolder = new File(path);
-        ensureExistence(storeFolder);
         File storedRulesFile = getStoredRulesFile();
         log.trace("Initializing {} at {}", TextFileBasedViolationStore.class.getSimpleName(), storedRulesFile.getAbsolutePath());
-        storedRules = new FileSyncedProperties(storedRulesFile);
+        storedRules = getOrCreateStoredRules(storedRulesFile);
         checkInitialization(storedRules.initializationSuccessful(), "Cannot create rule store at %s", storedRulesFile.getAbsolutePath());
+    }
+
+    private FileSyncedProperties getOrCreateStoredRules(File storedRulesFile) {
+        try {
+            return STORED_RULES_BY_PATH.computeIfAbsent(storedRulesFile.getCanonicalPath(), path -> new FileSyncedProperties(storedRulesFile));
+        } catch (IOException e) {
+            throw new StoreInitializationFailedException(e);
+        }
     }
 
     private File getStoredRulesFile() {
@@ -120,10 +129,6 @@ public final class TextFileBasedViolationStore implements ViolationStore {
                     ViolationStoreFactory.FREEZE_STORE_PROPERTY_NAME, ALLOW_STORE_CREATION_PROPERTY_NAME));
         }
         return rulesFile;
-    }
-
-    private void ensureExistence(File folder) {
-        checkState(folder.exists() && folder.isDirectory() || folder.mkdirs(), "Cannot create folder %s", folder.getAbsolutePath());
     }
 
     private void checkInitialization(boolean initializationSuccessful, String message, Object... args) {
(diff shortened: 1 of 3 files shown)


archunit/src/main/java/com/tngtech/archunit/library/freeze/TextFileBasedViolationStore.java
```diff
@@ -108,10 +111,18 @@ public void initialize(Properties properties) {
         ensureExistence(storeFolder);
```
StefanGraeber: technically, this line needs also synchronization since multiple threads might call `mkdir` at the same time and one fails with the creation since the other created it between the check time and create time.

I first stumbled over having multiple different File objects hanging around, but I just checked that they are only wrappers of the path as string, so it should be safe.

However, I think that it would be a cleaner code flow, if you push the existence check of the directory into `FileSyncedProperties.initializePropertiesFile` as well (next to the existence check and creation of the file) which removes that (very unlikely) race condition as well.

archunit/src/test/java/com/tngtech/archunit/library/freeze/TextFileBasedViolationStoreTest.java
```diff
@@ -108,6 +111,38 @@ public void stores_violations_of_multiple_rules() {
...
+    public void is_safe_when_multiple_instances_save_to_the_same_store_concurrently() throws Exception {
+        int ruleCount = 20;
+        try (ExecutorService executor = Executors.newFixedThreadPool(8)) {
+            CountDownLatch startSignal = new CountDownLatch(1);
+
+            IntStream.range(0, ruleCount).forEach(i -> executor.submit(() -> {
+                ArchRule concurrentRule = rule("concurrent rule " + i);
+                ViolationStore instanceStore = new TextFileBasedViolationStore();
+                instanceStore.initialize(defaultStoreProperties());
+                startSignal.await();
```
StefanGraeber: please await the startSignal before calling initialize to also cover the case of a non-existing directory (even though it will likely still not be cached by the test because of a very small time frame)

StefanGraeber (changes requested):
Thanks for contributing 🙂 
The code itself looks good, just one minor edge case in the case of a non-existing directory which remains as race condition.

to make the pipeline pass, you must include a `Signed-off-by:` footer to your commit
https://github.com/TNG/ArchUnit/pull/1656/checks?check_run_id=85373415790

StefanGraeber:
you might have a merge conflict with #1655 which will be submitted soon.
It migrates some tests from Junit4 to Junit5 and thus exchange the Temporary File provider Rule with an Extension.
I think it has not touched the same test as you, but I might have missed it.

kelunik:
> you might have a merge conflict with https://github.com/TNG/ArchUnit/pull/1655 which will be submitted soon.

Thanks for the heads up! It didn't have a conflict, but I rebased anyway and converted the test to JUnit 5.

>  The code itself looks good, just one minor edge case in the case of a non-existing directory which remains as race condition.

True, I added a separate test for that.

kelunik:
Oh, these old Java versions, should be fixed.

archunit/src/test/java/com/tngtech/archunit/library/freeze/TextFileBasedViolationStoreConcurrencyTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @RepeatedTest(5)
+    public void is_safe_when_multiple_instances_initialize_concurrently() throws Exception {
+        runConcurrently(8, (i, startSignal) -> {
+            ViolationStore instanceStore = new TextFileBasedViolationStore();
+            startSignal.await();
+            instanceStore.initialize(defaultStoreProperties());
+        });
+
+        // The real test here is that StoreInitializationFailedException is not thrown
```
StefanGraeber, line 50: optional: instead of writing the comment, you can also `assertThatNoException().isThrownBy(() -> instanceStore.initialize(defaultStoreProperties()));` in line 46
same in line 91

do you know whether the AssertionError (it's an Error instead of an Exception) propagates out of the Future?
If not, my proposal might not work.

archunit/src/test/java/com/tngtech/archunit/library/freeze/TextFileBasedViolationStoreConcurrencyTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+        Properties storedRules = readProperties(new File(configuredFolder, "stored.rules"));
+        assertThat(storedRules).hasSize(ruleCount);
+
+        ViolationStore readerStore = new TextFileBasedViolationStore();
+        readerStore.initialize(defaultStoreProperties());
+        for (int i = 0; i < ruleCount; i++) {
+            ArchRule concurrentRule = rule("concurrent rule " + i);
+            assertThat(readerStore.contains(concurrentRule)).as("store contains " + concurrentRule.getDescription()).isTrue();
+            assertThat(readerStore.getViolations(concurrentRule)).containsOnly("violation " + i);
```
StefanGraeber: I'd prefer `containsExactly` over `containsOnly` sicne it reads nicer and fails on duplicates

archunit/src/test/java/com/tngtech/archunit/library/freeze/TextFileBasedViolationStoreConcurrencyTest.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    private void runConcurrently(int taskCount, ConcurrentTask task) throws Exception {
+        try (ExecutorService executor = Executors.newFixedThreadPool(8)) {
+            CountDownLatch startSignal = new CountDownLatch(1);
+            List<Future<?>> futures = new ArrayList<>();
+
+            IntStream.range(0, taskCount).forEach(i -> futures.add(executor.submit(() -> {
+                task.run(i, startSignal);
+                return null;
```
StefanGraeber: Instead of adding the Future object to the list as side effect, you could also use a `.map` call to submit them and collect the futures into the list.
Or use a plain old indexed loop.
I find both easier to read in this case.

please consider to a default method to ConcurrentTask:
```
default Callable toCallable(int ruleIndex, CountDownLatch startSignal) {
    return () -> {
        this.run(ruleIndex, startSignal);
        return null;
    };
}
```
and simplify to `executor.submit(task.toCallable(i, startSignal))`

StefanGraeber (changes requested):
Please squash at least the last two commits to avoid having an invalid intermediate state.

The other comments are not hugely important to me. If you have an opposing preference, it's also ok for me

kelunik:
@StefanGraeber Should all be addressed except for the optional comment on the assertion move.

StefanGraeber (approved):

PR #1658, opened 2026-07-09
[part2]add support JUnit 6.x testing framework. 

arukiidou:
## Summary
Add support for JUnit 6
Remove duplicate entries of JUnit 5 and JUnit 6 from the release check
- Support JUnit 6.x testing framework via aligned module #1556
- similar Add Junit5 support #94

## action checklists 

- [x] add main module and add junit6 dependency
- [x] add example-test
- [ ] add integration-test
- [ ] add documents
- [ ] waiting for the restructuring to test setup by maintainers
- [ ] code cleanup

PR #1659, opened 2026-07-09
Add support for JUnit6

### LINKED ISSUE #1556: Support JUnit 6.x testing framework via aligned module
The JUnit 6.0 release introduces a unified versioning approach: 
> all modules of JUnit Platform, Jupiter, and Vintage now use the same version number: 6.0.0.

[source - JUnit 6.x upgrade guide](https://github.com/junit-team/junit-framework/wiki/Upgrading-to-JUnit-6.0)

As ArchUnit’s `archunit-junit5` module is built against JUnit 5.x and Platform 1.x, it cannot be used in environments which adopt JUnit 6.x (Platform 6.x + Jupiter 6.x). This creates a blocker for testing strategies in modern Java projects.

## Request
- Introduce a module (e.g., archunit-junit6) targeting JUnit 6.x compatibility.
- Alternatively upgrade archunit-junit5 to detect/enable JUnit 6 support when Platform 6.x and Jupiter 6.x are on the classpath.
- Handle any JUnit 6 API changes (e.g., removed modules, null‐safety annotations) as noted in the JUnit upgrade guide.

## Goal
Allow teams using ArchUnit for architectural tests to migrate to JUnit 6.x without compromising dependency alignment or test stability.

schulzjo-tng:
- [x] Check if #1658 covers anything not covered here
- [x] Consider cleaning up commit history 
- [x] Add documentation
- [x] Reminder to address https://github.com/TNG/ArchUnit/pull/1450/changes#r3570218458


arukiidou:
Part 1 and Part 2 are almost identical.
If possible, I　recommend cherry-picking from Part 2.

Only the following three things should be different:
- It starts over again from the addition of the shared tests
- The commits have been split into smaller ones to make things easier
- example-test has been added

archunit-integration-test/shared/src/main/java/com/tngtech/archunit/integration/ExamplesIntegrationTestBase.java
```diff
@@ -616,7 +616,15 @@ Stream<T> DependencyRulesTest(Class<?> dependencyRulesTestClass) {
 
     abstract Stream<T> FrozenRulesTest();
 
-    Stream<T> FrozenRulesTest(Class<?> frozenRulesTestClass, Consumer<Runnable> withTemporaryViolationStore) {
+    Stream<T> FrozenRulesTest(Class<?> frozenRulesTestClass) {
+        return frozenRulesExpectedTestFailures(frozenRulesTestClass).toDynamicTests();
+    }
```
schulzjo-tng, line 621: no `aroundTestInvoke` is needed for JUnit6 - it was only needed on Java 8 and 11, which JUnit6 doesn't support

archunit-junit/build.gradle
```diff
@@ -59,27 +59,8 @@ shadowJar {
...
-        from zipTree(jarPath)
-        archiveFileName = tempPath.name
-
-        doLast {
-            assert jarPath.delete()
-            assert tempPath.renameTo(jarPath)
-        }
-    }
-    finishArchive.dependsOn removeDuplicateThirdParty
-}
```
schulzjo-tng, line 79: This task caused build failures on Windows. After some investigation, I concluded that it is no longer needed - the existing `dependencies.exclude()` make this transformation an identity operation. Also, there is test coverage for missing `exclude` statements (see `:checkArtifact`), so we can't accidentally publish a release where the problem that this code meant to address is present.

archunit-maven-test/verification/TestResultTest.java
```diff
@@ -131,11 +131,16 @@ Set<SingleTest> getFailedTests(Match testCaseTags) {
 
             Set<SingleTest> result = new HashSet<>();
             for (int i = 0; i < classNames.size(); i++) {
-                result.add(new SingleTest(classNames.get(i), testNames.get(i)));
+                result.add(new SingleTest(toSimpleClassName(classNames.get(i)), testNames.get(i)));
```
schulzjo-tng, line 134: This was needed due to a change in the test reports (major version upgrade)

archunit-maven-test/build.gradle
```diff
@@ -188,6 +193,12 @@ def addMavenTest = { IntegrationTestConfig config ->
 
         tasks.create(name: runMavenTest, dependsOn: [prepareMavenTest, executeRules, verifyRules, cleanUpMavenTest])
 
+        [prepareMavenTest, executeRules, verifyRules, cleanUpMavenTest, runMavenTest].each { taskName ->
+            tasks[taskName].onlyIf("requires Java ${config.minimumJavaVersion} or higher") {
+                config.supports(integrationTestJavaVersion)
+            }
+        }
```
schulzjo-tng, line 199: We cannot run the JUnit6 tasks on Java 8 and 11, hence this new block

archunit-maven-test/build.gradle
```diff
@@ -119,23 +122,25 @@ def addMavenTest = { IntegrationTestConfig config ->
...
-                                                |            <artifactId>${targetDependency.name}</artifactId>
-                                                |            <version>${targetDependency.version}</version>${scopePart}
-                                                |        </dependency>""".stripMargin().trim()
-                                        }
+                        .replace('#{archunit.version}', "${version}")
+                        .replace('#{archunitTestArtifact}', config.archunitTestArtifact)
+                        .replace('#{repositories}', repositories)
+                        .replace('#{javaVersion}', "${config.javaVersion}")
+                        .replace('#{mavenSurefirePluginVersion}', "${mavenSurefirePlugin.version}")
+                        .replace('#{mavenCompilerPluginVersion}', "${mavenCompilerPlugin.version}")
```
schulzjo-tng, line 129: This way, the plugins are updated via dependabot

archunit-maven-test/build.gradle
```diff
@@ -198,35 +209,59 @@ def addMavenTest = { IntegrationTestConfig config ->
...
+def junit6VintageEngine = libs.junit6VintageEngine.get()
 
-def vintageEngine = libs.junit5VintageEngine.get()
-integrationTestConfigs << new IntegrationTestConfig(
-        javaVersion: integrationTestJavaVersion,
-        testType: TestType.JUNIT5,
-        surefireExampleConfiguration: '<groups>example</groups>',
-        archunitTestArtifact: 'archunit-junit5',
-        additionalDependencies: """
+def testConfigFor = (TestType testType) -> {
```
schulzjo-tng, line 211: This restructure makes it harder to forget adding a case for a new JUnit version

archunit-self-test/src/test/java/com/tngtech/archunit/ArchUnitArchitectureTest.java
```diff
@@ -115,33 +108,4 @@ private static DescribedPredicate<JavaCall<?>> classIsResolvedViaReflection() {
...
-        public boolean includes(Location location) {
-            boolean include = false;
-            for (String sourceRoot : SOURCE_ROOTS) {
-                if (location.contains(sourceRoot)) {
-                    include = true;
-                }
-            }
-            return include && DO_NOT_INCLUDE_TESTS.includes(location);
-        }
-    }
```
schulzjo-tng, line 146: We now decide on a subproject basis if something is "production code"

archunit-self-test/src/test/java/com/tngtech/archunit/ArchUnitExampleArchitectureTest.java
schulzjo-tng, line 1: moved to `archunit-self-test/src/main/java/com/tngtech/archunit/ArchUnitExampleArchitectureRules.java` with too many changes

archunit-self-test/build.gradle
```diff
@@ -9,14 +8,6 @@ description = "ArchUnit tests for ArchUnit and ArchUnit-Examples themselves"
...
-    implementation project(path: ':archunit')
-    implementation project(path: ':archunit-junit4')
-    implementation project(path: ':archunit-junit5')
-    implementation project(path: ':archunit-junit5-engine')
-    implementation project(path: ':archunit-example:example-plain')
-    implementation project(path: ':archunit-example:example-plain', configuration: 'tests')
-    implementation project(path: ':archunit-example:example-junit4', configuration: 'tests')
-    implementation project(path: ':archunit-example:example-junit5', configuration: 'tests')
-
-    testImplementation project(path: ':archunit-junit5')
```
schulzjo-tng, line 21: The self-tests are now imported by those packages. This was necessary because classes between archunit-junit5 and archunit-junit6 would collide, and only one of them would get tested

buildSrc/src/main/groovy/archunit.self-test-conventions.gradle
```diff
@@ -0,0 +1,29 @@
+sourceSets {
+    archTest {
```
schulzjo-tng, line 2: We use a separate source set because the self-tests need a specific archunit-junit* version, which would otherwise pollute the test classpath or interfere with other tests

buildSrc/src/main/groovy/archunit.java-release-check-conventions.gradle
```diff
@@ -20,13 +20,12 @@ task ensureReleaseCheckUpToDate {
             VersionCatalog versionCatalog = versionCatalogs.named("libs")
             def dependencies = versionCatalog.libraryAliases.collect { versionCatalog.findLibrary(it).get().get() }
             def matchingProjectDependency = dependencies.find {
```
schulzjo-tng, line 22: The old logic no longer worked on dependencies to `junit-platform-launcher` which would get mixed up

archunit-maven-test/build.gradle
```diff
@@ -198,35 +209,59 @@ def addMavenTest = { IntegrationTestConfig config ->
...
-        ),
-        new IntegrationTestConfig(
-                javaVersion: integrationTestJavaVersion,
-                testType: TestType.JUNIT4,
-                surefireExampleConfiguration: '<groups>com.tngtech.archunit.exampletest.junit4.Example</groups>',
-                archunitTestArtifact: 'archunit-junit4'
-        )
-]
+def junit5VintageEngine = libs.junit5VintageEngine.get()
+def junit6VintageEngine = libs.junit6VintageEngine.get()
```
hankem: Those could be moved to the `case TestType.JUNIT5` / `JUNIT6` blocks.

archunit-maven-test/verification/TestResultTest.java
hankem: Which should be isolated in a separate commit, as it's entirely unrelated to `Add JUnit6 test coverage`. 😉

archunit-integration-test/junit6/build.gradle
```diff
@@ -0,0 +1,24 @@
...
+    testImplementation project(path: ':archunit', configuration: 'tests')
+    testImplementation project(path: ':archunit-junit6-api')
+    testImplementation project(path: ':archunit-junit6-engine-api')
+    testImplementation project(path: ':archunit-integration-test:shared')
+    testImplementation project(path: ':archunit-example:example-plain')
+    testImplementation project(path: ':archunit-example:example-plain', configuration: 'tests')
+    testImplementation project(path: ':archunit-example:example-junit6', configuration: 'tests')
+
+    testRuntimeOnly project(path: ':archunit-junit6-engine')
+    // TODO still up-to-date?
```
hankem: TODO

archunit-maven-test/build.gradle
hankem: → 01f4f0188421a5281c53906fb6fbea27cf12045d

archunit-maven-test/verification/TestResultTest.java
hankem: → 2c2fcce0c9844d31ae35027a6f01c56cccfcc5af

schulzjo-tng: Fair enough. It is a necessary prerequisite since the old version of Surefire didn't work with JUnit 6.

buildSrc/src/main/groovy/archunit.self-test-conventions.gradle
```diff
@@ -0,0 +1,29 @@
...
+        compileClasspath += sourceSets.main.output + configurations.testCompileClasspath
+        runtimeClasspath += output + sourceSets.main.output + configurations.testRuntimeClasspath
+    }
+}
+
+dependencies {
+    archTestImplementation project(path: ':archunit-self-test')
+    // We stay on JUnit5 for the self-tests because of Java 8 compatibility.
+    archTestImplementation project(path: ':archunit-junit5-api')
+    archTestRuntimeOnly project(path: ':archunit-junit5-engine')
```
hankem: We can combine those:
```suggestion
    archTestImplementation project(path: ':archunit-junit5')
```

buildSrc/src/main/groovy/archunit.self-test-conventions.gradle
```diff
@@ -0,0 +1,29 @@
...
+    archTestRuntimeOnly libs.junit5PlatformLauncher
+}
+
+def architectureTest = tasks.register("architectureTest", Test) {
+    description = 'Runs the shared ArchUnit architecture rules.'
```
hankem: Very tiny detail: I'd drop "ArchUnit" here (it's an implementation detail), and would say that we "run tests" or "check rules". 😅
```suggestion
    description = 'Runs the shared architecture tests.'
```

archunit/src/archTest/java/com/tngtech/archunit/ArchUnitArchitectureTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+import static com.tngtech.archunit.ArchitectureTestImportOptions.DoNotIncludeSelfTests;
+import static com.tngtech.archunit.ArchitectureTestImportOptions.sourceRootOf;
+
+@AnalyzeClasses(
+        packages = "com.tngtech.archunit",
+        importOptions = {DoNotIncludeTests.class, DoNotIncludeSelfTests.class, ArchUnitArchitectureTest.DoNotIncludeTestResources.class})
+public class ArchUnitArchitectureTest {
+    private ArchUnitArchitectureTest() {
+    }
```
hankem: Do we need that?
```suggestion
```

archunit-self-test/src/main/java/com/tngtech/archunit/PublicAPIRules.java
```diff
@@ -254,10 +267,12 @@ public void check(JavaMember member, ConditionEvents events) {
...
-                List<JavaClass> enclosingClasses = newArrayList(member.getOwner());
-                while (getLast(enclosingClasses).getEnclosingClass().isPresent()) {
-                    enclosingClasses.add(getLast(enclosingClasses).getEnclosingClass().get());
-                }
+                List<JavaClass> enclosingClasses = new ArrayList<>();
+                Optional<JavaClass> last = Optional.of(member.getOwner());
+                do {
+                    enclosingClasses.add(last.get());
+                    last = last.get().getEnclosingClass();
+                } while (last.isPresent());
```
hankem: Probably just a personal taste, but I'd keep it a `while` loop even though the first `isPresent` check is unnecessary. Also I'm not sure about the name `last`... How about this?
```suggestion
                Optional<JavaClass> current = Optional.of(member.getOwner());
                while (current.isPresent()) {
                    enclosingClasses.add(current.get());
                    current = current.get().getEnclosingClass();
                }
```

hankem: Or even
```suggestion
                for (JavaClass current = member.getOwner(); current != null; current = current.getEnclosingClass().orElse(null)) {
                    enclosingClasses.add(current);
                }
```
😉

archunit/src/archTest/java/com/tngtech/archunit/ArchUnitArchitectureTest.java
```diff
@@ -0,0 +1,32 @@
...
+import com.tngtech.archunit.junit.AnalyzeClasses;
+import com.tngtech.archunit.junit.ArchTest;
+import com.tngtech.archunit.junit.ArchTests;
+
+import static com.tngtech.archunit.ArchitectureTestImportOptions.DoNotIncludeSelfTests;
+import static com.tngtech.archunit.ArchitectureTestImportOptions.sourceRootOf;
+
+@AnalyzeClasses(
+        packages = "com.tngtech.archunit",
+        importOptions = {DoNotIncludeTests.class, DoNotIncludeSelfTests.class, ArchUnitArchitectureTest.DoNotIncludeTestResources.class})
```
hankem: Is `DoNotIncludeTestResources` needed?

archunit/build.gradle
```diff
@@ -20,6 +21,7 @@ dependencies {
         exclude module: 'assertj-core'
         exclude module: 'guava'
     }
+    testImplementation project(path: ':archunit-junit5')
```
hankem: I don't think that's needed.

archunit-example/example-junit4/src/archTest/java/com/tngtech/archunit/exampletest/junit4/ArchUnitArchitectureTest.java
```diff
@@ -0,0 +1,18 @@
...
+import com.tngtech.archunit.ArchUnitExampleArchitectureRules;
+import com.tngtech.archunit.junit.AnalyzeClasses;
+import com.tngtech.archunit.junit.ArchTest;
+import com.tngtech.archunit.junit.ArchTests;
+import com.tngtech.archunit.junit.ArchUnitRunner;
+import org.junit.runner.RunWith;
+
+@RunWith(ArchUnitRunner.class)
+@AnalyzeClasses(packages = "com.tngtech.archunit.exampletest.junit4")
+public class ArchUnitArchitectureTest {
```
hankem: Should we give different names to the 12 `ArchUnitArchitectureTest` classes?
(They do reside in different packages, so they can already be distinguished... but it might be more convenient.)

archunit-junit/junit4/src/archTest/java/com/tngtech/archunit/junit/ArchUnitArchitectureTest.java
```diff
@@ -0,0 +1,16 @@
...
+import org.junit.runner.RunWith;
+
+@RunWith(ArchUnitRunner.class)
+@AnalyzeClasses(packages = "com.tngtech.archunit.junit", importOptions = {DoNotIncludeTests.class, DoNotIncludeSelfTests.class})
+public class ArchUnitArchitectureTest {
+    private ArchUnitArchitectureTest() {
+    }
+
+    @ArchTest
+    public static final ArchTests public_api_rules = ArchTests.in(PublicAPIRules.class);
```
hankem: I would execute all `ArchUnitArchitectureRules` (which include these `PublicAPIRules`) in the `junit*` modules `ArchUnitArchitectureTest`s.
There are two more that would require `allowEmptyShould(true)`, but I'd hope that we could configure that for those modules in `archunit.properties` (via [`archRule.failOnEmptyShould=false`](https://www.archunit.org/userguide/html/000_Index.html#_fail_rules_on_empty_should)).
(I didn't get it to work yet, but didn't have enough time unfortunately.)

archunit/src/archTest/java/com/tngtech/archunit/ArchUnitArchitectureTest.java
```diff
@@ -0,0 +1,32 @@
...
+public class ArchUnitArchitectureTest {
+    private ArchUnitArchitectureTest() {
+    }
+
+    @ArchTest
+    public static final ArchTests architecture_rules = ArchTests.in(ArchUnitArchitectureRules.class);
```
hankem: The field and class don't have to be public.

archunit-self-test/build.gradle
hankem: Should we rename this to something like self-test-library?

archunit-junit/junit4/src/archTest/java/com/tngtech/archunit/junit/ArchUnitArchitectureTest.java
```diff
@@ -0,0 +1,16 @@
+package com.tngtech.archunit.junit;
+
+import com.tngtech.archunit.ArchitectureTestImportOptions.DoNotIncludeSelfTests;
+import com.tngtech.archunit.PublicAPIRules;
+import com.tngtech.archunit.core.importer.ImportOption.DoNotIncludeTests;
+import org.junit.runner.RunWith;
+
+@RunWith(ArchUnitRunner.class)
```
hankem: Aren't `archictectureTest`s executed with `archunit-junit5`?

Same for `archunit-example/example-junit4/src/archTest/java/com/tngtech/archunit/exampletest/junit4/ArchUnitArchitectureTest.java`.

archunit/src/archTest/java/com/tngtech/archunit/ArchUnitArchitectureTest.java
schulzjo-tng: SonarLint suggested to make the constructors of static classes private since they shouldn't ever be instantiated. I don't mind reverting these changes.

archunit/src/archTest/java/com/tngtech/archunit/ArchUnitArchitectureTest.java
schulzjo-tng: Good catch! It was needed when the arch tests were on the same `sourceSet` as the unit tests, which is no longer the case.

Edit2: Restored original comment

archunit-junit/junit4/src/archTest/java/com/tngtech/archunit/junit/ArchUnitArchitectureTest.java
schulzjo-tng: I went over the tests before one by one and concluded that none of them make sense on the junit packages (just one example: the layers that are being tested don't exist in any of the junit* modules). I concluded that it makes more sense to only run the tests that actually make sense since it also reduces confusion for future readers.

I also prefer to keep the `allowEmptyShould` on a rule level since not all rules need it. That way it's easier to catch a misconfiguration.

buildSrc/src/main/groovy/archunit.self-test-conventions.gradle
```diff
@@ -0,0 +1,29 @@
...
+        runtimeClasspath += output + sourceSets.main.output + configurations.testRuntimeClasspath
+    }
+}
+
+dependencies {
+    archTestImplementation project(path: ':archunit-self-test')
+    // We stay on JUnit5 for the self-tests because of Java 8 compatibility.
+    archTestImplementation project(path: ':archunit-junit5-api')
+    archTestRuntimeOnly project(path: ':archunit-junit5-engine')
+    archTestRuntimeOnly libs.junit5PlatformLauncher
```
schulzjo-tng: There is another significant structural problem here (which has no impact right now but may have one in the future):
- If I introduce a violation in e.g. `archunit-junit5-jupiter`, the violation is currently found by `archunit-junit5-jupiter:architectureTest`
- If I introduce a violation in e.g. `archunit-junit6-jupiter`, the violation is currently found by `archunit-junit6-jupiter:architectureTest`
- However, if I replace this shared dependency with `archunit-junit6`, then the violations in `archunit-junit5-jupiter` are no longer found.

What I believe happens here is a classpath collision betwen `archunit-junit5` and `archunit-junit6` on the `architectureTests` sourceSet. I'd solve this the following way:
- Remove all dependencies on a specific JUnit version from this shared file
- Self-test all `archunit-junit*` subprojects with their own matching JUnit version
- Self-test all other subprojects with the most recent appropriate JUnit version (which would be JUnit5 in this case due to the Java 8 constraint)

@hankem I'll implement this fix when you are done with your changes.

archunit-integration-test/junit6/build.gradle
schulzjo-tng:
```suggestion
```
Forgot to remove the TODO comment. The dependency is still needed

archunit-junit/junit4/src/archTest/java/com/tngtech/archunit/junit/ArchUnitArchitectureTest.java
schulzjo-tng: I'll revisit this point after addressing https://github.com/TNG/ArchUnit/pull/1659#discussion_r3631231676

buildSrc/src/main/groovy/archunit.self-test-conventions.gradle
schulzjo-tng: Fixed in https://github.com/TNG/ArchUnit/pull/1659/changes/5968273c66d9aa7392625808aeddae8ce57026c4

archunit-example/example-junit4/src/archTest/java/com/tngtech/archunit/exampletest/junit4/ArchUnitArchitectureTest.java
schulzjo-tng: blocked by resolution of https://github.com/TNG/ArchUnit/pull/1659#discussion_r3628247220

archunit-self-test/build.gradle
schulzjo-tng: Just the module name, or the folder as well?

archunit-junit/junit4/src/archTest/java/com/tngtech/archunit/junit/ArchUnitArchitectureTest.java
hankem: > the layers that are being tested don't exist in any of the junit* modules

The fact that the layers are respected does not mean that the test does not apply. 🤷‍♂️

I thought that in particular the tests that reflection or ASM is not used would still make sense, and that it would be nice to test all rules without `allowEmptyShould` in the `archunit` module, and only allow that in the `junit*` modules.

archunit-self-test/build.gradle
hankem: I'd keep them consistent – and `archunit-self-test` has anyway not been around for long...

archunit-example/example-junit4/src/archTest/java/com/tngtech/archunit/exampletest/junit4/ArchUnitArchitectureTest.java
schulzjo-tng: pattern: `ArchUnitArchitectureTest`, `ArchUnitJUnit4ArchitectureTest`, etc.

archunit-self-test/build.gradle
schulzjo-tng: -> `archunit-self-test-library`

archunit-self-test-library/src/main/java/com/tngtech/archunit/ArchUnitArchitectureRules.java
```diff
@@ -58,11 +49,13 @@ public class ArchUnitArchitectureTest {
...
 
-            // This is a conscious exception, to allow a more concise API. Otherwise the configuration of the
+            // This is a conscious exception, to allow a more concise API. Otherwise, the configuration of the
             // ClassResolver would need to be moved to `importer` making it harder to use,
             // or we would need to remove the bound from Class<? extends ClassResolver>, which also makes the
             // API harder to use
-            .ignoreDependency(ArchConfiguration.class, ClassResolver.class);
+            .ignoreDependency(ArchConfiguration.class, ClassResolver.class)
+            // Not all layers are present on all archunit-junit* packages
+            .withOptionalLayers(true);
```
schulzjo-tng: @hankem this setting is not covered by `archRule.failOnEmptyShould=false`. We can either set this `withOptionalLayers` or self-check the entirety of ArchUnit in all production packages (i.e. `@AnalyzeClasses(packages = "com.tngtech.archunit"` instead of `@AnalyzeClasses(packages = "com.tngtech.archunit.junit.[...]"`). What would you prefer?

archunit-example/example-junit4/build.gradle
```diff
@@ -1,12 +1,22 @@
...
 ext.moduleName = 'com.tngtech.archunit.example.junit4'
 
 dependencies {
     testImplementation project(path: ':archunit-junit4')
     testImplementation project(path: ':archunit-example:example-plain')
+
+    // The ArchUnit self-tests use JUnit5 features, so we self-test this example subproject using JUnit5.
+    // The JUnit versions only need to match for the archunit-junit* subprojects.
+    archTestImplementation project(path: ':archunit-junit4')
+    archTestImplementation project(path: ':archunit-junit5')
```
hankem: We don't need both, right?
```suggestion
    archTestImplementation project(path: ':archunit-junit5')
```

docs/userguide/002_Installation.adoc
hankem: Should we flip the order of `JUnit 4` and `JUnit 5 & 6`?
The sequence of 4, 5, 6 is nice, but `JUnit 4` might be(come) less relevant.

hankem (approved):
Thanks for the effort! 💙

archunit-self-test-library/src/main/java/com/tngtech/archunit/ArchUnitArchitectureRules.java
hankem: I like the idea to consistently `@AnalyzeClasses(packages = "com.tngtech.archunit"`, then the tests would also catch if the junit modules introduced a class in a package they shouldn't.

PR #1660, opened 2026-07-09
detect custom Gradle test source sets in test import options

### LINKED ISSUE #1578: ImportOption.DoNotIncludeTests and ImportOption.OnlyIncludeTests don't detect tests in custom Gradle source sets
## Problem

The `ImportOption.Predefined.GRADLE_TEST_PATTERN` in ArchUnit fails to detect tests located in custom Gradle source sets (e.g., `integrationTest`, `functionalTest`, etc.) because the pattern only matches lowercase `test` in the build path.

## Current Implementation

From [`com.tngtech.archunit.core.importer.ImportOption$Predefined` (line ~102):](https://github.com/TNG/ArchUnit/blob/main/archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java#L102)

```java
static final PatternPredicate GRADLE_TEST_PATTERN =
    new PatternPredicate(".*/build/classes/([^/]+/)?test/.*");
```

This pattern matches paths like:
- ✅ `/build/classes/java/test/` (standard test source set)
- ✅ `/build/classes/kotlin/test/`
- ❌ `/build/classes/java/integrationTest/` (custom source set)
- ❌ `/build/classes/kotlin/functionalTest/` (custom source set)
- ❌ `/build/classes/java/e2eTest/` (custom source set)

## Impact

When using `ImportOption.DoNotIncludeTests` or `ImportOption.OnlyIncludeTests`:
- Custom test source sets are **not excluded** when using `DoNotIncludeTests`
- Custom test source sets are **not included** when using `OnlyIncludeTests`

This causes incorrect behavior when architectural rules should apply to or exclude tests from custom source sets.

## Example Scenario

```kotlin
// Gradle configuration with custom source set
sourceSets {
    create("integrationTest") {
        compileClasspath += sourceSets.main.get().output
        runtimeClasspath += sourceSets.main.get().output
    }
}

// ArchUnit test that fails to recognize integrationTest classes as tests
@AnalyzeClasses(
    packages = "com.example",
    importOptions = [ImportOption.DoNotIncludeTests::class]
)
class ArchitectureTest {
    // This will incorrectly analyze integrationTest classes
    // because the pattern doesn't recognize them as tests
}
```

## Proposed Solution

The pattern should be more flexible to match any path segment containing "test" (case-insensitive) or "Test" suffix:

### Option 1: Case-insensitive matching
```java
".*/build/classes/([^/]+/)?(?i).*test.*/.*"
```

### Option 2: Match common test source set naming patterns
```java
".*/build/classes/([^/]+/)?([a-z]+)?[Tt]est(s)?/.*"
```

This would match:
- `test` (standard)
- `integrationTest`
- `functionalTest`
- `e2eTest`
- `tests` (plural variant)

## Workaround

Currently, users must implement custom `ImportOption` for each custom test source set:

```java
public class DoNotIncludeIntegrationTests implements ImportOption {
    private static final Pattern PATTERN =
        Pattern.compile(".*/build/classes/([^/]+/)?integrationTest/.*");

    @Override
        return !location.matches(PATTERN);
    }
}
```

## Environment

- ArchUnit version: 1.4.1
- Build tool: Gradle
- Custom source set name: `integrationTest`

Develop-KIM:
## What & why
Fixes #1578. `ImportOption.DoNotIncludeTests` and `OnlyIncludeTests` did not recognize tests in custom Gradle source sets (e.g. `integrationTest`, `functionalTest`, `e2eTest`), because `GRADLE_TEST_PATTERN` only matched the standard lowercase `test` folder. Custom test classes were therefore treated as production code — not excluded by `DoNotIncludeTests`, and not included by `OnlyIncludeTests`.

## Change
- Broaden `GRADLE_TEST_PATTERN` from `.*/build/classes/([^/]+/)?test/.*` to `.*/build/classes/([^/]+/)?(test|[^/]+Test)s?/.*`, so it also matches source-set output folders that follow the Gradle `...Test` naming convention (plus the plural `tests`).
- Requiring the camelCase capital `Test` boundary keeps `main` excluded, preserves the separately handled `testFixtures` folder (`DoNotIncludeGradleTestFixtures`), and avoids false positives such as `latest`.
- Update the `DoNotIncludeTests` Javadoc accordingly.
- Add `ImportOptionsTest` data points for `integrationTest`, `functionalTest`, `e2eTest`, `tests`, and a `testFixtures` → main regression guard.

## Verification
`./gradlew :archunit:test --tests "*ImportOptionsTest"` → 92 tests, 0 failures (the new source-set folders resolve as test locations; `testFixtures`/`main` stay non-test). `:archunit:spotlessCheck` passes.

---
This change was prepared with AI assistance (Claude); I have reviewed and verified it and take responsibility for the contribution.

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
+++ b/archunit/src/main/java/com/tngtech/archunit/core/importer/ImportOption.java
@@ -99,7 +99,7 @@ public interface ImportOption {
         };
 
         static final PatternPredicate MAVEN_TEST_PATTERN = new PatternPredicate(".*/target/test-classes/.*");
-        static final PatternPredicate GRADLE_TEST_PATTERN = new PatternPredicate(".*/build/classes/([^/]+/)?test/.*");
+        static final PatternPredicate GRADLE_TEST_PATTERN = new PatternPredicate(".*/build/classes/([^/]+/)?(test|[^/]+Test)s?/.*");
         static final PatternPredicate INTELLIJ_TEST_PATTERN = new PatternPredicate(".*/out/test/.*");
         static final Predicate<Location> TEST_LOCATION = MAVEN_TEST_PATTERN.or(GRADLE_TEST_PATTERN).or(INTELLIJ_TEST_PATTERN);
         static final Predicate<Location> NO_TEST_LOCATION = TEST_LOCATION.negate();
@@ -122,6 +122,8 @@ public interface ImportOption {
      * Best effort {@link ImportOption} to check rules only on main classes.<br>
      * NOTE: This excludes all class files residing in some directory
      * ../target/test-classes/.., ../build/classes/test/.. or ../build/classes/someLang/test/.. (Maven/Gradle standard).
+     * This also covers custom Gradle test source sets that follow the {@code ...Test} naming convention,
+     * e.g. ../build/classes/someLang/integrationTest/...
      * Thus it is just a best guess, how tests can be identified,
      * in other environments, it might be necessary, to implement the correct {@link ImportOption} yourself.
      */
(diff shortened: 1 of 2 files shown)


hankem (approved):
Thanks for the contribution! 💙

Issue #1633, opened 2026-06-03
Rule Priority can not be set


mmoossen:
Hi,

I realized that on violations a priority is logged:

    java.lang.AssertionError: Architecture Violation [Priority: MEDIUM] - Rule 'xxx.' was violated (x times)

MEDIUM is the default, but I do have rules i would like to mark with a different priority.

I was expecting it to be as easy as:

    classes().should(...).as("...").priority(Priority.HIGH);

but there is no way to change the priority in ArchRule.
The only way i found was to use `ArchRule.Factory.create(...)` which far to difficult.

The docs do not mention the priority at all.

any help greatly appreciated.

TheManWhoStaresAtCode:
Hi @mmoossen,

thanks for making us aware of this lack of documentation.

To answer your question - you can use `ArchRuleDefinition#priority` for that e.g.: 
```java
ArchRuleDefinition.priority(Priority.HIGH).classes().should(...).as("...");
```

hankem:
For reference: see also #129.

PR #1671, opened 2026-07-22
Add Documentation regarding ArchRule priority

TheManWhoStaresAtCode:
As pointed out in #1633 we lack documentation how to configure the ArchRule priority.

Besides updating the userguide I also added JavaDoc for the priority enum. It might be beneficial to also add JavaDoc to the ArchRuleDefinition method `priorities` to support IDE users. However I think this should be a dedicated task where we document all public methods of that class or even all of the language package.

### DIFF
--- a/archunit/src/main/java/com/tngtech/archunit/lang/Priority.java
+++ b/archunit/src/main/java/com/tngtech/archunit/lang/Priority.java
@@ -16,9 +16,17 @@
 package com.tngtech.archunit.lang;
 
 import com.tngtech.archunit.PublicAPI;
+import com.tngtech.archunit.core.domain.JavaClasses;
+import com.tngtech.archunit.lang.syntax.ArchRuleDefinition;
 
 import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
 
+/**
+ * Classifies the importance of an {@link EvaluationResult}, which is the result of {@link ArchRule#evaluate(JavaClasses)}.
+ * <p>
+ * {@link ArchRuleDefinition#priority(Priority)} can be used to create {@link ArchRule} implementations with a given priority.
+ * </p>
+ */
 @PublicAPI(usage = ACCESS)
 public enum Priority {
     @PublicAPI(usage = ACCESS)
--- a/docs/userguide/007_The_Lang_API.adoc
+++ b/docs/userguide/007_The_Lang_API.adoc
@@ -330,6 +330,21 @@ classes().that(haveAFieldAnnotatedWithPayload).should(onlyBeAccessedBySecuredMet
     .as("Payload may only be accessed in a secure way");
 ----
 
+=== Priority
+
+Most entrypoints in `ArchRuleDefinition` create ``ArchRule``s that evaluate to ``EvaluationResult``s with ``Priority.MEDIUM``; `ArchRuleDefinition.priority` allows to create rules with other `Priority`, e.g.:
+
+[source,java,options="nowrap"]
+----
+ArchRule rule = ArchRuleDefinition.priority(Priority.LOW).noClasses() // ...
+----
+
+The resulting error messages show the priority accordingly:
+
+[source,bash]
+----
+java.lang.AssertionError: Architecture Violation [Priority: LOW] - Rule 'no classes // ...
+----
 === Ignoring Violations
 
 In legacy projects there might be too many violations to fix at once. Nevertheless, that code


archunit/src/main/java/com/tngtech/archunit/lang/Priority.java
```diff
@@ -19,6 +19,20 @@
...
+ * The priority is included in the violation message of a failing rule
+ * (e.g. {@code Architecture Violation [Priority: HIGH] - ...}) and can be queried
+ * programmatically via {@link EvaluationResult#getPriority()}.
+ * </p>
+ * <p>
+ * If no priority is specified explicitly (for example via
+ * {@link com.tngtech.archunit.lang.syntax.ArchRuleDefinition#priority(Priority)
+ * ArchRuleDefinition.priority(Priority)}), an {@code ArchRule} is created with
+ * {@link #MEDIUM} priority.
+ * </p>
```
hankem: I actually wouldn't put this here: IMO, that's an individual choice of every single factory method in `ArchRuleDefinition`, not a fundamental concept of priorities.

When we have a complete documentation of `ArchRuleDefinition`s, each of them can have a (small) note saying that it generates rules with `Priority.MEDIUM`.

hankem: Maybe it's enough to hint at this?
```suggestion
 * <p>
 * {@link ArchRuleDefinition#priority(Priority)} can be used to create {@link ArchRule} implementations with a given priority.
 * </p>
```

archunit/src/main/java/com/tngtech/archunit/lang/Priority.java
```diff
@@ -19,6 +19,20 @@
 
 import static com.tngtech.archunit.PublicAPI.Usage.ACCESS;
 
+/**
+ * Classifies the importance of an {@link ArchRule}.
```
hankem: Priorities are not part of the `ArchRule` interface, but only of the `EvaluationResult` it generates.
I think it's an implementation detail that current implementations may use a single fixed priority, but it is entirely possible to have an `ArchRule` that generates `EvaluationResult`s with different priorities, based on whatever conditition.

hankem:
```suggestion
 * Classifies the importance of an {@link EvaluationResult}, which is the result of {@link ArchRule#evaluate()}.
```

archunit/src/main/java/com/tngtech/archunit/lang/Priority.java
```diff
@@ -19,6 +19,20 @@
...
+ * <p>
+ * The priority is included in the violation message of a failing rule
+ * (e.g. {@code Architecture Violation [Priority: HIGH] - ...}) and can be queried
```
hankem: That's an implementation detail of `FailureDisplayFormatFactory.DefaultFailureDisplayFormat`, and `FailureDisplayFormatFactory` supports injecting other `FailureDisplayFormat` implementations via `ArchConfiguration`.

TheManWhoStaresAtCode:
Thanks for the feedback @hankem. I updated the JavaDoc based on your suggestions.

docs/userguide/007_The_Lang_API.adoc
```diff
@@ -330,6 +330,34 @@ classes().that(haveAFieldAnnotatedWithPayload).should(onlyBeAccessedBySecuredMet
...
+[source,bash]
+----
+java.lang.AssertionError: Architecture Violation [Priority: LOW] -
+Rule 'no classes that reside in a package '..service..'
+should access classes that reside in a package '..controller..'' was violated (1 times):
+Method <some.pkg.service.SomeService.callController()>
+calls method <some.pkg.controller.SomeController.execute()>
+in (SomeService.java:14)
+----
+
```
hankem: I have a similar feeling about the user guide as I had about the JavaDoc. 🙈
* This sounds like `Priority` was a concept that has a 1:1 mapping to rules.
* I'd personally keep the example shorter.

```suggestion
=== Priority

Most entrypoints in `ArchRuleDefinition` create `ArchRule`s that evaluate to `EvaluationResult`s with `Priority.MEDIUM`; `ArchRuleDefinition.priority` allows to create rules with other `Priority`, e.g.:

[source,java,options="nowrap"]
----
ArchRule rule = ArchRuleDefinition.priority(Priority.LOW).noClasses() // ...
----

The resulting error messages show the priority accordingly:

[source,bash]
----
java.lang.AssertionError: Architecture Violation [Priority: LOW] - Rule 'no classes // ...
----
```

What do you think about it?
If you agree, I'd suggest to squash all commits and we'll merge it as a single change to `main`.

TheManWhoStaresAtCode: I applied your suggestion and also squashed the commits to a single one.

thanks for your feedback - really appreciated.

docs/userguide/007_The_Lang_API.adoc
```diff
@@ -330,6 +330,21 @@ classes().that(haveAFieldAnnotatedWithPayload).should(onlyBeAccessedBySecuredMet
     .as("Payload may only be accessed in a secure way");
 ----
 
+=== Priority
+
+Most entrypoints in `ArchRuleDefinition` create `ArchRule`s that evaluate to `EvaluationResult`s with `Priority.MEDIUM`; `ArchRuleDefinition.priority` allows to create rules with other `Priority`, e.g.:
```
hankem: <details><summary>Ah, sorry; it turned out that my first suggestion doesn't render properly.</summary>

Image
</details>

This should work, cf. [AsciiDoc → Text Formatting and Punctuation → Monospace → Unconstrained](https://docs.asciidoctor.org/asciidoc/latest/text/monospace/#unconstrained):
> As with other types of text formatting, if the text is bounded by word characters on either side, it must be enclosed in a double pair of backtick characters (`﻿``﻿`) in order for the formatting to be applied.

```suggestion
Most entrypoints in `ArchRuleDefinition` create ``ArchRule``s that evaluate to ``EvaluationResult``s with `Priority.MEDIUM`;
`ArchRuleDefinition.priority` allows to create rules with other `Priority`, e.g.:
```

TheManWhoStaresAtCode: good catch - adjusted accordingly. I amend the squashed commit.

hankem: Thank you! Sorry for the back and forth... 🙈

hankem (approved):

PR #1672, opened 2026-07-23
add package matcher section to user guide

### LINKED ISSUE #1507: better documentation for custom dsl
could you add a full section to the user guide that goes over how `..`, `.(*)..` `.(**)` etc work. I'm hoping it starts with you have this list of packages and classes. something like this

```
com.myorg.myapp.foo.Foo
com.myorg.myapp.foo.service.baz.BazService
com.myorg.myapp.controller.bar.BarController
com.myorg.myapp.controller.bar.baz.Baz
```

```
..controller..
```
would match only 

```
com.myorg.myapp.controller.bar.BarController
com.myorg.myapp.controller.bar.baz.Baz
```

```
..baz
```

would match?

```
com.myorg.myapp.foo.service.baz.BazService
com.myorg.myapp.controller.bar.baz.Baz
```

```
..foo.(*)
```

would match

```
com.myorg.myapp.foo.Foo
```

I'm kind of starting to guess... I found out some of my rules weren't doing what I thought they did. I know there's a smattering of this, but I feel like it needs expansion.

TheManWhoStaresAtCode:
* add a detailed section to the core API section, as PackageMatching resides there.
* reference the detailed section where the current user guide already touches the topic

Closes #1507

### DIFF
--- a/docs/userguide/003_Getting_Started.adoc
+++ b/docs/userguide/003_Getting_Started.adoc
@@ -50,8 +50,9 @@ ArchRule myRule = classes()
     .should().onlyBeAccessed().byAnyPackage("..controller..", "..service..");
 ----
 
-The two dots represent any number of packages (compare AspectJ Pointcuts). The returned
-object of type `ArchRule` can now be evaluated against a set of imported classes:
+The two dots represent any number of packages (compare AspectJ Pointcuts); for details on
+the supported package pattern syntax see <<Package Identifiers>>. The returned object of
+type `ArchRule` can now be evaluated against a set of imported classes:
 
 [source,java,options="nowrap"]
 ----
--- a/docs/userguide/006_The_Core_API.adoc
+++ b/docs/userguide/006_The_Core_API.adoc
@@ -396,3 +396,116 @@ ArchUnit's own rule APIs (compare <<The Lang API>>) never rely on the
 classpath though. Thus the evaluation of default rules and syntax combinations, described in the
 next section, does not depend on whether the classes were imported from the classpath or
 some JAR / folder.
+
+==== Package Identifiers
+
+Several ArchUnit methods accept a `String` package identifier to describe a set of packages,
+for example `resideInAPackage(..)` and `resideInAnyPackage(..)` (and their negated variants)
+in the Lang API, `slices().matching(..)` and `modules().definedByPackages(..)` in the Library
+API, or the component stereotypes of a PlantUML component diagram. All of them delegate to
+the same underlying https://javadoc.io/doc/com.tngtech.archunit/archunit/latest/com/tngtech/archunit/core/domain/PackageMatcher.html[`PackageMatcher`], whose syntax is inspired by AspectJ type patterns and
+extended with capturing groups.
+
+===== Wildcards
+
+The syntax is built from the following elements:
+
+[cols="1l,4"]
+|===
+|Pattern | Meaning
+
+|*
+| Matches any non-empty sequence of characters not containing the dot `+.+`, i.e. exactly one package segment.
+
+|..
+| Matches any (possibly empty) sequence of characters that may contain the dot `+.+`, i.e. any number of package segments (including zero).
+
+|(*)
+| Like `+*+`, but additionally _captures_ the matched segment, so it can be referenced later (e.g. as a slice identifier, see <<Slices>>).
+
+|(**)
+| Like `+..+`, but additionally _captures_ the matched segments.
+
+|[a\|b]
+| Alternation: matches either `a` or `b`. Alternations are only allowed inside brackets `+[...]+` or inside capturing groups `+(...)+`.
+|===
+
+Note that `(..)` is not a valid capturing group — use `(**)` instead. `()` and `[]` cannot
+be nested inside each other.
+
+The segments matched by capturing groups can be retrieved from the result of
+https://javadoc.io/doc/com.tngtech.archunit/archunit/latest/com/tngtech/archunit/core/domain/PackageMatcher.html#match(java.lang.String)[`PackageMatcher#match(String)`]
+via
+https://javadoc.io/doc/com.tngtech.archunit/archunit/latest/com/tngtech/archunit/core/domain/PackageMatcher.Result.html#getGroup(int)[`PackageMatcher.Result#getGroup(int)`]
+(groups are 1-based, in the order the capturing groups appear in the pattern).
+
+===== Worked Example
+
+Consider the following five classes and the packages they reside in:
+
+[cols="2,2"]
+|===
+| Class | Package
+
+| `com.myapp.controller.SomeController`     | `com.myapp.controller`
+| `com.myapp.service.SomeService`           | `com.myapp.service`
+| `com.myapp.service.impl.SomeServiceImpl`  | `com.myapp.service.impl`
+| `com.myapp.persistence.dao.SomeDao`       | `com.myapp.persistence.dao`
+| `com.myapp.persistence.dao.jpa.SomeJpa`   | `com.myapp.persistence.dao.jpa`
+|===
+
+The following table shows a range of package identifiers and which of the five packages
+above each of them matches:
+
+[cols="2,3"]
+|===
+| Package Identifier | Matches
+
+| `com.myapp.service`
+| `com.myapp.service` only (exact match).
+
+| `com.myapp.*`
+| `com.myapp.controller` and `com.myapp.service`
+(exactly one more segment after `com.myapp`).
+
+| `com.myapp..`
+| All five packages — `com.myapp..` matches `com.myapp` itself and every subpackage of it.
+
+| `..service`
+| `com.myapp.service` only (segment `service` at the end).
+
+| `..service..`
+| `com.myapp.service` and `com.myapp.service.impl`
+(any package containing a segment `service`).
+
+| `..dao..`
+| `com.myapp.persistence.dao` and `com.myapp.persistence.dao.jpa`.
+
+| `..impl`
+| `com.myapp.service.impl` only.
+
+| `com.myapp.(*)`
+| `com.myapp.controller` (group 1 = `controller`) and
+`com.myapp.service` (group 1 = `service`).
+The deeper packages do not match because `(*)` allows only a single segment.
+
+| `com.myapp.(*)..`
+| All five packages; group 1 captures the first sub-package below `com.myapp`
+(`controller`, `service`, `service`, `persistence`, `persistence` respectively).
+This is the typical pattern used by `slices().matching(..)`.
+
+| `..[service\|controller]..`
+| `com.myapp.controller`, `com.myapp.service` and `com.myapp.service.impl`.
+|===
+
+===== What is Matched Against the Pattern
+
+An important detail is that the package identifier is matched against the
+**package name of a class**, not against the fully qualified class name.
+So for `com.myapp.service.SomeService` the string that is checked against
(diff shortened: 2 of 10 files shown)


docs/userguide/003_Getting_Started.adoc
```diff
@@ -50,8 +50,9 @@ ArchRule myRule = classes()
     .should().onlyBeAccessed().byAnyPackage("..controller..", "..service..");
 ----
 
-The two dots represent any number of packages (compare AspectJ Pointcuts). The returned
-object of type `ArchRule` can now be evaluated against a set of imported classes:
+The two dots represent any number of packages (compare AspectJ Pointcuts); for details on
+the supported package pattern syntax see <<Package Identifiers>>. The returned object of
```
StefanGraeber, line 54: This reference works in IntelliJ with the proper plugin, but not on the github page in the browser.
Since this targets new users, they are unlikely to check out the repo to get working links. 

The correct syntax seems to be
```
<<006_The_Core_API.adoc#package-identifiers,Package Identifiers>>
```
here 
```
[#package-identifiers]
=== Package Identifiers
```
in the 006 file

Since all links are currently broken, you can also stick with the current version and I fix all links in another PR

docs/userguide/006_The_Core_API.adoc
```diff
@@ -396,3 +396,114 @@ ArchUnit's own rule APIs (compare <<The Lang API>>) never rely on the
...
+
+| `com.myapp.service`
+| `com.myapp.service` only (exact match).
+
+| `com.myapp.*`
+| `com.myapp.controller` and `com.myapp.service`
+(exactly one more segment after `com.myapp`).
+
+| `com.myapp..`
+| All five packages (any number of segments under `com.myapp`, including zero).
```
StefanGraeber: "including zero" is currently not fully demonstrated as there is no example class in the root package.
Maybe add a class there as well?

I haven't found any referecne of the classes so far, so it might be not relevant

docs/userguide/006_The_Core_API.adoc
```diff
@@ -396,3 +396,114 @@ ArchUnit's own rule APIs (compare <<The Lang API>>) never rely on the
...
+| `*`
+| Matches any sequence of characters **not containing** the dot `.`,
+i.e. exactly one package segment.
+
+| `..`
+| Matches any sequence of characters that **may contain** the dot `.`,
+i.e. any number of package segments (including zero).
+
+| `(*)`
+| Like `*`, but additionally _captures_ the matched segment,
```
StefanGraeber: I find no explanation of what it means that it is captured.
Can you access it later froim somewhere in the test? can you have a back reference in the match string?
apart from
```
`com.myapp.controller` (group 1 = `controller`) and
`com.myapp.service` (group 1 = `service`).
The deeper packages do not match because `(*)` allows only a single segment.
```
I find no information about that part.
What is the use of group1? 
Is there any beyond allowing
```
com.myapp.(*)..
```
which seems to be the only way to exclude the parent package `com.myapp`. Or is that even excluded because there is no class in it? (I don't think so)
The section in file 8 adds some usage of it. Maybe link to that section as usage hint from here?

docs/userguide/008_The_Library_API.adoc
```diff
@@ -130,7 +130,10 @@ com.tngtech.archunit.library.dependencies.SlicesRuleDefinition
 ----
 
 The API is based on the idea to sort classes into slices according to one or several package
-infixes, and then write assertions against those slices. At the moment this is for example:
+infixes, and then write assertions against those slices. The `matching(..)` argument follows
+the package pattern syntax described in <<Package Identifiers>>, where the parentheses
+`(*)` / `(**)` mark the captured segment(s) that are used as slice identifiers.
```
StefanGraeber: this renders weirdly for me as "() / (*)", but 
```
`+++(*)+++` / `+++(**)+++`
```
instead works fine (+++ means "skip all inline styling).
This is lso works but seems to be the variant for larger blocks
```
`pass:[(*)]` / `pass:[(**)]`
```

The parser is confused by multiple stars meaning a bold section

StefanGraeber (changes requested):

docs/userguide/003_Getting_Started.adoc
StefanGraeber: definitely keep the current syntax.
The files seem to be concatenated into a single file used to build documentation on https://www.archunit.org/userguide/html/000_Index.html
there the references work and that is the most important viewer

the unified file can also be viewed in the 000_index file. it should work there

docs/userguide/006_The_Core_API.adoc
hankem: See [`SlicesRuleDefinition`](https://javadoc.io/doc/com.tngtech.archunit/archunit/latest/com/tngtech/archunit/library/dependencies/SlicesRuleDefinition.html), which can even make use of multiple capturing groups, see [`Slice.as`](https://javadoc.io/doc/com.tngtech.archunit/archunit/latest/com/tngtech/archunit/library/dependencies/Slice.html#as(java.lang.String)).

docs/userguide/006_The_Core_API.adoc
```diff
@@ -396,3 +396,114 @@ ArchUnit's own rule APIs (compare <<The Lang API>>) never rely on the
...
 next section, does not depend on whether the classes were imported from the classpath or
 some JAR / folder.
+
+=== Package Identifiers
+
+Several ArchUnit methods accept a `String` package identifier to describe a set of packages,
+for example `resideInAPackage(..)` and `resideInAnyPackage(..)` (and their negated variants)
+in the Lang API, `slices().matching(..)` and `modules().definedByPackages(..)` in the Library
+API, or the component stereotypes of a PlantUML component diagram. All of them delegate to
+the same underlying `PackageMatcher`, whose syntax is inspired by AspectJ type patterns and
```
hankem: I wonder whether it would be beneficial to additionally link to [the JavaDoc](https://javadoc.io/doc/com.tngtech.archunit/archunit/latest/com/tngtech/archunit/core/domain/PackageMatcher.html).
(I personally like cross-references, and did that for the [Domain model](https://www.archunit.org/userguide/html/000_Index.html#_domain).)

docs/userguide/006_The_Core_API.adoc
StefanGraeber: @hankem the file 8 talks about Slices and tests of them, so I think we mean the same thing.
Just from documentation here, I have no idea what these groups are used for

hankem: The immediate effect is that captured substrings can be obtained from `PackageMatcher.match`'s `Result`'s `getGroup(int)` (similar `to java.util.regex.MatchResult#group(int)`).

Slices according to package identifiers are just a prominent use case.

docs/userguide/006_The_Core_API.adoc
TheManWhoStaresAtCode: Honestly, I was wondering the same. However, my current understanding was that we somehow "implicitly" reference the JavaDoc by mentioning the respective class names in backticks (e.g. `PackageMatcher`). Therefore, I decided to stick with the convention for now.

However, I'm all up to adding the link as suggested by you.

docs/userguide/006_The_Core_API.adoc
TheManWhoStaresAtCode: First of all: thanks Stefan for raising this question. To me the use-cases for capturing and groups here are also not that clear based on the [JavaDoc in PackageMatcher.](https://javadoc.io/doc/com.tngtech.archunit/archunit/latest/com/tngtech/archunit/core/domain/PackageMatcher.html).

Based on this discussion I think we should mention that they can be obtained programmatically, as outlined by Manfred. Furthermore, I like the idea of linking to file 8 where one exemplary use-case of this feature is documented.

docs/userguide/006_The_Core_API.adoc
TheManWhoStaresAtCode: Not sure if I get this comment: Do you suggest to extend the example by a sixth class e.g. `com.myapp.ExampleClass` to demonstrate the statement "including zero" better? 

Let me rephrase this part and avoid "including zero". However, this does not mean that I'm against adding a class in the com.myapp package at all.

TheManWhoStaresAtCode:
Thanks for all the feedback - I implemented it as discussed above.

I also fixed the local docs setup as part of this task. This lead to changes in the committed html docs files (see commit f592b4ce2a12b71206fd6da046075130dd6ea595) which only contains the changes in the generated files.

docs/userguide/html/000_Index.html
hankem, line 1: You can drop the `Update docs html content to latest version` commit.
That happens automatically during the release (e.g. e45aaa20543b74daeaa09f501b8890742dc7dd8d).

TheManWhoStaresAtCode: ah good to know :bow:

docs/userguide/006_The_Core_API.adoc
hankem: It's absolutely fair to document that capturing groups are relevant for the result of `{@link #match}`.

Slices are a good example; the architecture rules wouldn't allow a reference from `PackageMatcher` (in `core`) to `Slices` (in `library`), but the JavaDoc might not be covered... 😅

docs/userguide/006_The_Core_API.adoc
```diff
@@ -396,3 +396,120 @@ ArchUnit's own rule APIs (compare <<The Lang API>>) never rely on the
 classpath though. Thus the evaluation of default rules and syntax combinations, described in the
 next section, does not depend on whether the classes were imported from the classpath or
 some JAR / folder.
+
+=== Package Identifiers
```
hankem: I feel that `Package Identifiers` should be inside `Domain`, not at the same level as `Import` and `Domain`.
```suggestion
==== Package Identifiers
```

TheManWhoStaresAtCode: done.

docs/userguide/006_The_Core_API.adoc
```diff
@@ -396,3 +396,120 @@ ArchUnit's own rule APIs (compare <<The Lang API>>) never rely on the
...
+|===
+| Pattern | Meaning
+
+| `*`
+| Matches any sequence of characters **not containing** the dot `.`,
+i.e. exactly one package segment.
+
+| `..`
+| Matches any sequence of characters that **may contain** the dot `.`,
+i.e. any number of package segments (including zero).
```
hankem: "any sequence of characters" does not have the same meaning in both lines. How about:

```suggestion
| Matches any non-empty sequence of characters **not containing** the dot `.`,
i.e. exactly one package segment.

| `..`
| Matches any (possibly empty) sequence of characters that **may contain** the dot `.`,
i.e. any number of package segments (including zero).
```

hankem: And maybe a personal taste, but I wouldn't render "**not containing**" and "**may contain**" in bold.
As a user, I'd actually find "_exactly one_ package segment" and "_any number_ of package segments" the more relevant consequence, whereas getting the dots right is just an implementation detail. But the description is short enough that I wouldn't need the highlight at all.
In any case, I won't object if you prefer the bold part. 🙂

hankem (approved):
I've added some more links to the new Package Identifiers section in the user guide (as a `fixup!` commit to be rebased, if you like it), and another commit to improve the `PackageMatcher` JavaDoc.

PR #1677, opened 2026-07-29
add support for sealed classes

### LINKED ISSUE #394: Add support for sealed classes (Java 15 preview feature)
Java 15 will bring us [JEP 360](https://openjdk.java.net/jeps/360) (preview feature) which introduces sealed classes and permitted subclasses.

Currently, we are using ASM 8.0.1 which already contains _experimental_ support for permitted subtypes. [This merge request for ASM 9](https://gitlab.ow2.org/asm/asm/-/merge_requests/291) contains the official support for this feature.

To do: Which API methods should be added? And should the permitted subclasses be considered as dependencies (if so, in which direction)?

hankem:
Read the PermittedSubclasses class file attribute via ASM's `visitPermittedSubclass` and expose it on `JavaClass` through `getPermittedSubclasses()` and `isSealed()`. Permitted subclasses are  resolved transitively (subject to `maxIterationsForPermittedSubclasses`) and folded into the existing class-hierarchy completion step.

_Current limitation:_
ArchUnit does not recognize the corner case of a class with an empty array of permitted subclasses, which is valid bytecode (https://docs.oracle.com/javase/specs/jvms/se17/html/jvms-4.html#jvms-4.7.31), even though Java requires at least one TypeName within a permits clause (https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-8.1.6)

Resolves #394.


archunit/src/jdk16test/java/com/tngtech/archunit/core/importer/ClassFileImporterSealedTest.java
```diff
@@ -0,0 +1,281 @@
...
+     */
+    @Test
+    void empty_permitted_subclasses_are_currently_not_recognized_by_ArchUnit() {
+        var location = Location.of(relativeResourceUri(getClass(), "EmptyPermittedSubclasses.class"));
+        JavaClass emptyPermittedSubclasses = getOnlyElement(new ClassFileImporter().importLocations(List.of(location)));
+
+        assertThat(emptyPermittedSubclasses.isSealed()).isFalse(); // current behavior; actually expected: .isTrue();
+        assertThat(emptyPermittedSubclasses.getPermittedSubclasses()).isEmpty(); // current behavior; actually expected: .isNotEmpty();
+        assertSealedWithPermittedSubclasses(emptyPermittedSubclasses.reflect(), new Class[0]);
+    }
```
hankem, line 172: If we don't want to keep the option to support this (pathological) use case in the future, we could simplify the `JavaClass` API so `getPermittedSubclasses()` returns a `Set<JavaClass>`.

schulzjo-tng: I might be missing something here, but don't we need `Optional.Empty` for non-sealed classes or interfaces? In case your idea was to use an empty set to indicate "class is not sealed", I don't think that's a good idea.

archunit/src/jdk16test/java/com/tngtech/archunit/core/importer/ClassFileImporterSealedTest.java
```diff
@@ -0,0 +1,281 @@
...
+    /**
+     * The Java compiler would probably not create such bytecode, but it is technically valid:
+     * {@code EmptyPermittedSubclasses.class.getPermittedSubclasses() == new Class[0] != null},
+     * i.e. {@code EmptyPermittedSubclasses.class.isSealed() == true}.
+     * As ASM's {@link ClassReader} invokes {@link ClassVisitor#visitPermittedSubclass(String)}
+     * for every permitted subclass, ArchUnit cannot know of the (empty) permitted subclasses from this information.
+     */
+    @Test
+    void empty_permitted_subclasses_are_currently_not_recognized_by_ArchUnit() {
+        var location = Location.of(relativeResourceUri(getClass(), "EmptyPermittedSubclasses.class"));
```
schulzjo-tng, line 166: Just curious: How did you create this file? Do you maybe want to add a comment with a short explanation here, just in case someone ever needs to do something similar in the future?

archunit/src/jdk16test/java/com/tngtech/archunit/core/importer/ClassFileImporterSealedTest.java
schulzjo-tng: I took a bit of a closer look.
- I agree with your assessment that we don't need to handle this edge case correctly. If there ever is a way of producing such classes in the future (which I sincerely doubt, since this is effectively a `final` class), we can revisit that point.
- I would strongly prefer to keep the API as it is. I would read `getPermittedSubclasses() == empty set` as "no subclasses are permitted", not "all classes are allowed to subclass". Even if we cannot return an `Optional.of(empty set)` at the moment, the API makes more sense to me and also more closely resembles the actual underlying data structure.

archunit/src/main/java/com/tngtech/archunit/core/importer/ClassGraphCreator.java
```diff
@@ -298,6 +298,19 @@ public List<JavaClass> createInterfaces(JavaClass owner) {
         return result.build();
     }
 
+    @Override
+    public Optional<Set<JavaClass>> createPermittedSubclasses(JavaClass owner) {
+        List<String> permittedSubclassNames = importRecord.getPermittedSubclassNamesFor(owner.getName());
```
schulzjo-tng, line 305: Maybe add a comment here that refers to the test and the known edge case

docs/userguide/010_Advanced_Configuration.adoc
```diff
@@ -83,11 +83,12 @@ For further details, compare the sources of `SelectedClassResolverFromClasspath`
 
 It is also possible to apply a more fine-grained configuration to the import dependency resolution behavior.
 
-In particular, the ArchUnit importer distinguishes 6 types of import dependencies:
+In particular, the ArchUnit importer distinguishes 7 types of import dependencies:
```
schulzjo-tng:
```suggestion
In particular, the ArchUnit importer distinguishes the following types of import dependencies:
```
maybe? Then we don't need to keep it in sync in the future

archunit/src/main/java/com/tngtech/archunit/core/importer/DependencyResolutionProcess.java
```diff
@@ -49,6 +49,11 @@ class DependencyResolutionProcess {
     private final int maxRunsForSupertypes = getConfiguredIterations(
             MAX_ITERATIONS_FOR_SUPERTYPES_PROPERTY_NAME, MAX_ITERATIONS_FOR_SUPERTYPES_DEFAULT_VALUE);
 
+    static final String MAX_ITERATIONS_FOR_PERMITTED_SUBCLASSES_PROPERTY_NAME = "maxIterationsForPermittedSubclasses";
```
schulzjo-tng, line 52: Found by AI: This value is missing in `com.tngtech.archunit.core.importer.DependencyResolutionProcessTestUtils:65` (`setAllIterationsToZeroExcept()`)

schulzjo-tng (commented):
Looks very good! I haven't explicitly verified the `maxIterationsForPermittedSubclasses`, but the code looks sound.

I have added a few minor comments. Don't forget to amend the commit with `--signoff`.

archunit/src/jdk16test/java/com/tngtech/archunit/core/importer/ClassFileImporterSealedTest.java
hankem: I started with
```java
package com.tngtech.archunit.core.importer;

sealed class EmptyPermittedSubclasses permits X {
}

final class X extends EmptyPermittedSubclasses {
}
```
and used opencode to drop the reference to `X` from the class file's `PermittedSubclasses` attribute and the constant pool. (It came up with 277 lines of Python to edit the bytes of the file. 🙈)

schulzjo-tng:
```suggestion
        // this .class file was compiled from the simplest example of a sealed class and then transformed using AI to have zero permitted subclasses
        var location = Location.of(relativeResourceUri(getClass(), "EmptyPermittedSubclasses.class"));
```
then maybe like this?

Issue #1095, opened 2023-03-31
Consider linking ArchUnit sample with best-practices


hohwille:
Dear ArchUnit-Team,
first of all I want to thank you for creating ArchUnit what is a wonderful tool to automate architecture and code-quality testing with great flexibility.
We as the community of [devonfw](https://devonfw.com/) have decided to promote ArchUnit as a best-in-class tool and have spend quite some invest into creating a small demo app with layered architecture and well crafted ArchUnit tests to demonstrate it in action and give projects a better starting point than beginning from scratch so there are no excuses for projects to omit architecture testing.
Also we started creating draft PRs with violations that show via our github action CI workflow how ArchUnit and our rules report the violations so projects can see the benefits at a glance. You can find our work here:
https://github.com/devonfw-sample/archunit

If you have a look and like what we have created, you may consider to link it from your README or other documentation place.
Further, if you like, we would have some feedback how to improve ArchUnit. So in case you are interested, please let us know, if you would prefer some general list of ideas e.g. in here as comment or I should better break down small feature-requests as individual issues.

FYI: We also had created a [architecture validation plugin for SonarQube](https://github.com/devonfw/sonar-devon4j-plugin) (see also [architecture rules](https://github.com/devonfw/sonar-devon4j-plugin/blob/master/documentation/rules.asciidoc#architectural-dependency-rules)) but ran into the problems that the plugin ecosystem of SonarQube became quite unstable causing severe maintenance issues as well as feedback from real world projects about the need for more agility and flexibility that a closed box SonarQube plugin could not easily offer but now is perfectly solved with ArchUnit.

Thanks again for your wonderful work that adds to the great Java OSS ecosystem and makes the world a better place.
Cheers
  Jörg and the entire devonfw team

PR #1692, opened 2026-08-08
docs: add community examples section to README

vijayagopalsb:
Adds a link to the devonfw-sample/archunit project as a community example for best practices with ArchUnit.

Relates-to: #1095

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,12 @@ public class MyArchitectureTest {
 #### Let the API guide you
 ![ArchUnit Fluent API](ArchUnit-API.gif)
 
+## Community Examples
+
+The following projects provide additional examples and best practices for using ArchUnit in real-world applications:
+
+- [devonfw-sample/archunit](https://github.com/devonfw-sample/archunit) — A complete layered architecture demo with well-crafted ArchUnit tests and a CI workflow demonstrating violation reporting.
+
 ## Where to look next
 
 For further information, check out the user guide at [https://archunit.org](https://archunit.org) 


vijayagopalsb:
I agree to the DCO for this contribution.

Signed-off-by: Vijayagopal SB <vijayagopal.sb@gmail.com>

README.md
hankem, line 67: Thanks for your contribution!

Three thoughts about it:

1. For the DCO check to pass, you need to amend your git commit and add the line from [your comment](https://github.com/TNG/ArchUnit/pull/1692#issuecomment-5226077272) there, see also [DCO](https://github.com/TNG/ArchUnit/blob/main/DCO).
2. I think community examples should be mentioned after the `Where to look next` section with the user guide and the "official" [ArchUnit Examples](https://github.com/TNG/ArchUnit-Examples).
3. While I'm not opposed to linking community examples in general, I'm not specifically convinced either. I don't want to offend anybody who created [devonfw-sample/archunit](https://github.com/devonfw-sample/archunit); it was a nice example when it was created! I just wonder what is so special about it that's "worth" the direct link (and not just be discovered via search engines like other good examples and blog posts on the internet):
   * It's a Maven project using [ArchUnit 1.0.1](https://github.com/TNG/ArchUnit/releases/tag/v1.0.1) (released on 2022 Nov 21), illustrating the integration with JUnit5. It basically hasn't been updated for 3 years.
   * [TNG/ArchUnit-Examples](https://github.com/TNG/ArchUnit-Examples/), on the other hand, contain Gradle and Maven examples for plain ArchUnit as well as its integration with JUnit 4 / 5 / 6; they're guaranteed to work as they're derived from ArchUnit's own build, and automatically updated with every release. If an example was missing, shouldn't we rather add it there?

PR #1713, opened 2026-09-04
Make ArchUnit compatible with Gradle test retry plugin

### LINKED ISSUE #1711: ArchUnit doesn't work with Gradle's test-retry plugin
The [test-retry](https://github.com/gradle/test-retry-gradle-plugin) plugin works by adding *filters* to only include classes it wants to retry. However, it has no control over the discovery 
*selectors* - those are always ClassSelectors in Gradle, one for each test class that the test task contained. ArchUnit currently only applies filters to classpath roots and packages, but not to ClassSelectors. As a result, whenever any test is rerun, *all* ArchUnit tests are also rerun.

ArchUnit should apply the filters to ClassSelectors as well, before this line:

https://github.com/TNG/ArchUnit/blob/30cebca6583c259f6a7165e8b1057224f5b41593/archunit-junit/junit5/engine/src/main/java/com/tngtech/archunit/junit/internal/ArchUnitTestEngine.java#L129

oehme:
Gradle always uses individual ClassSelectors for each test class in the test task. The test retry plugin then layers filters on top to include only the tests it wants to rerun. This change ensures these filters are actually honored.

Without this, all ArchUnit tests are rerun, regardless if they failed.

Fixes #1711


oehme:
Ping :slightly_smiling_face:

oehme:
My bad, after some more digging I realized that the issue should be fixed on the Gradle side

